PR #22, opened 2019-09-17
2: Adds funding column to roster 

### LINKED ISSUE #2: Roster has funding column
**Story:** As a provider, I want to see which of my kids are enrolled in a CDC (Child Day Care contract) slot.

**Acceptance criteria:**

* My roster has a column called "Funding".
* This column is not sortable.
* This column includes a tag for those enrollments that are actively funded by a public source.
  * "Actively funded" means that the funding entry date is less than or equal to today, and the exit date is null, greater than, or equal to today.
* The text of the tag is determined by the funding source. Currently, this is just "CDC" for Child Day Care contract funding.
* The tag should follow the [USWDS tag component](https://designsystem.digital.gov/components/tag/).

**Design:**

Screen Shot 2019-09-15 at 9 14 04 AM

jlhogan:
closes #2

### DIFF
--- a/ClientApp/src/components/Tag/Tag.tsx
+++ b/ClientApp/src/components/Tag/Tag.tsx
@@ -1,0 +1,11 @@
+import React from 'react';
+
+export type TagProps = {
+    text: String;
+};
+
+export default function Tag({ text }: TagProps) {
+    return (
+        <span className="usa-tag">{text}</span>
+    );
+}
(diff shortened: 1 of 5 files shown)


ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -65,6 +66,17 @@ export default function Roster() {
 				cell: ({ row }) => <td>{dateFormatter(row.child.birthdate)}</td>,
 				sort: row => row.child.birthdate,
 			},
+			{
+				name: 'Funding',
+				cell: ({ row }) => (
+					<td>
+						{row.fundings.length
+							? <Tag text={`${row.fundings[0].source}`} />
```
cailyncodes: So this would only show one funding stream, correct? I think this is fine because we are only concerned with CDC but I just want to make sure.

ClientApp/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export type TagProps = {
+    text: String;
+};
+
+export default function Tag({ text }: TagProps) {
+    return (
+        <span className="usa-tag">{text}</span>
+    );
+}
```
cailyncodes: It's so tidy! I love it

Issue #24, opened 2019-09-18
Setup AWS CodeBuild project to build & package hedwig artifacts


speedcm:
Setup AWS CodeBuild to build the hedwig artifacts.  This requires a) setting up an AWS CodeBuild project  and b) creating the CodeBuild buildspec.yml file to build the dotnetcore app & package the artifacts.  

The artifacts for the build consist of a) the app site files and b) a SQL script for the associated database context (SQL Server).  The site files must meet the packaging requirements to deploy to AWS Beanstalk. 

The site files required for a Beanstalk deployment will require committing new files to the repository such as web.config & aws-windows-deployment-manifest.json.

This issue is dependent upon a)  issue #17  and b) refreshing the old GitHub repository with the latest hedwig source. 

The old GitHub repository will be used for unit coding & unit testing the code associated with this effort.

speedcm:
I am unable to setup a CodeBuild project due to insufficient rights.  when I access CodeBuild via the AWS console I get the following notice:

Error
User: arn:aws:iam::906392371336:user/xxx.xxx is not authorized to perform: codebuild:ListProjects

Please note that I need full access to CodeBuild and full access to IAM.  Please make me an Administrator.

speedcm:
CodeBuild was successfully working prior to the directory structure change.   However, recent changes to the directory structure broke the build.  The developer (speedcm) attempted to refactor the driver file based on the new directory structure.  However, there was no way to unit test the change (code was not refreshed to sandbox environment).  The buildspec.yml file is still broken and needs to be updated for the directory structure change.

speedcm:
All tasks identified in this issue are now complete.  The AWS CodeBuild project is building with the live repository.   After the build is complete, the build artifacts are packaged in the project's S3 bucket.

PR #30, opened 2019-09-19
Devops codebuild initial

speedcm:
add codebuild support (#24)


speedcm:
add codebuild support (#24)

PR #32, opened 2019-09-19
Support windows development with docker

cailyncodes:
Fully support Docker development on Windows with only the addition of an environment variable.

(This isn't a perfect solution, of course, but it provides a consistent experience across all platforms. And is much better than previous solutions.)

### DIFF
--- a/ClientApp/entrypoint.sh
+++ b/ClientApp/entrypoint.sh
@@ -1,12 +1,27 @@
+if [ "$DOCKER_ON_WINDOWS" ]; then
+    cd /home/node/mapped-app
+fi
+
 nm=$(ls -A ./node_modules/*)
 if [ -z "$nm" ]; then
     echo "installing yarn dependencies..."
     yarn install
 fi
 
-echo "starting client..."
 if [ "$DOCKER_ON_WINDOWS" ]; then
-    CHOKIDAR_USEPOLLING=true CHOKIDAR_INTERVAL=2500 yarn start
+    # Reproduce files from shared mount point to container-internal directory 
+    rsync -aq --exclude node_modules /home/node/mapped-app/* /home/node/app > /dev/null 2>& 1
+    cd /home/node/app
+    echo "starting client..."
+    yarn start &
+
+    # Poll for changes and sync them
+    # This enables chokidar watcher to process inotify events
+    while true;
+    do
+        rsync -aq /home/node/mapped-app/* /home/node/app > /dev/null 2>& 1
+    done
 else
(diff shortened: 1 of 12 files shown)


cmgiven:
I'm going to take myself off this and leave this to Julia if that's alright. I think her feedback will be far more useful than mine!

docker-compose-windows.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - 5000:5000
+    entrypoint:
+      - sh
+      - /app/entrypoint.sh
+    depends_on:
+      - db
+  client:
+    build:
+      context: .
+      dockerfile: ./ClientApp/Dockerfile
```
jlhogan: for consistency should this be named Dockerfile-dev as well ?

dc-dotnet
```diff
@@ -1 +1,5 @@
-docker-compose exec backend dotnet $@
+if [ "$DOCKER_ON_WINDOWS" ]; then
```
jlhogan: also for consistency should this wrapper just be around `docker-compose exec backend` like the new `dc-client` is ? that seems like a more intuitive interaction. I guess it doesn't really matter much; i'm not clear on if OEC devs would ever use this docker-compose set up to develop?

ClientApp/entrypoint.sh
```diff
@@ -1,12 +1,24 @@
...
 nm=$(ls -A ./node_modules/*)
 if [ -z "$nm" ]; then
     echo "installing yarn dependencies..."
     yarn install
 fi
 
-echo "starting client..."
 if [ "$DOCKER_ON_WINDOWS" ]; then
-    CHOKIDAR_USEPOLLING=true CHOKIDAR_INTERVAL=2500 yarn start
+    rsync -aq --exclude node_modules /home/node/mapped-app/* /home/node/app > /dev/null 2>& 1
```
jlhogan: can you capture the necessity of this in comment somewhere? To make it clear what the issue was, and how this solves it?

dc-dotnet
cailyncodes: I'm good with either approach -- I was trying to minimize the amount that I changed the interface you had set up, but can go entirely consistent with one approach.

cailyncodes:
Oops yeah totally! Once I got the npm modules thing working, I excitedly pushed, forgetting to add the comment in.

cailyncodes:
Idea for `rsync` based solution inspired by https://github.com/docker/for-win/issues/969. Addition comments added to `entrypoint.sh` for client container.

PR #34, opened 2019-09-20
Add ability to print roster (WIP)

### LINKED ISSUE #3: Roster can be printed
**Story:** As a provider, I want to print out my roster for my records.

**Acceptance criteria:**

* My roster has a button below the table of enrollments labeled "Print".
* This button uses the "outline" style from USWDS.
* Clicking this button opens the browser print dialog.
* A print stylesheet changes the page:
  * The header is hidden.
  * Buttons like the "Print" button are hidden.
  * Links are ink-colored and not underlined.
  * The content fits to the width of a portrait 8.5x11" page.

**Design:**

Screen Shot 2019-09-15 at 9 15 24 AM

Screen Shot 2019-09-15 at 9 27 22 AM

cailyncodes:
WIP -- need to figure out how to test components that use print stylesheets.

Should resolve #3.

### DIFF
--- a/ClientApp/src/contexts/PrintMode/PrintModeHOC.tsx
+++ b/ClientApp/src/contexts/PrintMode/PrintModeHOC.tsx
@@ -1,0 +1,43 @@
+import React, { useState } from 'react';
+import { Subtract } from 'utility-types';
+import { PrintModeConsumer, PrintModeProvider as InternalPrintModeProvider } from './PrintModeContext';
+
+export type PrintModeProviderPropsType = {
+  value: boolean
+}
+
+const PrintModeProvider: React.FC<PrintModeProviderPropsType> = ({value, children}) => {
+  const [printMode, setPrintMode] = useState(value);
+  
+  const togglePrintMode = (_: any) => {
+    setPrintMode(!printMode);
+  }
+
+  return (
+    <InternalPrintModeProvider value={{
+      value: printMode,
+      setValue: togglePrintMode
+    }}>
+      {children}
+    </InternalPrintModeProvider>
+  );
+}
+
+export type WithPrintModePropsType = {
+  isPrintMode: boolean,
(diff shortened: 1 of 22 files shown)


ClientApp/src/contexts/PrintMode/PrintModeHOC.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { useState } from 'react';
```
jlhogan: what does `HOC` mean? (PrintModeHOC.tsx)

cailyncodes: Higher order component -- a function that takes in a component and returns a new component (https://reactjs.org/docs/higher-order-components.html).

cailyncodes:
Oh yeah you're right that we don't need that check on the header in `App.tsx`. However, I think the context (or at the very least prop passing) is necessary in `Roster.tsx` for the print cancel button.

cailyncodes:
Actually, I take that back. Because the "print mode" display vs the actual print rendering use slightly different style rules. And even if they did use the exact same, it would require adding a class to the element (or `<App>`) dynamically to cause the "print mode" styles to be rendered.

cailyncodes:
The reason for the "print mode" display at all, is because Firefox doesn't give a rendering like Chrome does of what the page looks like printed. It only gives a print dialog.

cmgiven:
Do our users need that rendering? Is there a reason I'd want to see that in addition to the print dialog? I speculate that the print mode, which necessitates the cancel, actually makes this more complicated for users as well.

cmgiven:
All we've heard from users at this point is "I'd like to print my roster." If there's a more minimal version of this feature that could satisfy that need, I'd suggest we start there, only pursuing more involved approaches if future user research indicates it's needed.

PR #37, opened 2019-09-23
35 backend tests

jlhogan:
Reorganizes the repo, adding a top-level solution file (Hedwig.sln) that includes two projects: the application (`src/Hedwig`), and the tests (`test/HedwigTests`). 

Actual functioning tests to come


entrypoint-client.sh
```diff
@@ -1,27 +0,0 @@
-if [ "$DOCKER_ON_WINDOWS" ]; then
-    cd /home/node/mapped-app
```
cailyncodes: Note to @cailyncodes -- This probably needs to be moved after the `yarn install`. I think it's related to my current issues with running tests. Not related to this PR but in reviewing it gave me inspiration haha.

.devcontainer/docker-compose.yaml
```diff
@@ -35,5 +37,4 @@ services:
     #   - seccomp:unconfined
 
     # Overrides default command so things don't shut down after the process ends.
-    command: sleep infinity 
- 
+    command: sleep infinity
```
cailyncodes: Oh this is cool! Much easier to remember than my usual solution for never ending process.

.devcontainer/docker-compose.yaml
```diff
@@ -20,13 +20,13 @@ services:
     # Uncomment if you want to expose any additional ports. The snippet below exposes port 3000.
     # ports:
     #   - 3000:3000
-    
+
     volumes:
       # Update this to wherever you want VS Code to mount the folder of your project
-      - ..:/app
+      - ../..:/app/
```
cailyncodes: I'm confused about this. If this is in `[root]/.devcontainer` wouldn't `../..` take us up one too many directories?

jlhogan: yes!!! you're right. Wonder why it was working...

PR #39, opened 2019-09-23
Add print stylesheet

cailyncodes:
Simplify implementation for User story #3. This PR adds a print style sheet with the use of CSS media queries.

This implementation has a known limitation on being able to render a display on Firefox in the browser. This is due to how Firefox implements printing (versus Chrome-based browsers). For an implementation that provides a potential solution, see #34 (on branch 3-print-roster-with-context).

### DIFF
--- a/src/Hedwig/ClientApp/src/assets/styles/index.scss
+++ b/src/Hedwig/ClientApp/src/assets/styles/index.scss
@@ -34,3 +34,4 @@
 // Import theme custom styles
 
 @import 'uswds-theme-custom-styles';
+@import 'print';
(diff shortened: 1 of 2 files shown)


cailyncodes:
@cmgiven Optional review

ClientApp/src/containers/App/App.test.tsx
```diff
@@ -16,3 +17,19 @@ it('renders without crashing', () => {
...
+		<MockedProvider>
+			<BrowserRouter>
+				<div className="print">
+					<App />
+				</div>
+			</BrowserRouter>
+		</MockedProvider>
+	);
+
+	expect(wrapper).toMatchSnapshot();
```
cmgiven: Seems like this wouldn't actually test anything other than that the className has been added. The snapshots just run a diff on the markup; they're not doing a visual comparison.

I'm not sure that there is a way to test this feature without adding visual regression testing.

cailyncodes: Welp okay yeah I'm not sure we can test this then. I'll clean up the test and style sheet.

cailyncodes:
Closes #3

PR #47, opened 2019-09-26
40: Add meaningful tests for backend to test project

jlhogan:
Also:
- adds backend test run to azure and aws build
- removes unnecessary using directives
- organizes service configuration in Startup.cs with ServiceExtension.cs


test/HedwigTests/Helpers/FamilyHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+		}
+
+		public static List<Family> CreateFamilies(HedwigContext context, int numberOfFamilies)
+		{
+			if (numberOfFamilies < 1) {
+				throw new ArgumentException(
+					message: "Number of families cannot be less than 1",
+					paramName: nameof(numberOfFamilies)
+				);
+			}
```
cailyncodes: Same as above

test/HedwigTests/Helpers/FundingHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+            context.Fundings.Add(funding);
+            context.SaveChanges();
+            return funding;
+        }
+
+        public static Funding CreateFundingWithEnrollmentId(HedwigContext context, int enrollmentId)
+        {
+            var funding = new Funding{
+                EnrollmentId = enrollmentId,
+                Source = FundingSource. CDC
```
cailyncodes: Looks like there is an extra space here before CDC

test/HedwigTests/Repositories/ChildRepositoryTests.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace HedwigTests.Repositories
+{
+	public class ChildRepositoryTests
+	{
+		[Fact]
+		public async Task Get_Children_By_Ids()
+		{
+			using (var context = new TestContextProvider().Context) {
+				// If 5 children with auto-incrementing Ids exist
```
cailyncodes: How are we enforcing the auto id increment?

test/HedwigTests/Repositories/EnrollmentRepositoryTests.cs
```diff
@@ -0,0 +1,56 @@
...
+				Assert.Equal(enrollment.Id, res.Id);
+				Assert.Equal(enrollment.ChildId, res.ChildId);
+			}
+		}	
+
+		[Fact]
+		public async Task Get_Enrollments_By_Site_Id()
+		{
+			using (var context = new TestContextProvider().Context) {
+				// If enrollments exists with side Ids
```
cailyncodes: Site instead of side?

test/HedwigTests/Repositories/FamilyRepositoryTests.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace HedwigTests.Repositories
+{
+    public class FamilyRepositoryTests
+    {
+        [Fact]
+        public async Task Get_Families_By_Ids()
+        {
+            using (var context = new TestContextProvider().Context){
+                // If 5 families with auto-incrementing Ids exist
```
cailyncodes: Same as above

src/Hedwig/hedwig.csproj
```diff
@@ -30,16 +30,6 @@
...
 
-	<Target Name="DebugEnsureNodeEnv" BeforeTargets="Build" Condition=" '$(Configuration)' == 'Debug' And !Exists('$(SpaRoot)node_modules') ">
-		<!-- Ensure Node.js is installed -->
-		<Exec Command="node --version" ContinueOnError="true">
-			<Output TaskParameter="ExitCode" PropertyName="ErrorCode" />
-		</Exec>
-		<Error Condition="'$(ErrorCode)' != '0'" Text="Node.js is required to build and run this project. To continue, please install Node.js from https://nodejs.org/, and then restart your command prompt or IDE." />
-		<Message Importance="high" Text="Restoring dependencies using 'npm'. This may take several minutes..." />
-		<Exec WorkingDirectory="$(SpaRoot)" Command="npm install" />
-	</Target>
```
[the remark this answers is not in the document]
jlhogan: if i understand correctly, this check is only for `Debug` builds which are development. 

In development now, the client and backend are separated, so backend should not (and can not!) check for these client deps: they don't exist in the backend container. 

The reason we weren't running into issues with this before was likely due to `node_modules` existing and being mounted into the container where build occurred.

test/HedwigTests/Repositories/ChildRepositoryTests.cs
jlhogan: autoincrementing is the wrong word here b/c these are Guids not ints (will update!) but the fields are  annotated with "SqlServer:ValueGenerationStrategy": identity column in the migration. I believe fields named 'Id' are assumed to be identity columns by ef.

PR #48, opened 2019-09-26
Add filter for enrollments by date (#4)

### LINKED ISSUE #4: Enrollment data can be filtered by date range
**Story:** As an API consumer, I want to filter enrollments by date range to only fetch the data I need.

**Acceptance criteria:**

* I should be able to make a GraphQL query like the following:

```
{
  me {
    sites {
      id
      enrollments(from: "2018-09-01" to: "2019-08-31") {
        id
      }
    }
  }
}
```
* The results should only include enrollments that intersect the provided range.
  * An enrollment that ends on the first day of the range or starts on the last day of the range should be included.
  * An enrollment with a null exit date should be included if its entry date is less than or equal to the last day of the range.

cailyncodes:
Adds arguments to EnrollmentRepository fields, expose queryable interface for more modular searching.

I'm very much not sure if this is a good way about implementing this. @jlhogan If there's an article you reference about your structure of sub-queries, I would love to take a look! None of the documentation I found describes the structure we have, so I read between the lines and came up with this.

Should resolve #4

### DIFF
--- a/src/Hedwig/Schema/Types/EnrollmentType.cs
+++ b/src/Hedwig/Schema/Types/EnrollmentType.cs
@@ -11,8 +11,8 @@ namespace Hedwig.Schema.Types
 		public EnrollmentType(IDataLoaderContextAccessor dataLoader, IChildRepository children, IFundingRepository fundings)
 		{
 			Field(e => e.Id);
-			Field(e => e.Entry);
-			Field(e => e.Exit, nullable: true);
+			Field<DateTime>(e => e.Entry);
+			Field<DateTime?>(e => e.Exit, nullable: true);
 			Field<NonNullGraphType<ChildType>>(
 				"child",
 				resolve: context =>
--- a/src/Hedwig/Schema/Types/SiteType.cs
+++ b/src/Hedwig/Schema/Types/SiteType.cs
@@ -1,5 +1,7 @@
+using System;
 using Hedwig.Models;
 using Hedwig.Repositories;
+using GraphQL;
 using GraphQL.DataLoader;
 using GraphQL.Types;
 
@@ -13,11 +15,33 @@ namespace Hedwig.Schema.Types
 			Field(s => s.Name);
 			Field<NonNullGraphType<ListGraphType<NonNullGraphType<EnrollmentType>>>>(
 				"enrollments",
+				// Add arguments to subfield because graphql does not support
+				// using top-level queries on a subfield.
+				arguments: new QueryArguments(
+					new QueryArgument<DateGraphType> { Name = "from" },
+					new QueryArgument<DateGraphType> { Name = "to" }
+				),
 				resolve: context =>
 				{
+					// dynamically determine which query to use
+					// depending on supplied parameters
+					var query = enrollments.GetQuery();
+					String queryName;
+					var from = context.GetArgument<DateTime?>("from");
+					var to = context.GetArgument<DateTime?>("to");
+
+					if (!from.HasValue && !to.HasValue) {
+						queryName = "GetEnrollmentsBySiteIdsAsync";
+					} else if (!from.HasValue || !to.HasValue) {
+						throw new ExecutionError("Both from and to must be supplied");
+					} else {
+						query = enrollments.FilterByDates(query, (DateTime)from, (DateTime)to);
+						queryName = "GetEnrollmentsBySideIdsFilteredByDatesAsync";
+					}
+						
 					var loader = dataLoader.Context.GetOrAddCollectionBatchLoader<int, Enrollment>(
-						"GetEnrollmentsBySiteIdsAsync",
-						enrollments.GetEnrollmentsBySiteIdsAsync);
+						queryName,
+						(ids) => enrollments.GetEnrollmentsBySiteIdsAsync(query, ids));
 
 					return loader.LoadAsync(context.Source.Id);
 				}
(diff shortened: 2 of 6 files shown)


cailyncodes:
Oh also no tests yet -- waiting for the other PR to land before trying to write the tests. Will do before merging though.

jlhogan:
> Oh also no tests yet -- waiting for the other PR to land before trying to write the tests. Will do before merging though.

other pr landed :)

src/Hedwig/Schema/Types/SiteType.cs
```diff
@@ -13,11 +15,29 @@ public SiteType(IDataLoaderContextAccessor dataLoader, IEnrollmentRepository enr
...
 				{
+					var query = enrollments.GetQuery();
+					String queryName;
+					var from = context.GetArgument<DateTime?>("from");
+					var to = context.GetArgument<DateTime?>("to");
+
+					if (!from.HasValue && !to.HasValue) {
+						queryName = "GetEnrollmentsBySiteIdsAsync";
+					} else if (!from.HasValue || !to.HasValue) {
+						throw new ExecutionError("Both from and to must be supplied");
```
cmgiven: It looks like `"to"` is nullable?

cailyncodes: My understanding was that `.HasValue` works on nullable types and is sort of a null/empty-string check combined?

cmgiven: Sorry, should have been clearer. Why would `"to"` be nullable if we're going to throw an error here? And it looks like the repository supports `"to"` being null?

jlhogan: not sure what you mean by this. Could implement the query & repository methods to only require a from, and make to optional but that is not how this is implemented right now, nor is that in the AC. If that's the ask, let's update the ticket!

cmgiven: That’s fine. But in that case the API should not be marked as nullable.

I’m looking at line 27 of the repository for reference.

cailyncodes: Okay I'm not quite sure if I fully understand but this is what I potentially think might be the confusion.

There are actually two sub-queries we support: 1) selecting all enrollments in a site and 2) selecting all enrollments from date A to date B. The former looks like:
```
     

{
  me {
    sites {
      id
      enrollments {
        id
      }
    }
  }
}
```
and the later like
```
     

{
  me {
    sites {
      id
      enrollments(from: "2018-09-01" to: "2019-08-31") {
        id
      }
    }
  }
}
```
So both "from" and "to" are nullable on the first query and required on the second query. If only one is provided, then we throw the error. The if statements sort of hide that logic. The first if condition checks for both null, meaning use the first query; the second if condition while it looks like is an inclusive is actually an exclusive or (because the first conditional handles the inclusive case), which means error; the else then implies that both are included hence use the second query.

I hope that helps clear this up! But any case, a learning moment to be more clear in the code.

src/Hedwig/Schema/Types/EnrollmentType.cs
```diff
@@ -11,8 +11,8 @@ public class EnrollmentType : ObjectGraphType<Enrollment>
 		public EnrollmentType(IDataLoaderContextAccessor dataLoader, IChildRepository children, IFundingRepository fundings)
 		{
 			Field(e => e.Id);
-			Field(e => e.Entry);
-			Field(e => e.Exit, nullable: true);
+			Field<DateTime>(e => e.Entry);
```
cmgiven: Should this be marked as nullable as well, in that case?

src/Hedwig/Schema/Types/SiteType.cs
cmgiven: Gotcha, thanks for clarifying, and sorry to create so much back and forth. New comment added, though, as I think this means that `"from`" should also be marked as nullable.

src/Hedwig/Schema/Types/EnrollmentType.cs
cailyncodes: Oh this is different from the query parameter -- My understanding is this is what's actually stored in the db, which an enrollment only makes sense within a context of a non-null entry date.

src/Hedwig/Schema/Types/SiteType.cs
```diff
@@ -13,11 +15,29 @@ public SiteType(IDataLoaderContextAccessor dataLoader, IEnrollmentRepository enr
 			Field(s => s.Name);
 			Field<NonNullGraphType<ListGraphType<NonNullGraphType<EnrollmentType>>>>(
 				"enrollments",
+				arguments: new QueryArguments(
+					new QueryArgument<DateGraphType> { Name = "from" },
+					new QueryArgument<DateGraphType> { Name = "to" }
```
cailyncodes: Both of these are marked nullable here in the query.

cmgiven: Ah! Thank you!

cailyncodes:
@jlhogan Definitely agree with "instead of with chaining if/elses which will become cumbersome as we add more queries/filter options" -- any ideas we can think of down the road would be great to update the current approach.

And "Overall, i'm a fan of commenting all functions with a docs-style format" -- done! I've never used this XML based format before but can definitely get behind documenting all methods like this!

PR #60, opened 2019-10-01
Update typescript and CRA

cmgiven:
Simple dependency update. Needed to update typescript to pull in bug fix for infinite loop of "Files successfully emitted, waiting for typecheck results..." → " Compiled successfully!". Only saw that in Docker, not sure if others too.

Unfortunately this doesn't solve high CPU usage which is what I was checking out, particularly in the client container. But hey, it's one thing ruled out.

### DIFF
--- a/src/Hedwig/ClientApp/package.json
+++ b/src/Hedwig/ClientApp/package.json
@@ -34,10 +34,10 @@
 		"react-docgen-typescript-loader": "^3.1.1",
 		"react-dom": "^16.9.0",
 		"react-router-dom": "^5.0.1",
-		"react-scripts": "3.1.1",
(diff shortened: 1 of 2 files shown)


cailyncodes:
Looks fine to me. My first thought on high CPU usage is of you have the DOCKER_ON_WINDOWS env var set. You probably wouldn't of course, but I wonder if there is a difference in how a mac parses the conditions in entry point-client.sh causing rsync to run? Its probably not this but just thought I would flag it

cmgiven:
That was my first guess too. But doesn’t look like it. It is something that the node process is doing. 🤷🏻‍♂️

PR #78, opened 2019-10-07
8: Adds support for system-versioned temporal tables

jlhogan:
Adds GraphQL Client for tests

Add docker-compose test run to azure pipeline

Better types and defaults for query arguments

Add verbosity to migration generation

Try to fix flakiness of test by implementin g GetChildByIdAsync in child repo

Block obj and bin mounts to avoid file permission errors

Use DockerCompose task

Extends temporal support and adds more abstraction

Types with temporal support:
    - Child
    - Family
    - FamilyDetermination
    - Enrollment
    - Funding

GraphQL Types with temporal support:
    - ChildType (get versioned family)
    - EnrollmentType (get versioned child, fundings)
    - FamilyType (get versioned determinations)
    - SiteType (get versioned enrollments)

GraphQL Queries with temporal support:
    - ChildQuery
    - EnrollmentQuery

Adds classes:
    - TemporalGraphType<T>
    - TemporalRepository
    - HedwigRepository

Adds tests for most of the above

Resolves conflits with date-filterable Enrollments on sites

Adds:
    - "NotMapped" Sites property to User to make testing easier
Refactors:
    - Handling of date filter params; now passed into enrollment
repository
    - Application of date filter params; adds static query extension
methods to enable chaining of filtering with other query logic

Makes tests run serially


src/Hedwig/Schema/TemporalGraphType.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+
+namespace Hedwig.Schema
+{
+    public class TemporalGraphType<T> : ObjectGraphType<T>
+    {
+        public const String AS_OF_KEY = "asOf";
+        public static void SetAsOfGlobal(ResolveFieldContext<T> context, DateTime? asOf)
+        {
+            var userContext = context.UserContext as UserContext;
```
cmgiven: I assume that once we start authorizing requests, we would attach that authorization to a `UserContext`? And I would expect that such a context would persist across multiple requests for the lifetime of an authentication token.

Perhaps we should instead call this context `RequestContext`? And ensure that it doesn't persist across multiple requests? (I can't tell whether it does or does not right now. Perhaps I'll find that as I keep going. If not, that would be a good test to write: a second request after a request with an `asOf` does not persist the `asOf`.)

src/Hedwig/Schema/Queries/EnrollmentQuery.cs
```diff
@@ -1,22 +1,26 @@
...
+	public class EnrollmentQuery : TemporalGraphType<object>, IAppSubQuery
 	{
 		public EnrollmentQuery(IEnrollmentRepository repository)
 		{
             Field<EnrollmentType>(
                 "enrollment",
                 arguments: new QueryArguments(
-                    new QueryArgument<IdGraphType> { Name = "id" }
+                    new QueryArgument<NonNullGraphType<IdGraphType>> { Name = "id" },
+                    new QueryArgument<DateGraphType> { Name = "asOf" }
```
cmgiven: This approach does mean that we want to ensure that we only ever allow a single `asOf` argument and it must be in the top-level query (as if it was anywhere in the GraphQL tree, it would still apply to all of it because the global would be set before the queries are run). That's not a problem. Just flagging that fact.

src/Hedwig/Schema/Types/ChildType.cs
```diff
@@ -29,10 +30,12 @@ public ChildType(IDataLoaderContextAccessor dataLoader, IFamilyRepository famili
 					if (!context.Source.FamilyId.HasValue) { return null; }
 
 					var familyId = context.Source.FamilyId.Value;
+	 				var asOf = GetAsOfGlobal(context);
+					String loaderCacheKey = $"GetFamiliesByIdsAsync{asOf.ToString()}";
```
cmgiven: This will throw an exception if `asOf` is `null`, won't it?

src/Hedwig/Schema/Types/SiteType.cs
```diff
@@ -7,45 +7,40 @@
...
-					var query = enrollments.GetQuery();
-					String queryName;
 					var from = context.GetArgument<DateTime?>("from");
 					var to = context.GetArgument<DateTime?>("to");
-
-					if (!from.HasValue && !to.HasValue) {
-						queryName = "GetEnrollmentsBySiteIdsAsync";
-					} else if (!from.HasValue || !to.HasValue) {
+					if(!ValidateQueryArguments(from, to)) {
 						throw new ExecutionError("Both from and to must be supplied");
```
cmgiven: Makes sense to break this out, but I'd suggest the error message should be associated with the validation. So the method might be `ValidateDateRangeArguments`, and if it throws in that method if it doesn't pass. And then if we want to add other validations, they have their own methods, and their own errors.

src/Hedwig/Schema/Types/ChildType.cs
```diff
@@ -29,10 +30,12 @@ public ChildType(IDataLoaderContextAccessor dataLoader, IFamilyRepository famili
...
 
 					var loader = dataLoader.Context.GetOrAddBatchLoader<int, Family>(
 						"GetFamiliesByIdsAsync",
```
cailyncodes: Should this be `loaderCacheKey` here?

src/Hedwig/Schema/Types/EnrollmentType.cs
```diff
@@ -28,9 +30,11 @@ public EnrollmentType(IDataLoaderContextAccessor dataLoader, IChildRepository ch
 				"fundings",
 				resolve: context =>
 				{
+					DateTime? asOf = GetAsOfGlobal(context);
+					String loaderCacheKey = $"GetFundingsByEnrollmentIdsAsync{asOf.ToString()}";
```
cailyncodes: What happens if `asOf` is null when serializing?

src/Hedwig/Schema/Types/FamilyType.cs
```diff
@@ -2,21 +2,29 @@
...
+					DateTime? asOf = GetAsOfGlobal(context);
+					String loaderCacheKey = $"GetDeterminationsByFamilyIdsAsync{asOf.ToString()}";
+
 					var loader = dataLoader.Context.GetOrAddCollectionBatchLoader<int, FamilyDetermination>(
-						"GetDeterminationsByFamilyIdsAsync",
-						determinations.GetDeterminationsByFamilyIdsAsync);
+						loaderCacheKey,
+						(ids) => {
+							return determinations.GetDeterminationsByFamilyIdsAsync(ids, asOf);
+						}
```
cailyncodes: Can we write this without the additional block and return?

cailyncodes: Can we write this without the additional block and return?

src/Hedwig/Schema/Types/SiteType.cs
```diff
@@ -7,45 +7,40 @@
...
+						(ids) => enrollments.GetEnrollmentsBySiteIdsAsync(ids, asOf, from, to));
 					return loader.LoadAsync(context.Source.Id);
 				}
 			);
 		}
+        private static Boolean ValidateQueryArguments(DateTime? from, DateTime? to)
+		{
+			if(from.HasValue && to.HasValue) return true;
+			if(!from.HasValue && !to.HasValue) return true;
+			return false;
```
cmgiven: Could be `return from.HasValue == to.HasValue;`

test/HedwigTests/TestContextProvider.cs
```diff
@@ -1,29 +1,39 @@
...
+                    services.AddLogging(configure => configure.AddConsole());
+                }
+                
+                var options = new DbContextOptionsBuilder<HedwigContext>()
+                    .UseSqlServer(Environment.GetEnvironmentVariable("SQLCONNSTR_HEDWIG"))
+										.EnableSensitiveDataLogging()
+                    .UseInternalServiceProvider(services.BuildServiceProvider())
+                    .Options;
+                Context = new HedwigContext(options);
+                Context.Database.Migrate();
```
cmgiven: I think this is probably fine for now. But I wonder whether, particularly as projects get older and have many more migrations, this could become a major performance sink.

Did you explore whether there are options to delete the data instead of dropping the database? There might be a basic version of that where we iterate through all the tables running `DELETE FROM [TABLE]`.

src/Hedwig/Migrations/20190909160144_InitialCreate.cs
```diff
@@ -8,6 +8,7 @@ public partial class InitialCreate : Migration
     {
         protected override void Up(MigrationBuilder migrationBuilder)
         {
+            migrationBuilder.CreateHistorySchema();
```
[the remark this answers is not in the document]
jlhogan: I am kind of not a fan of building up a lot of migrations during initial dev when there's no live QA/staging cluster etc. To my mind, it just creates unnecessary clutter, and I don't really think there's value in capturing the steps by which initial schema was created. So I would +1 wipe all migrations and start from scratch, but would advocate for doing that at some point when we're going to deploy this to a use-able QA/staging place!

src/Hedwig/Schema/Queries/EnrollmentQuery.cs
jlhogan: depends on implementation. right now, asof globals are only set in queries, which are always top-level queries in our implementation. sub-queries are implemented on type definitions. Do you think there's a more ideal way to handle asof in query? this was what my brain came up with to be least error-prone, but very interested in others' thoughts

src/Hedwig/Schema/Types/ChildType.cs
jlhogan: no, `ToString()` on a nullable type returns empty string if it is null

test/HedwigTests/TestContextProvider.cs
jlhogan: yeah, discussed this with @cailyncodes and opened a placeholder ticket that needs to be fleshed out to capture this work. The way the test fixtures work now was kind of just pulled from examples on the internet, and is far from ideal for our situation, but I didn't want to block getting this out on fixing that up. It can be the thing I tackle next tho! https://github.com/ctoec/ecis-experimental/issues/79

src/Hedwig/Schema/Types/EnrollmentType.cs
jlhogan: empty string

src/Hedwig/Schema/TemporalGraphType.cs
```diff
@@ -0,0 +1,23 @@
+using GraphQL.Types;
+using System;
+
+namespace Hedwig.Schema
+{
+    public class TemporalGraphType<T> : ObjectGraphType<T>
+    {
+        public const String AS_OF_KEY = "asOf";
```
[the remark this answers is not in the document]
jlhogan: made it public for testing :grimacing: not sure what C# standard is for accessing private members in tests, will find out

src/Hedwig/Data/HedwigContext.cs
```diff
@@ -17,6 +18,42 @@ public HedwigContext(DbContextOptions<HedwigContext> options) : base(options)
 		public DbSet<SitePermission> SitePermissions { get; set; }
 		public DbSet<User> Users { get; set; }
 
+		public DbSet<T> GetDbSetForType<T>() where T : class
+		{
+			if(typeof(T) == typeof(Child)) {
```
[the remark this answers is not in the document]
jlhogan: you can't use `typeof(T)` as the match expression in a switch statement but also i deleted this :)

src/Hedwig/Schema/TemporalGraphType.cs
jlhogan: I believe the lifetime of the user context is a single query. We'll need to re-add auth to the context on each query. It's called `UserContext` only because that's what it is called in the graphQL world, but definitely down to rename. Will add those tests as well!

jlhogan:
> My only substantial concern is the global arguments and user context stuff. Can you provide some context to this or a reference link?
Sadly, not really. Most of what I have internalized about UserContext I got from reading source code and GitHub issues. I can add some comments to our source that explain at least a bit what is going on! 

> More minor: I also wonder if this is better broken up into smaller PRs - there is a lot going on and if were ever need to revert, it would be nice not to have to undo everything. I'm not super committed to this suggestion but did want to raise it.
Not really sure  that there are parts of this that we'd want to roll back without other parts, if that makes sense. Its a big honking PR, and definitely a large unit of work, but to me still feels compellingly like a single unit.

> Even more minor: The spacing/tabbing is inconsistent - I wonder if there is a linter we can use to fix all of this?
i truly am flummoxed by how much trouble I'm having figuring this out. I can't bring myself to deal with it right now, but it's on the mental todo and we can prioritize. It should get a real-life todo aka ticket :)

src/Hedwig/Migrations/20190909160144_InitialCreate.cs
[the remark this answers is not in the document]
jlhogan: i would +1 wiping all migrations and starting from scratch, but not becasue we _need_ to do this in the initial. More b/c i dont think there's really any value to maintaining migration history during dev as the schema gets worked out. When we're deploying to a fixed staging/QA place where the DB is persisted, then we obviously need to leverage migrations. But migrations before that point are just extra code IMO

Issue #81, opened 2019-10-09
Technical Debt - SonarCloud - Switch Hedwig Publish to SonarCloud Organization


speedcm:
As a project manager, I must be able to see the overall technical debt for the hedwig project. 

The Azure DevOps Services build pipeline performs static code analysis during the build. The results of the analysis are currently published to SonarCloud under a private organization.  

A new SonarCloud organization was setup with Issue #72 for ctoec.  As a result, the analysis should now post to this organization instead.  

What uniquely identifies a project in SonarCloud is the applications GUID.  Since we are migrating to a new SonarCloud project location with the existing SonarCloud project still present, the Hedwig application GUID must be updated to a new unique GUID.  
 
The azure-pipelines.yml file must be updated to point to the new SonarCloud organization and project.

speedcm:
This issue is now complete.  The builds are now publishing to the new SonarCloud organization accessible via:
https://sonarcloud.io/dashboard?id=ecis-experimental

PR #83, opened 2019-10-09
sonarcloud organization publish (#81)

speedcm:


speedcm:
SonarCloud requires the master branch to be analyzed before it can be analyzed on a PR branch.  For pull request 83 we get the following error:
ERROR: Project was never analyzed. A regular analysis is required before analyzing a pull request

Can we approve this pull request even though the status checks fail?  We need to build this on master first

PR #87, opened 2019-10-10
79: Refactors testing to create and cleanup data per-test

jlhogan:
- Exposes Context in addition to Client for integration tests for easier
seeding
- Creates custom TestHedwigContext to track and clean up created
entities per-test
- Create other test fixtures to enable use of TestHedwigContext in
integration tests
- Moves responsibility of DB migration out of tests and into pre-test
set up (see entrypoint-test.sh); this is due to the complexity of trying
to migrate the real sqlserver DB one time before any tests start running
- suppresses logs from db container in test docker-compose project to
make output more useful/less noisy

TODO:
    - leverage a test data creation package to replace the custom
*Helper classes currently used for object creation
    - determine if tests need to clean up history schema as well


test/HedwigTests/Fixtures/TestClientProvider.cs
```diff
@@ -5,16 +5,15 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Configuration;
 using Hedwig.Data;
-using Hedwig;
 
-namespace HedwigTests
+namespace HedwigTests.Fixtures
 {
-	public class TestClientProvider : IDisposable
+	public class TestWebHostProvider : IDisposable
```
cmgiven: Maybe TestApiProvider?

jlhogan: agreed. this was a stretch :)

Issue #91, opened 2019-10-14
Sites are owned by Organizations


cmgiven:
Although the MVP will just include single-site organizations (née programs, but switching to calling "Organization" to avoid needing to specify that "Program" is not the .NET global), this distinction is important for reports, as CDC reports are associated with an organization, not a site.

PR #92, opened 2019-10-14
Add Organization schema

cmgiven:
connects #91 

Sites are grouped into organizations. A user might have permissions at the per-site level, or the organization level.

Doing this now because the initial type of report is at the organization level, and it seemed like this would be harder to retrofit after the fact.

Notably there's now some inheritance going on in the Permissions models. I'm expecting to do something similar with Report. I was looking for how EF handles polymorphic relationships, and this seems to be the preferred approach (yes, not technically polymorphic). Docs here: https://docs.microsoft.com/en-us/ef/core/modeling/relational/inheritance


test/HedwigTests/Integration/GraphQLQueries/UserQueryTests.cs
```diff
@@ -9,17 +9,18 @@
 
 namespace HedwigTests.Integration.GraphQLQueries
 {
-    [Collection("SqlServer")]
-    public class UserQueryTests
-    {
+	[Collection("SqlServer")]
```
jlhogan: this collection annotation should have been removed in my #79 PR; can you delete it for me ?

cmgiven:
It's definitely the case that some users have permissions only to some sites. I.e. I'm responsible for maintaining the roster for a particular site, but I can't update the other sites. I did debate with myself whether that eventuality was too far in the future to worry about here, but landed on the side of it being useful to play that out.

cmgiven:
@speedcm Can you check SonarCloud for me? The code coverage doesn't appear to be working. The new lines in SiteRepository are definitely covered. But also it appears to be looking at code coverage of migrations and seed data, which should be ignored.

test/HedwigTests/Integration/GraphQLQueries/UserQueryTests.cs
cmgiven: Can do.

speedcm:
To exclude directory or directories, add the extraProperties element (see sample below). 
 
 - task: SonarCloudPrepare@1
    inputs:
      SonarCloud: "sonar-cloud-ctoec"
      organization: "ctoec"
      scannerMode: "MSBuild"
      projectKey: "ecis-experimental"
      extraProperties: |
         sonar.exclusions="**/Migrations/*"

https://community.sonarsource.com/t/analysis-running-in-azure-pipeline-ignoring-file-exclusions/7416

cmgiven:
And how about this one? https://sonarcloud.io/component_measures?id=ecis-experimental&metric=new_lines_to_cover&pullRequest=92&selected=ecis-experimental%3ARepositories%2FSiteRepository.cs&view=list

speedcm:
In order to update SonarCloud for Code Coverage, we need to specify the reports path.  Can you add the following line under extraProperties. 

sonar.cs.vstest.reportsPaths="**\TestResults\*.trx"

Please adjust the path accordingly.

cmgiven:
Can you clarify? Do you want exactly that line in the SonarCloud config? Or does this require more thought? If the latter, I'd suggest we create a new ticket for that, and disable code coverage checks until we have that figured out. @speedcm

speedcm:
I also agree the quality check should be disabled.  I will create the ticket.  Looks like you will have to disable the quality gate in SonarCloud since you are the administrator.

Issue #11, opened 2019-09-15
Reports page lists reports


cmgiven:
**Story:** As a provider, I want to see the reports that my programs/sites have submitted to OEC.

**Acceptance criteria:**

* When I click on the 'Reports' tab, I see a table with all of the reports I have submitted to OEC, as well as any reports I still need to submit (that are due).
* Some reports are owned by a `Site`. Others are owned by a `Program`, a grouping of `Sites` (that's not currently in the schema). We currently blur the distinction between the two because the initial set of users for our MVP is expected to only include single-site programs.
  * The initial type of report, a "CDC" report, is owned by a `Program`.
* Each report has a "Period" the time frame it reports on. These time frames are snapped to weeks, so the September 2019 CDC report might have the range of 9/1/2019 – 9/28/2019. Although this period might overlap the month before and/or after, as does the October report, it is still described as the "October 2019" report.
  * A schedule can be obtained from OEC, or we can just use dummy data.
* The report has a type. The initial type is "CDC", for Child Day Care contracts. We anticipate adding a School Readiness or "SR" type after MVP.
* If the report has not been submitted, it would have a due date (this should be stored in the database).
* If the report has been submitted, it would have a date submitted.
* New reports will ultimately need to be generated automatically every month. This will be a separate issue.

**Design:**

![Reports V2 Copy](https://user-images.githubusercontent.com/5375557/65296262-800d2d00-db29-11e9-8c11-ba3b66825db6.png)

**Related:** #12, #13

PR #98, opened 2019-10-15
Add reports to API

cmgiven:
Breaking out the API portion of #11, because this is already a massive PR.


src/Hedwig/Schema/Types/ReportType.cs
```diff
@@ -0,0 +1,12 @@
+using GraphQL.Types;
+
+namespace Hedwig.Schema.Types
+{
+	public class ReportType : UnionGraphType
+	{
+		public ReportType()
+		{
+			Type<CdcReportType>();
```
jlhogan: how will this work when other report types are added?

src/Hedwig/Schema/Types/CdcReportType.cs
```diff
@@ -0,0 +1,34 @@
...
+using Hedwig.Repositories;
+using GraphQL;
+using GraphQL.DataLoader;
+using GraphQL.Types;
+
+namespace Hedwig.Schema.Types
+{
+	public class CdcReportType : ObjectGraphType<CdcReport>
+	{
+		public CdcReportType(IDataLoaderContextAccessor dataLoader, IOrganizationRepository organizations)
```
jlhogan: just want to confirm: This is just report meta right now? And actual report context has not been implemented yet?

src/Hedwig/Schema/Types/ReportType.cs
cmgiven: This is a union type ("documentation" here: https://graphql-dotnet.github.io/docs/getting-started/unions). What it allows us to do is just grab a bundle of reports of different types and GraphQL will serialize them using the correct ObjectGraphType and pack them into a single array in the API response. So when we just need to list all of the various types here.

Eventually, we may want to add an [interface](https://graphql-dotnet.github.io/docs/getting-started/interfaces), which would allow us to pull out those attributes that are common to all `Reports` and request those attributes for everything at once, as opposed to per-type (see how that looks here: https://github.com/ctoec/ecis-experimental/pull/100/files#diff-f0d4ca9417637497cdcc62bc3923c240R14). But it worked fine without the interface, and I didn't want to figure out how to do that, so we'll cross that bridge when we add a second type of report.

src/Hedwig/Schema/Types/CdcReportType.cs
cmgiven: That's right. There shouldn't be much. Current designs show just a checkbox and two currency fields. As we walk through the ancestry of this class:

* Report (where most of the stuff is, attributes common to all reports)
* OrganizationReport (defines that this report is owned by an Organization)
* CdcReportType (doesn't currently have any attributes, but it does define that the value in the [discriminator column](https://docs.microsoft.com/en-us/ef/core/modeling/relational/inheritance#fluent-api) should be CDC).

cmgiven:
Closing, as I'll be rebasing and merging from #100.

PR #103, opened 2019-10-15
Add TemporalEntity type with AuthoredBy field

### LINKED ISSUE #82: Add authoredBy to temporal tables

jlhogan:
Also, codify some test running environment var flags and improve docker-compose testing experience and documentation in README

closes #82


src/Hedwig/Models/TemporalModel.cs
```diff
@@ -0,0 +1,8 @@
+namespace Hedwig.Models
+{
+
+    public abstract class TemporalEntity
```
cmgiven: Filename mismatch. I'm not sure which name is better.

PR #113, opened 2019-10-17
Add graphql frontend auth (97)

cailyncodes:
Include access token as a Bearer Token in all graphql requests.

Create a wrapper around `useQuery` hook so queries requiring authentication will instead use `useAuthQuery` -- with the same API.

Update queries from `user(id: 1)` to `me`.

Initial page load will show no data and "Log in" in the header. Clicking on login will redirect user to IS and then redirect back to roster page. After some time (initial load is slow) the data will load and "Log in" will be replaced with "Hi, Chris". Currently, no loading sign exists for after redirect.

### DIFF
--- a/src/Hedwig/ClientApp/package.json
+++ b/src/Hedwig/ClientApp/package.json
@@ -51,7 +51,7 @@
 		"eject": "react-scripts eject",
 		"storybook": "start-storybook -p 9009 -s public",
 		"build-storybook": "build-storybook -s public",
-		"apollo-generate": "NODE_TLS_REJECT_UNAUTHORIZED=0 apollo codegen:generate --excludes=node_modules/* --includes=**/*.tsx --endpoint https://localhost:5001/graphql --target typescript --tagName gql --outputFlat src/generated --passthroughCustomScalars --customScalarsPrefix OEC"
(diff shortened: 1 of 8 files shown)


jlhogan:
>  After some time (initial load is slow) the data will load and "Log in" will be replaced with "Hi, Chris". Currently, no loading sign exists for after redirect.

I'm not finding this to be the case! Loads up just fine for me :) 

Wondering about the permission setting the user is required to do when authenticating thru IS -- is that something that we/mike can customize? explicitly having to allow access to the backend service seems unnecessary and confusing for a user. Obviously as POC for us, this is AMAZING!!!!!!!!! but just wondering if you know how extensible that part is.

cailyncodes:
Yeah I agree it's definitely not the best UX. The page is customizable but I'm not sure about permissions on the consent screen -- only the UI is changeable I think. That being said, we might be able to configure the API and IS with a direct client id/secret registration. I'm only like 60% sure that something like that will work. It might be a limitation of OAuth2, however.

More detail: The access token we get back IS only has a `sub` claim when an API is requested in its scope. Hence the `hedwig_backend` claim being requested. It might be possible to get the `sub` claim from the API directly from IS however that would result in another network round trip on every request b/c we build the user context on each request instead of storing it in a session. I am anti-session broadly but if the UX causes us to need to use sessions I'm not super strongly opinionated. Alternatively, we might be able to use the identity token in conjunction with the access token to get the sub claim but then we should add JWE and we approach some less than best practices relying that heavily on the identity token (generally only name/user-preference claims should be respected b/c they run the risk of client impersonation)

cailyncodes:
I'm going to close out this PR but create an issue to explore consent screen UX.

Issue #110, opened 2019-10-17
Azure DevOps - winged-keys - Setup build and release pipeline


speedcm:
The Hedwig product uses winged-keys for user authentication.  Winged-keys is  a developed C# product which uses an open source framework and is built with dotnetcore 3.0.  Winged-keys has its own Azure DevOps project. 

The goal of this issue is to setup winged-keys in Azure DevOps project with a build and release pipeline.  The release pipeline should deploy to AWS Beanstalk.

speedcm:
This issue is now complete.  Winged keys has its own Azure DevOps Project with its own build and deploy pipelines.  Winged keys is deployed to AWS Beanstalk. 

Winged keys is not executed in AWS Beanstalk because the code base is a) hard wired to local system and b) not configurable to run with without SSL.   These issues are being addressed in separate issues in the winged keys GitHub project.

PR #116, opened 2019-10-17
setup winged-keys build pipeline (#110)

speedcm:


speedcm:
We will be using this PR branch to perform testing for some time.  This PR does not need to be approved right away.

PR #120, opened 2019-10-18
Modify GraphQL.Authorization to support iptables-like permissions

cailyncodes:
Due to the limitations of GraphQL.Authorization, this PR pulls several key files into our code base and modifies them to support an authorization flow similar to iptables.

Files that are pulled in are denoted with the GraphQL.Authorization's license at the start of the file with a link to the fixed Github file.

This needs tests, but I wanted to get it out there to see if the permissions flow makes sense.

How it would work in practice:

In a `HedwigGraphType` class, such as `UserType` is must implement the `IAuthorizedGraphType` interface and implement the `Permissions` method like so
```cs
public AuthorizationRules Permissions(AuthorizationRules rules)
{
	rules.DenyNot("IsAuthenticatedUserPolicy");
	rules.Allow();
	return rules;
}
```


jlhogan:
would it be beneficial to outline lightly what has changed in the GraphQL.Authorization files you've pulled down?

PR #105, opened 2019-10-16
Add graphql authorization to API (66)

cailyncodes:
Integrates the GraphQL.Authorization library into Hedwig.

Creates `AuthenticatedUserRequirement` and `IsAuthenticatedPolicy` but does not apply them to any endpoints (because no frontend code exists just yet to forward the bearer token).

Adds `me` GraphQL query.

Should resolve #66.

PR #135, opened 2019-10-21
Add API endpoint protection (106)

### LINKED ISSUE #106: Protect API endpoints
Building from #66: The authorization policies need to be applied to the endpoints accordingly. Separated from other issue because no frontend code existed for including access token in requests, which would have broken all the frontend.

**User Story:** As a user, I want to ensure that I can view sensitive data if and only if I have authenticated myself and am authorized to view said data.

**AC:**
* All endpoints should require an authenticated user
* All endpoints should require a user authorized to use Hedwig
* The user endpoint should only allow oneself to view data
* [Additional roles and permissions to be decided later]
* In development, staging, and test, developers should be able to bypass all security restrictions

cailyncodes:
Add initial protection to API endpoints.

Several new requirements are added:
* `CurrentUserRequirement` --> the user query id is the authenticated user id
* `DeveloperUserRequirement` --> the user has the "developer" role claim
* `DevelopmentRequirement` --> the current environment is development

Pulls down more of the `GraphQL.Authorization` library; needed to allow requirements to inspect the arguments in queries.

Relies on more DI to support injecting `HedwigContext` and `IHostingEnvironment` objects into specific requirements, instead of carrying them along on the `RequestContext` object.

Should resolve #106

### DIFF
--- a/auth-dev/OpenIDProvider/Config.cs
+++ b/auth-dev/OpenIDProvider/Config.cs
@@ -24,7 +24,9 @@ namespace OpenIDProvider
 					Claims = new []
 					{
 						new Claim("name", "Voldemort"),
-						new Claim("allowed_apps", "hedwig")
+						new Claim("allowed_apps", "hedwig"),
+						new Claim("role", "developer"),
+						new Claim("role", "oec-admin"),
 					}
 				}
 			};
@@ -39,11 +41,30 @@ namespace OpenIDProvider
 			};
 	}
 
+	// http://docs.identityserver.io/en/latest/topics/resources.html#defining-api-resources
 	public static IEnumerable<ApiResource> GetApis()
 	{
 	  return new List<ApiResource>
 			{
-				new ApiResource("hedwig_backend", "Hedwig API")
+				new ApiResource
+            {
+                Name = "hedwig_backend",
+
+                Description = "Hedwig API",
+
+                // include the following using claims in access token (in addition to subject id)
+                UserClaims = { "role" },
+
+								Scopes =
+                {
+                    new Scope()
+                    {
+                        Name = "hedwig_backend",
+                        DisplayName = "Full access to Hedwig API",
+                        UserClaims = new [] { "role", "allowed_apps" }
+                    }
+                }
+            }
 			};
 	}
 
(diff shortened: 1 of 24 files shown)


cailyncodes:
Outstanding (but depending on a larger conversation on roles/permissions) is additional protections/permissions on Child and Enrollment endpoints.

cailyncodes:
Tests resolved! When testing, we bypass the JWT validation and just assume the request user (via `TestClientProvider`) is authenticated. We are still using the authorization rules which will allow testing authorization rules/permissions. A future PR will add those tests -- I know I said I would do it on #105 but I think it will be cleaner to do it after this PR lands than try to add the new test flow portion from this PR to that one.

src/Hedwig/Schema/Types/ChildType.cs
```diff
@@ -41,5 +42,14 @@ public ChildType(IDataLoaderContextAccessor dataLoader, IFamilyRepository famili
 				}
 			);
 		}
+
+		public AuthorizationRules Permissions(AuthorizationRules rules)
+		{
+			rules.DenyNot("IsAuthenticatedUserPolicy");
```
jlhogan: this rules. so semantically UNDERSTANDABLE!

auth-dev/OpenIDProvider/Config.cs
```diff
@@ -43,7 +45,25 @@ public static IEnumerable<ApiResource> GetApis()
 	{
 	  return new List<ApiResource>
 			{
-				new ApiResource("hedwig_backend", "Hedwig API")
+				new ApiResource
```
jlhogan: can you link openID doc that explains this resource defn?

src/Hedwig/Security/AuthorizationContext.cs
```diff
@@ -0,0 +1,29 @@
+using System.Collections.Generic;
```
jlhogan: should this also have the graphql.authorization license comment ?

cailyncodes: Oops yep - ty!!

auth-dev/OpenIDProvider/Config.cs
cailyncodes: Will do!

cailyncodes:
> 
> 
> the public API looks great, but the weeds still feel confusing to me. can you provide a plain language high level run down of how the authentication parts fit together; i.e. heirarchy/relationship of policies, rules, requirements, settings?

Do you want this with in-code comments or on this PR convo?

Issue #13, opened 2019-09-15
Reports can be submitted


cmgiven:
**Story:** As a provider, I want to submit my CDC reports to OEC, informing OEC of whether my program has maintained its NAEYC accreditation.

**Acceptance criteria:**

* Clicking into an unsubmitted report on the Reports page (#11) opens the report.
* An "Accredited" checkbox records whether my program was accredited for the reporting period.
* Clicking a "Submit" button will mark the report as submitted to OEC. In addition to recording the time the report was submitted, it should record that I (my user) was the one to submit it.
* The checkbox on a submitted report cannot be changed. The submit button is no longer shown.

**Design:**

![Unsubmitted](https://user-images.githubusercontent.com/5375557/65296350-c19dd800-db29-11e9-9240-b9026c57fa10.png)

![Submitted](https://user-images.githubusercontent.com/5375557/65296455-15102600-db2a-11e9-8d1d-e05de6251774.png)

**Dependencies:** #11

jlhogan:
@cmgiven does the AC here imply the creation of an accredited bool on the cdc report model ?

cmgiven:
It does. Although I really only have that in there as a MacGuffin so we're not submitting a report about nothing at all. Up to you whether you want to include that or just cut it from the AC.

Issue #132, opened 2019-10-21
Checklist component


cmgiven:
**Story:** As a developer, I want to be able to build forms with Checklist controls, per the [USWDS](https://designsystem.digital.gov/components/form-controls/#checklist).

The [CMS Design System](https://design.cms.gov), which is implemented in React, may be a useful reference.

**AC:**

* A React component, matching the USWDS, is available in the Storybook.
* The component can have one or more checkboxes.
* The component can be disabled, preventing changing the selected options. Notably, a disabled checkbox could still be checked.
* If the selected option is invalid, a warning state highlights the offending control, and a warning message informs the user what needs to be done to correct the problem.

<img width="261" alt="Screen Shot 2019-10-21 at 8 15 40 AM" src="https://user-images.githubusercontent.com/5375557/67209882-c3cfac80-f3dd-11e9-9585-8d310234487d.png">
<img width="295" alt="Screen Shot 2019-10-21 at 8 20 20 AM" src="https://user-images.githubusercontent.com/5375557/67209889-c6320680-f3dd-11e9-94c1-9812ebf6848c.png">
<img width="345" alt="Screen Shot 2019-10-21 at 8 13 04 AM" src="https://user-images.githubusercontent.com/5375557/67209894-c8946080-f3dd-11e9-8e0c-16f9e3fcb955.png">

cmgiven:
There is a checkbox on the ReportSubmitForm component that should be replaced with this component when ready.

PR #136, opened 2019-10-21
15: Reports can be submitted

jlhogan:
Looking for feedback specifically on:
- how /where the form component for the report detail page should live. Feels business-logic specific, and thus container-y, but I don't know what's best practice
- how to test the conditional display logic in the form and the container

ALSO: the check icon in the check boxes is not showing up, altho it should be coming from the uswds styling. I presume theres some missing import or something, but this is really not my area of expertise so I was wondering if more adept eyes might be able to ID what's wrong. Generally there might need to be some styling aid to make it clear that the box is checked/unchecked in disabled state


src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const [submittedAt, setSubmittedAt] = useState(report.submittedAt);
+
+    const reportInput: CdcReportInput = {
+        id: report.id,
+        reportingPeriodId: report.reportingPeriodId,
+        organizationId: report.organization.id,
+        submittedAt: new Date().toUTCString(),
+        accredited: accredited
+    };
+    const [updateReport] = useMutation<ReportUpdateMut>(REPORT_SUBMIT, {
```
cailyncodes: Ah I see there is a `useMutation` hook -- that makes a lot of sense

src/Hedwig/Schema/Types/CdcReportType.cs
```diff
@@ -13,11 +11,13 @@ public CdcReportType(IDataLoaderContextAccessor dataLoader, IOrganizationReposit
 		{
 			Field(r => r.Id);
 			Field(r => r.Type, type: typeof(NonNullGraphType<FundingSourceEnumType>));
+			Field(r => r.ReportingPeriodId);
 			Field(r => r.ReportingPeriod.Period);
 			Field(r => r.ReportingPeriod.PeriodStart);
 			Field(r => r.ReportingPeriod.PeriodEnd);
 			Field(r => r.ReportingPeriod.DueAt);
 			Field(r => r.SubmittedAt, nullable: true);
+			Field(r => r.Accredited, nullable: true);
```
cailyncodes: Is we are doing a nullable boolean, does it make sense to abstract this out into a enum?

src/Hedwig/Schema/Mutations/ReportMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Hedwig.Schema.Mutations
+{
+    public class CdcReportInputType : InputObjectGraphType
+    {
+        public CdcReportInputType()
+        {
+            Name = "CdcReportInput";
+            Field<NonNullGraphType<IntGraphType>>("id");
+            Field<NonNullGraphType<IntGraphType>>("reportingPeriodId");
+            Field<NonNullGraphType<IntGraphType>>("organizationId");
```
cmgiven: I would expect these fields to not be writeable. Particularly the `organizationId`, as that'll be what controls who's authorized to see the report. Is there a way to enforce that?

src/Hedwig/Schema/Mutations/ReportMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            Field<ReportType>(
+                "updatedCdcReport",
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<CdcReportInputType>> { Name = "reportInput" }
+                ),
+                resolve: context =>
+                {
+                    var reportInput = context.GetArgument<CdcReport>("reportInput");
+                    reportInput.SubmittedAt = System.DateTime.Now.ToUniversalTime();
```
cmgiven: We might want to support saving without submitting. It seems that would be easy to do while still guaranteeing that it always records the current time by checking whether the input value is null.

cmgiven: Another constraint, though, I don't think there should be a way to unsubmit a report. There's a separate ticket for that. So actually, maybe the above requires more thought. Ignore me.

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+function ReportDetailForm(report: ReportDetailQuery_report) {
+    const [accredited, setAccredited] = useState(!!report.accredited);
+    const [submitted, setSubmitted] = useState(!!report.submittedAt);
+    const [submittedAt, setSubmittedAt] = useState(report.submittedAt);
+
+    const reportInput: CdcReportInput = {
+        id: report.id,
+        reportingPeriodId: report.reportingPeriodId,
+        organizationId: report.organization.id,
```
cmgiven: The API would be a lot nicer if I didn't have to provide these IDs, even apart from the fact that they shouldn't be updateable. I'll look below for why that is the case in the update logic.

src/Hedwig/Schema/Mutations/ReportMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+            Field<DateTimeGraphType>("submittedAt");
+            Field<NonNullGraphType<BooleanGraphType>>("accredited");
+        }
+    }
+    public class ReportMutation : ObjectGraphType<Report>, IAppSubMutation
+    {
+        public ReportMutation(IReportRepository repository)
+        {
+            Field<ReportType>(
+                "updatedCdcReport",
```
cmgiven: Should this be a verb? `updateCdcReport`?

src/Hedwig/Schema/Mutations/ReportMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<CdcReportInputType>> { Name = "reportInput" }
```
cmgiven: Maybe `report` instead of `reportInput`?

src/Hedwig/Schema/Mutations/ReportMutation.cs
cmgiven: Although it does seem odd that the API is asking me to add a `submittedAt`, and then ignoring the result.

src/Hedwig/Schema/Mutations/ReportMutation.cs
```diff
@@ -0,0 +1,39 @@
...
+                ),
+                resolve: context =>
+                {
+                    var reportInput = context.GetArgument<CdcReport>("reportInput");
```
cmgiven: Instead of casting this immediately to a `CdcReport`, could we use the `id` to retrieve the record from the database and iterate each property individually by iterating over the `reportInput` properties? It seems like we'll probably need to do that anyway once we start authorizing these mutations.

src/Hedwig/Schema/Mutations/ReportMutation.cs
cmgiven: Okay, putting this all together. What if the mutation becomes `submitCdcReport`, we don't include `submittedAt` in `CdcReportInputType`, and then we add it here?

src/Hedwig/Schema/Mutations/ReportMutation.cs
cmgiven: Or not even iterating, right? We know which values are coming in from the mutation, nothing's optional. We just need to update them all.

src/Hedwig/Schema/Mutations/ReportMutation.cs
jlhogan: yup! i forgot I had done this in the backend. I think logically it makes sense for this timestamp to be generated in the backend. will fix

src/Hedwig/Schema/Mutations/ReportMutation.cs
jlhogan: so, i started with the verb, but then I had the mutation returning a small slice of the updated report, and found the client code to be less intuitive when handling a report object with a verb name. Not sure I know what my preference is. Open to whatever.

src/Hedwig/Schema/Mutations/ReportMutation.cs
jlhogan: they're required in the mutation input due to my decision not to implement partial updates. We want to ensure they cannot change tho. Not sure where the best place for this validation to live is.

cmgiven: My comments are all out of order. Sorry about that. But I think we probably should do partial updates? Suggestion here: https://github.com/ctoec/ecis-experimental/pull/136#discussion_r337547072

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
jlhogan: i agree, but it's the requirement if we're not doing partial updates. 

I can put something together that takes an id and couple fields from the clients, grabs the model from the database, matches the names of the fields from the input to the properties of the model, updates those specific fields. 

Does that seem like how we want to do this? 

went back and forth a lot, but I do see the merit in that in terms of simplifying the API. I ended up going this way b/c it was simpler to just leverage the model representations, and easy enough to have the frontend grab and pass back the full model data, but it's not my ideal solution

src/Hedwig/Models/Reports/CdcReport.cs
```diff
@@ -2,5 +2,6 @@ namespace Hedwig.Models
 {
 	public class CdcReport : OrganizationReport
 	{
+		public bool? Accredited { get; set; }
```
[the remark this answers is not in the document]
jlhogan: this is simply b/c I added the column to the existing data, and didn't add the DDL in the migration to update all existing columns to be false. I agree that this shouldn't be nullable. Will do that.

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+        {submitted &&
+        <p><b>Submitted At:</b> {submittedAt} </p>
+        }
+        <div className="usa-checkbox">
+            <input 
+                className="usa-checkbox__input" 
+                id="accredited" 
+                type="checkbox" 
+                disabled={submitted} 
+                defaultChecked={!!accredited}
```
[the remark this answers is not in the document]
jlhogan: I think because accredited was nullable, typescript was mad at me. But this will go away when I fixup the field to be boolean not nullable.

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
```diff
@@ -0,0 +1,114 @@
...
+        updatedCdcReport(reportInput: $reportInput) {
+            ...on CdcReportType {
+                id,
+                submittedAt
+            }
+        }
+    }
+`;
+
+function ReportDetailForm(report: ReportDetailQuery_report) {
```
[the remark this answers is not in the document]
jlhogan: in the `components/` dir? wasn't sure, because my reading indicated that components should be "dumb" and not have business logic in them, and that stuff should live in containers. This form is totally container-specific and not really reusable so i didn't know what SOP was.

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
cmgiven: I think so? The concern is that otherwise we'd be letting people scribble all over the database, or writing a lot of fragile stuff to try and prevent that. Making clear what can be mutated is better not just for API ergonomics, but also as a way of signaling on the backend: These are the things that can be changed. I will handle those changes explicitly. I don't have to worry about accidentally exposing a different organization because the update switched the organization of a report, or some other kind of attack.

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportDetail.tsx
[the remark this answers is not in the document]
cmgiven: I would keep it in the `ReportDetail` directory, because it is a dumb component, but it will only ever be used by `ReportDetail`. I think the things in `components` should be reusable, yes, so that's my way of squaring the circle.

src/Hedwig/ClientApp/src/routes.ts
```diff
@@ -32,7 +33,13 @@ export const routes: RouteConfig[] = [
 	},
 	{
 		path: '/reports',
-		component: Reports,
+		exact: true,
+		component: Reports
+	},
+	{
+		path: '/reports/:id',
```
jlhogan: @cmgiven I could not get the subroutes feature to work, i.e. the following never rendered the Report container
```
{
    path: '/reports',
    exact: true,
    component: Reports,
    routes: [{
        path: '/reports/:id',
        exact: false,
        component: Report
    }]
}
```

cmgiven: If I recall, the subroutes feature is intended when you're nesting components inside each other. So let's say we have a form component with multiple steps (spoiler alert: we're going to have this). Then I can match with the parent component, while also providing information about which child component is active at that moment.

Because this is rendering an entirely different view, even though the path is under `/reports/`, the way you've done it is correct.

src/Hedwig/ClientApp/src/containers/Report/ReportSubmitForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+        {submittedAt &&
+        <p><b>Submitted At:</b> {submittedAt} </p>
+        }
+        <div className="usa-checkbox">
+            <input 
+                className="usa-checkbox__input" 
+                id="accredited" 
+                type="checkbox" 
+                disabled={!!submittedAt} 
+                defaultChecked={accredited}
```
jlhogan: The checkbox has different styling for "checked" vs "unchecked", but this styling never displays the actual check mark. Due to this, in disabled state it is impossible to tell if box is checked or unchecked. This seems mostly just like a styling issue but I didn't want to spend more time on it. Hopefully someone with more exp here can figure it out.

cmgiven: There's a ticket to create a checklist component (#132). I'll add a note to replace this checkbox with that component, and we can solve that issue there.

src/Hedwig/ClientApp/src/routes.ts
jlhogan: AH! makes sense

jlhogan:
closes #13

Issue #133, opened 2019-10-21
Radiolist component


cmgiven:
**Story:** As a developer, I want to be able to build forms with Radio Buttom controls, per the [USWDS](https://designsystem.digital.gov/components/form-controls/#radio-buttons).

The [CMS Design System](https://design.cms.gov), which is implemented in React, may be a useful reference.

**AC:**

* A React component, matching the USWDS, is available in the Storybook.
* Only one option can be selected. Selecting an option will deselect the others.
* If the selected option is invalid, a warning state highlights the offending control, and a warning message informs the user what needs to be done to correct the problem.

<img width="265" alt="Screen Shot 2019-10-21 at 8 22 17 AM" src="https://user-images.githubusercontent.com/5375557/67210190-4193b800-f3de-11e9-8ac2-93daddd5a042.png">
<img width="292" alt="Screen Shot 2019-10-21 at 8 22 09 AM" src="https://user-images.githubusercontent.com/5375557/67210193-435d7b80-f3de-11e9-8818-61b6413d974e.png">

mmazanec22:
I think this was created incidentally in #141 and just needs to be modified to include an error message

PR #141, opened 2019-10-24
6 past enrollments

mmazanec22:


cmgiven:
Neat! Appearance notes on the date picker:

* The border should match the USWDS. `1px solid #5c5c5c`
* The USWDS focus style is being used, but so is the style to add a teal underline. I suggest cutting the latter.
* We don't want to require the 200-weight font. Let's again match USWDS, which is the same size and color as the body font.
* The popup is using a teal color. Let's use the theme colors. I'd suggest `$theme-color-primary` for the selected date and the keyboard shortcut icon, and `theme-color-primary-light` for the dates in the selected range.

I'm checking out the code now.

Sorry this turned into a bit of a massive one!

src/Hedwig/ClientApp/src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,102 @@
...
+					<div>
+						<Button
+							text="Update"
+							onClick={() => onSubmit(selectedRange)}
+							disabled={
+								selectedRange.startDate === dateRange.startDate &&
+								selectedRange.endDate === dateRange.endDate
+							}
+						/>
+						<Button text="Reset" onClick={this.resetState} />
```
cmgiven: If we want this to be a generalized component, I'd suggest that these buttons should not be included. They would be more appropriate in the container, as you did with the radio buttons.

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -79,11 +110,66 @@ export default function Roster() {
...
+						<DatePicker
+							byRange={byRange}
+							onSubmit={dateRange => handleDateRangeChange(dateRange)}
+							dateRange={currentDateRange}
+							onReset={newRange => {
+								handleToggleByRange(false);
+								handleDateRangeChange(newRange);
+							}}
+						/>
+					</React.Fragment>
```
cmgiven: This feels like a component to me. It'll only ever be used in this container, though, so our convention would be to have it live at `containers/Roster/DateSelectionForm.tsx` or something.

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -1,42 +1,73 @@
...
 import { Table, TableProps } from '../../components/Table/Table';
 import { RosterQuery, RosterQuery_me_sites_enrollments } from '../../generated/RosterQuery';
 import nameFormatter from '../../utils/nameFormatter';
 import dateFormatter from '../../utils/dateFormatter';
 import Tag from '../../components/Tag/Tag';
+import DatePicker from '../../components/DatePicker/DatePicker';
+import RadioGroup from '../../components/RadioGroup/RadioGroup';
+import Button from '../../components/Button/Button';
 
 export default function Roster() {
```
cmgiven: It would be nice to get some basic tests going here. Nothing too complicated, but just ensuring that the number of kids enrolled text and the roster itself get updated would be good.

src/Hedwig/Repositories/EnrollmentRepository.cs
```diff
@@ -50,7 +50,7 @@ public static IQueryable<Enrollment> FilterByDates(this IQueryable<Enrollment> q
 				return query.Where(e => (
 					(e.Exit != null && e.Entry <= to && e.Exit >= from)
 					||
-					(e.Exit == null && e.Entry >= from && e.Entry <= to)
+					(e.Exit == null && e.Entry <= to)
```
cmgiven: Yay for catching this!

src/Hedwig/ClientApp/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import RadioButton from './RadioButton';
+
+interface RadioButton {
+	selected: boolean | undefined;
```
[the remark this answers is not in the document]
mmazanec22: @cmgiven I started to change this in a separate branch to take care of #133.  What do you think about me killing that branch and bringing those changes into this PR, and closing #133 as well after this is merged?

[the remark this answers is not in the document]
cmgiven: That seems fine to me. So much of it is already done here. 👍

cmgiven:
Oh, a thing that occurred to me in another conversation. Is there a way to prevent selecting future dates? If that's a thing we get easily from the date picker, that'd be great to include. If not, I don't think it's worth spending a lot of time there.

mmazanec22:
@cmgiven I think I've made all the changes you requested.  Since it was kind of a lot, I would appreciate another look-- especially on the tests.

src/Hedwig/ClientApp/src/containers/Roster/Roster.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+				</BrowserRouter>
+			</MockedProvider>
+		);
+		await waitForUpdate(wrapper);
+		const introSpanText = wrapper.find('.usa-intro span').text();
+		expect(introSpanText).toBe('2 kids enrolled.');
+		wrapper.unmount();
+	});
+
+	it('updates the number of kids', async () => {
```
cmgiven: Would be good to check the table too. But thanks for adding the first tests on this file!

PR #142, opened 2019-10-24
[WIP] Validation

cmgiven:
Two ways this is used.

From any schema type, I can add the following to create a field with an array of validation issues for the record:

```C#
Field<NonNullGraphType<ListGraphType<NonNullGraphType<DataValidationIssueType>>>>(
  "validations",
  resolve: context => validator.Validate(context.Source)
);
```

Or, from a mutation resolver, I can run the following after I update the record to add errors for any blocking validations to the context. It returns a `bool` that I can use to determine whether or not to save the record. If I don't specify which properties I would like it to check, it will use the context's `Arguments` and only validate properties with the same name as an argument (imagine we add a new blocking mutation, making a whole bunch of records retroactively invalid; a screen that touches the same model but doesn't update or display the problematic field would be broken and there'd be no way to display to the user what to fix).

```C#
context.AddDataValidationErrors(validator, record);
```

### DIFF
--- a/src/Hedwig/ServiceExtensions.cs
+++ b/src/Hedwig/ServiceExtensions.cs
@@ -1,12 +1,14 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.EntityFrameworkCore;
 using Hedwig.Data;
+using Hedwig.Models;
 using Hedwig.Repositories;
 using Hedwig.Schema.Types;
 using Hedwig.Schema.Queries;
 using Hedwig.Schema.Mutations;
 using Hedwig.Schema;
 using Hedwig.Security;
+using Hedwig.Validations;
 using GraphQL;
 using GraphQL.Server;
 using GraphQL.Authorization;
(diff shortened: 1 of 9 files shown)


src/Hedwig/ServiceExtensions.cs
```diff
@@ -17,96 +19,105 @@
...
-			services.AddScoped<ReportType>();
-			services.AddScoped<SiteType>();
-			services.AddScoped<UserType>();
+    public static void ConfigureDataValidations(this IServiceCollection services)
+    {
+      // Enrollment
+      services.AddScoped<IDataValidator<Enrollment>, DataValidator<Enrollment>>();
+      services.AddTransient<IDataValidationRule<Enrollment>, EnrollmentEntryCannotBeBeforeBirthdate>();
+      services.AddTransient<IDataValidationRule<Enrollment>, EnrollmentExitCannotBeBeforeEntry>();
+    }
```
cmgiven: Sorry for all the tabs/spaces changes. I don't really care, it's just my text editor "fixes" it automatically. This is the new stuff here.

src/Hedwig/Startup.cs
```diff
@@ -12,62 +12,63 @@
...
-            services.ConfigureAuthentication();
-        }
+    // This method gets called by the runtime. Use this method to add services to the container.
+    public virtual void ConfigureServices(IServiceCollection services)
+    {
+      services.ConfigureSqlServer(Configuration.GetConnectionString("HEDWIG"));
+      services.ConfigureCors();
+      services.ConfigureSpa();
+      services.ConfigureRepositories();
+      services.ConfigureDataValidations();
```
cmgiven: This is the only change to this file.

src/Hedwig/Validation/Rules/EnrollmentEntryCannotBeBeforeBirthdate.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Hedwig.Validations
+{
+  class EnrollmentEntryCannotBeBeforeBirthdate : IDataValidationBlockingRule<Enrollment>
+  {
+    private const string _message = "Cannot be before the child's birthdate.";
+
+    public IEnumerable<DataValidationIssue> Validate(Enrollment enrollment)
+    {
+      if (enrollment.Entry < enrollment.Child.Birthdate)
```
cmgiven: Performance problems could lurk here, when a validation relies on another table.

src/Hedwig/Validation/ResolveFieldContextExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+  public static class ResolveFieldContextExtensions
+  {
+    public static bool AddDataValidationErrors<T>(this ResolveFieldContext<T> context, IDataValidator<T> validator, T obj, IEnumerable<string> props = null)
+    {
+      props = props ?? context.Arguments.Keys.Select(key => Char.ToUpperInvariant(key[0]) + key.Substring(1));
+      var errors = validator.BlockingErrorsForProperties(obj, props);
+      context.Errors.AddRange(errors);
+      return errors.Any();
+    }
```
cmgiven: This might be altogether too fancy. But if it is the general use case, it would be nice not to have this cluttering up all of the mutation resolvers.

src/Hedwig/Validation/DataValidator.cs
```diff
@@ -0,0 +1,48 @@
...
+    private const ConcurrentDictionary<T, IEnumerable<DataValidationIssue>> _blockingResults = new ConcurrentDictionary<T, IEnumerable<DataValidationIssue>>();
+
+    public DataValidator(IEnumerable<IDataValidationRule<T>> rules) => _rules = rules;
+
+    public IEnumerable<DataValidationIssue> Validate(T obj)
+    {
+      return FetchBlockingResults(obj).Concat(FetchNonBlockingResults(obj));
+    }
+
+    public IEnumerable<ExecutionError> BlockingErrorsForProperties(T obj, IEnumerable<string> props)
```
jlhogan: this is a function to filter all of the blocking validation errors by property (i.e. the field on the model to which the validation applies?), and convert them to an enumerable of ExecutionErrors?

Will all validation errors always apply to a single property or is there the potential for 'compound' validation rules about relationships of property values?

src/Hedwig/Validation/ResolveFieldContextExtensions.cs
jlhogan: i dont think so ! if we're gonna use it twice, it's worth modularizing IMO. This way we get added benefit of free documentation from smart function naming. 
I would add a comment explaining line 12... props is either a provided list of model properties to grab potential errors for OR automatically generated from the arguments supplied in the query? that makes the practice of writing mutations with all updates as individual query arguments a requirement, which I'm down with, but is there anywhere we can/should capture that? Besides just here in PR comments ?
also, can/shouldwe /is it worth it to figure out how to enforce this is only called in a mutation resolve field context? we have the mutation marker interface already out there? But not sure what that would entail.

src/Hedwig/Validation/DataValidator.cs
cmgiven: The DataValidationError only applies to a single property. However, a single DataValidationRule can return multiple errors, so it could apply to multiple properties, albeit only for a single model.

This does mean that we can't have an error that flags two different fields as being wrong. But I think that's okay?

src/Hedwig/Validation/ResolveFieldContextExtensions.cs
cmgiven: Yeah, agreed that we need to document a convention like this somewhere. Not quite sure what the best way is.

Oh, good point, that ResolveFieldContext also exists when you're resolving a query. Possibly there's something in the context that tells us what it's about? What's the mutation marker interface?

src/Hedwig/Validation/DataValidationIssue.cs
```diff
@@ -0,0 +1,13 @@
+namespace Hedwig.Validations
+{
+  public struct DataValidationIssue
+  {
+    public readonly string property, message;
```
cailyncodes: Nit: Should these be uppercase? I feel like that's the approach I've seen in dotnet world but not 100% sure?

src/Hedwig/Validation/DataValidator.cs
cailyncodes: I think this means that we just have to have a standard for saying which of the multiple fields is the "bad" field. In the enrollment example, `entry < exit` so we have a choice as to which one (`entry` or `exit`) is invalid if we can't flag both, but I don't see that as blocking.

src/Hedwig/Validation/ResolveFieldContextExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Hedwig.Validations
+{
+  public static class ResolveFieldContextExtensions
+  {
+    public static bool AddDataValidationErrors<T>(this ResolveFieldContext<T> context, IDataValidator<T> validator, T obj, IEnumerable<string> props = null)
+    {
+      props = props ?? context.Arguments.Keys.Select(key => Char.ToUpperInvariant(key[0]) + key.Substring(1));
+      var errors = validator.BlockingErrorsForProperties(obj, props);
+      context.Errors.AddRange(errors);
```
cailyncodes: Slightly (but also tentatively) concerned that this will affect authorization. For better or worse, the `AuthorizationEvaluator` determines whether a requirement "fails" by counting the number of errors beforehand and after. Relatedly, the authorization error messages might include these validation errors. I haven't finished going through this PR so it could be totally not an issue but did want to flag it.

src/Hedwig/Validation/ResolveFieldContextExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+      return errors.Any();
```
cailyncodes: Related to above, an authorization error might cause this to return true without there actually being a validation error.

src/Hedwig/Validation/DataValidationIssue.cs
jlhogan: +1, public things seems to be Uppercased

src/Hedwig/Validation/ResolveFieldContextExtensions.cs
jlhogan: authorization _should_ be processed first, right? Before we get to validation? Because if you're not authorized, who cares if the inputs are valid, and you probably shouldn't be seeing informative errors about fields anyway. 

If errors do exist from auth, we should short-circuit processing before reaching validation

cailyncodes: Authorization is implemented by using graphql's validation contexts 🤷‍♀

PR #159, opened 2019-10-24
Add authentication/authorization tests

cailyncodes:
Add tests of the authentication and authorization system, both unit and integration.

### DIFF
--- a/src/Hedwig/Security/AuthorizationEvaluator.cs
+++ b/src/Hedwig/Security/AuthorizationEvaluator.cs
@@ -93,7 +93,7 @@ namespace Hedwig.Security
 					else
 					{
 						// Final rule must succeed or fail
-						throw new Exception();
+						throw new Exception("Final rule must be Allow or Deny");
 					}
 				}
 				else
--- a/test/HedwigTests/Integration/GraphQLQueries/ChildQueryPermissionsTests.cs
+++ b/test/HedwigTests/Integration/GraphQLQueries/ChildQueryPermissionsTests.cs
@@ -1,0 +1,96 @@
+using System;
+using System.Threading.Tasks;
+using System.Linq;
+using System.Collections.Generic;
+using Xunit;
+using Hedwig.Data;
+using Hedwig.Models;
+using HedwigTests.Helpers;
+using HedwigTests.Fixtures;
+using GraphQL.Common.Response;
+
+namespace HedwigTests.Integration.GraphQLQueries
+{
+	public class ChildQueryPermissionsTests
+	{
+		[Fact]
+		public async Task When_No_Auth_Get_Child_By_Id_Fails()
+		{
+			using (var api = new TestApiProvider())
+			{
+				// If
+				var firstName = "FIRSTNAME";
+				var child = ChildHelper.CreateChild(api.Context, firstName: firstName);
+				api.Client.DefaultRequestHeaders.Add("no_auth", "true");
+
+				// When
+				var response = await api.Client.GetGraphQLAsync(
+					$@"{{
+						child (id: ""{child.Id}"" ) {{
+							firstName
+						}}
+					}}"
+				);
+
+				// Then
+				response.EnsureSuccessStatusCode();
+				GraphQLResponse res = await response.ParseGraphQLResponse();
+				Assert.NotNull(res.Errors);
+			}
+		}
+
+		[Fact]
+		public async Task When_Not_Developer_Get_Child_By_Id_Fails()
+		{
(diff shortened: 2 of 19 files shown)


cailyncodes:
These tests are passing locally. @speedcm what environment do we run the tests in? "Development" or something else?

jlhogan:
> These tests are passing locally. @speedcm what environment do we run the tests in? "Development" or something else?

Tests should be running in "development" environment -- that's the default for dotnet unless something else is specified; they're run by the docker-compose project: "docker-compose.test.yaml"; have you tried running that locally to isolate?

src/Hedwig/Security/AuthorizationEvaluator.cs
```diff
@@ -93,7 +93,7 @@ public AuthorizationEvaluator(AuthorizationSettings settings)
 					else
 					{
 						// Final rule must succeed or fail
-						throw new Exception();
+						throw new Exception("Final rule must by Allow or Deny");
```
jlhogan: nit: must *be*

test/HedwigTests/Security/AuthorizationRuleTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+using Moq;
+using Hedwig.Security;
+
+namespace HedwigTests.Security
+{
+    public class AuthorizationRuleTests
+    {
+        [Fact]
```
jlhogan: you can write Xunit tests as "theories" which are just unit tests for which you can configure a data provider, for instance with an iterable set of auth rules, and expected results. Not sure if you tried that out !

test/HedwigTests/Security/Policies/CurrentUserRequirementTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        public async void When_Current_User_Authorize_Produces_No_Error()
+        {
+          // If
+          var requirement = new CurrentUserRequirement();
+          var context = new AuthorizationContext();
+          var claims = new Dictionary<string, string>() {
+            { "sub", "1" }
+          };
+          context.User = AuthorizationRequirementHelper.CreatePrincipal("password", claims);
+          var nameNode = new NameNode("id");
```
jlhogan: if not going to abstract this into some helper( totally fair!), might be useful to throw a comment explaining that this is simulating a graphQL request with query argument {id: 1}

test/HedwigTests/Integration/GraphQLQueries/ChildQueryPermissionsTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+				// Then
+				response.EnsureSuccessStatusCode();
+				GraphQLResponse res = await response.ParseGraphQLResponse();
+				Assert.NotNull(res.Errors);
+			}
+		}
+
+		[Fact]
+		public async Task When_Not_Developer_Get_Child_By_Id_Fails()
```
jlhogan: does this mean right now the requirement for retrieve child data is developer role?

test/HedwigTests/Security/AuthorizationRuleTests.cs
cailyncodes: Oh did not know that. I'll have to check it out!

test/HedwigTests/Integration/GraphQLQueries/ChildQueryPermissionsTests.cs
cailyncodes: The team needs to have a discussion on roles and permissioning generally and so I just have it set as developer required for now, but should be changed once we hash out the requirements.

cailyncodes:
> Tests should be running in "development" environment -- that's the default for dotnet unless something else is specified; they're run by the docker-compose project: "docker-compose.test.yaml"; have you tried running that locally to isolate?

Thanks for the tip! I needed to explicitly set the environment to development in `docker-compose.test.yaml`. The way I stubbed out an extension method for testing which environment it is required an non-implicit string.

test/HedwigTests/Security/AuthorizationRuleTests.cs
cailyncodes: Okay this is super cool!

test/HedwigTests/Security/Policies/CurrentUserRequirementTests.cs
cailyncodes: Refactored!

Issue #51, opened 2019-09-27
Application Logging – Application should allow for logging 


speedcm:
As a member of Ops team, I should be able to examine log files to aid in determining issues with the run-time application.

The logs should be local to the system and well formatted to allow for forwarding to CloudWatch.   Logging should be enabled or disabled via an app setting.

speedcm:
This issue is now complete.  For more information see https://github.com/ctoec/ecis-experimental/wiki/Paper:-Application-Logging

PR #160, opened 2019-10-24
application logging (#51)

speedcm:


speedcm:
Great point! This issue for 51 and it will not be closed until there is a group discussion on it implementation. Here is the the wiki information on this topic, https://github.com/ctoec/ecis-experimental/wiki/Paper:-Application-Logging.  I will remove the hard code for us-east-2.

PR #161, opened 2019-10-25
Add enrollment mutation

### LINKED ISSUE #7: Enrollment data can be mutated
**Story:** As an API consumer, I want to be able to mutate an enrollment to change the entry and/or exit date.

**Acceptance criteria:**

* I should be able to make a GraphQL mutation like the following:

```
{
  updateEnrollment(id: 1) {
    id
    entry
    exit
  }
}
```

**Related:** #8

cailyncodes:
I have a question about the process for clearing an exit date (i.e. setting it to null). I'm not sure if this is a supported use case but needed to check in. /cc @cmgiven 

Should resolve #7

### DIFF
--- a/src/Hedwig/Repositories/EnrollmentRepository.cs
+++ b/src/Hedwig/Repositories/EnrollmentRepository.cs
@@ -29,6 +29,12 @@ namespace Hedwig.Repositories
 			return GetBaseQuery<Enrollment>(asOf)
 				.SingleOrDefaultAsync(e => e.Id == id);
 		}
+
+		public Enrollment UpdateEnrollment(Enrollment enrollment)
+		{
+			_context.SaveChanges();
+			return enrollment;
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


cmgiven:
I think clearing the exit date should be supported. But that's just a guess for now.

cailyncodes:
Are we comfortable requiring the exit parameter unless we are clearing exit date? That is, an exit date is cleared if and only if no exit parameter is supplied.

cmgiven:
I guess that feels a little unintuitive? I think I would expect:

No argument supplied: It doesn't change anything.
Argument supplied with a null value: It clears the date.

cailyncodes:
Yeah ~~I was concerned that you'd say that~~ I think that's a better flow -- I just need to figure out how to allow either a string representing a date or a string representing null to coexist as an argument value.

cmgiven:
Lol. :+1:

PR #163, opened 2019-10-25
115: Use HttpContext.User name as AuthoredBy for temporal entity writes

jlhogan:


src/Hedwig/Data/HedwigContext.cs
```diff
@@ -1,13 +1,19 @@
...
+using System;
 
 namespace Hedwig.Data
 {
 	public class HedwigContext : DbContext
 	{
-		public HedwigContext(DbContextOptions<HedwigContext> options) : base(options)
-		{ }
+		private readonly IHttpContextAccessor _httpContextAccessor;
+		private const string UNKNOWN_AUTHOR = "Unknown";
```
cailyncodes: I don't think we use this variable anywhere?

src/Hedwig/Data/HedwigContext.cs
```diff
@@ -91,9 +97,27 @@ private bool IsTemporalEntityType<T>()
 		/// author name is a place holder; in the future it will pull from session information.
```
cailyncodes: This can probably be updated.

src/Hedwig/Data/HedwigContext.cs
jlhogan: was from a prior pass

PR #165, opened 2019-10-25
96: Implement automatic token refresh

### LINKED ISSUE #96: Use refresh tokens to get updated access tokens
**User Story:** As a user, I want to be able to continue using the application without needing to sign in additional times every few minutes.

**AC:**
* The client retrieves a refresh token from authentication server
* The client periodically requests new access tokens from the authentications server with the refresh token
* The client updates the access token store with the new access token

jlhogan:
Extends the auth flow to automatically exchange tokens when they have expired using refresh tokens. openid/appauth lib handles exchanging tokens _before_ they have expired; the default buffer is 10 mins.

weird thing I cannot figure out: the availability of instance and state variable, will add further comments in line, but the gist is that a state variable i set using the provided mutator and a value-y value is undefined on the next line. very mysterious, and is resulting in some duplicated state in the component --  not idea but it works.  

closes #96


src/Hedwig/ClientApp/src/contexts/Login/LoginHOC.tsx
```diff
@@ -51,63 +50,77 @@ const LoginProvider: React.FC<LoginProviderPropsType> = (
   const location = useLocation();
   const [firstLoad, setFirstLoad] = useState(true);
   const [accessToken, setAccessToken] = useState<string | null>(null);
+  const [persistentConfiguration, setPersistentConfiguration] = useState<AuthorizationServiceConfiguration>();
```
jlhogan: this is the duplicated state.
both `makeAuthorizationCodeTokenRequest` and `makeRefreshTokenRequest` require an authorization service config. Both the instance variable `configuration` (defined below on line 57) and the state variable `persistentConfiguration` are set in `fetchServiceConfiguration` (lines 110-112). Directly after the call to `fetchServiceConfiguration` in the login callback, `persistentConfiguration` is undefined (despite having been set with a value-ful value). At that point, `configuration` is defined. Thus, in `makeAuthorizationCodeTokenResponse`, `persistentConfiguration` is undefined, and `configuration` contains the authorization service config object.
Later, when calling `makeRefreshTokenRequest` via context injection in `useAuthQuery`, the state variable `persistentConfiguration` is the service config object, and `configuration` is undefined.

I can't figure out how to manage to store this state in only one place that is available to both functions that require it. thoughts very very welcome. im confused.

src/Hedwig/ClientApp/src/contexts/Login/LoginHOC.tsx
```diff
@@ -51,63 +50,77 @@ const LoginProvider: React.FC<LoginProviderPropsType> = (
...
+   * This function should only need to be called once for the lifetime of the app
+   */
+  async function fetchServiceConfiguration(): Promise<void> {
+    if (persistentConfiguration) {
+      return;
     }
     return AuthorizationServiceConfiguration.fetchFromIssuer(openIdConnectUrl, new FetchRequestor())
       .then(resp => {
+        if(resp) {
         configuration = resp;
```
jlhogan: this is where configuration and persistentConfiguration are set. However console.log-ing persistentConfiguration right here on line 112 shows it has value of `undefined`

src/Hedwig/ClientApp/src/contexts/Login/LoginHOC.tsx
```diff
@@ -124,95 +137,89 @@ const LoginProvider: React.FC<LoginProviderPropsType> = (
...
-      return;
-    }
+  /**
+   * Create and perform authorization code-based token request.
+   * This type of token request is only required after initial auth request.
+   * Once a token is retrieve, the refresh token can be used to request subsquent
+   * tokens (until the refresh token expires)
+   */
+  async function makeAuthorizationCodeTokenRequest(code: string, verifier: string | undefined) {
+    if (!configuration) return;
```
jlhogan: `persistentConfiguration` is undefined here

src/Hedwig/ClientApp/src/contexts/Login/LoginHOC.tsx
```diff
@@ -124,95 +137,89 @@ const LoginProvider: React.FC<LoginProviderPropsType> = (
...
   }
 
-  function checkForAuthorizationResponse() {
-    authorizationHandler.completeAuthorizationRequestIfPossible();
+  /**
+   * Create and perform token refresh request.
+   * Requires that initial code-based token request has already been performed.
+   */
+  async function makeRefreshTokenRequest() {
+    if (!persistentConfiguration) return;
```
jlhogan: `configuration` is undefined here

src/Hedwig/ClientApp/src/hooks/useAuthQuery.ts
```diff
@@ -1,17 +1,21 @@
...
+  useEffect(() => {
+    withFreshToken();
+  });
+    
   return useQuery<T>(query, {
     ...options,
     context: {
       headers: {
-        Authorization: accessToken ? `Bearer ${accessToken}` : ''
+        authorization: accessToken ? `Bearer ${accessToken}` : ''
```
cailyncodes: Nit: HTTP headers are generally capitalized. It works with lowercase, though.

src/Hedwig/ClientApp/src/hooks/useAuthQuery.ts
```diff
@@ -1,17 +1,21 @@
...
 
 export default <T>(query: DocumentNode, options: QueryHookOptions = {}): QueryResult<T> => {
-  const { accessToken } = useContext(LoginContext);
+  const { accessToken, withFreshToken } = useContext(LoginContext);
 
+  useEffect(() => {
+    withFreshToken();
+  });
+    
   return useQuery<T>(query, {
```
cailyncodes: We might have to return a new `refetch` value from `useQuery` -- right now I don't think `refetch` will use with the updated accessToken -- only on a remount, which means the header probably won't be updated?

jlhogan: Setting the `AccessTokenLifetime` really low results in the header constantly flipping btwn `log in` and then back to `Hi Chris` as the accesstoken is exchanged. This indicates to me that when accessToken changes, refetch is called using the correct token! (b/c it is able to get the data and log in chris)

PR #195, opened 2019-10-31
169 Add create family with child mutation

### LINKED ISSUE #169: Family can be created
**Story**: As an API consumer, I want to be able to create a family

**Acceptance criteria**:
- I should be able to make an authenticated GraphQL mutation like the following:
```
{
createFamilyWithChild(
  addressLine1: String,
  addressLine2: String,
  town: String,
  state: String,
  zip: String,
  homelessness: Boolean,
  childId: Id!
  ){}
}
```
IFF my user has access to provided childId

- this mutation should create a family
- this mutation should associated the provided child with the created family 

**Related**: #168 (for discussion of how to handle create mutation inputs)

cailyncodes:
This adds the create family method to family and the update family method to child.

Removes saving changes to db context from repositories to a commit mutation middleware.

Cleans up async/await code.

Should resolve #169

### DIFF
--- a/src/Hedwig/Repositories/FamilyRepository.cs
+++ b/src/Hedwig/Repositories/FamilyRepository.cs
@@ -19,10 +19,39 @@ namespace Hedwig.Repositories
 				.ToDictionaryAsync(x => x.Id);
 			return dict as IDictionary<int, Family>;
 		}
+
+		public Family CreateFamily(
+			string addressLine1 = null,
+			string addressLine2 = null,
+			string town = null,
+			string state = null,
+			string zip = null,
+			bool homelessness = false)
+		{
+			var family = new Family {
+				AddressLine1 = addressLine1,
+				AddressLine2 = addressLine2,
+				Town = town,
+				State = state,
+				Zip = zip,
+				Homelessness = homelessness
+			};
+
+			_context.Add<Family>(family);
+			return family;
+		}
 	}
 
 	public interface IFamilyRepository
 	{
 		Task<IDictionary<int, Family>> GetFamiliesByIdsAsync(IEnumerable<int> ids, DateTime? asOf = null);
+		Family CreateFamily(
+			string addressLine1 = null,
+			string addressLine2 = null,
+			string town = null,
+			string state = null,
+			string zip = null,
+			bool homelessness = false
+		);
 	}
 }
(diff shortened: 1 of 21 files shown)


cailyncodes:
@jlhogan Not ready for a full review yet but penny for your thoughts on the middleware approach.

src/Hedwig/Schema/Mutations/FamilyMutation.cs
```diff
@@ -0,0 +1,67 @@
...
+
+					var family = repository.CreateFamily(
+						addressLine1,
+						addressLine2,
+						town,
+						state,
+						zip,
+						homelessness
+					);
+					Console.WriteLine(family);
```
cailyncodes: Oops -- @cailyncodes remove this

cailyncodes:
Ready for review.

`FamilyMutationTests` > `Create_Family_With_All_Fields` is failing, and I can't figure out how to fix it.

src/Hedwig/Schema/AppSchema.cs
```diff
@@ -1,11 +1,11 @@
-using GraphQL;
+using graphQL = GraphQL;
```
jlhogan: If we're going to add a GraphQL namespace to our project, should `Schema` live inside it? Not blocking for this PR (in fact, we need to move on this so def don't impl in this PR) but wanted to flag

src/Hedwig/Repositories/FamilyRepository.cs
```diff
@@ -19,10 +19,39 @@ public class FamilyRepository : TemporalRepository, IFamilyRepository
...
+		}
 	}
 
 	public interface IFamilyRepository
 	{
 		Task<IDictionary<int, Family>> GetFamiliesByIdsAsync(IEnumerable<int> ids, DateTime? asOf = null);
+		Family CreateFamily(
+			string addressLine1 = null,
+			string addressLine2 = null,
+			string state = null,
```
jlhogan: @cailyncodes in the interface defn, town and state are in a different order. even tho this _shouldn't matter at all_ because you're supplying the arguments as named, it seems to be the thing swapping the value of town and state !

cailyncodes: I'll try flipping the arguments. But yeah it really shouldn't matter but maybe 🤷‍♀

src/Hedwig/Schema/AppSchema.cs
cailyncodes: I would support adding `Schema` as a subnamespace to `GraphQL`. I'm also open to not having a `GraphQL` namespace; I just don't know where to put the files in that namespace -- def open to suggestions.

The reason this namespace renaming is required is due to `Hedwig.GraphQL` and `GraphQL` conflicting when resolving the namespace later in this file.

PR #198, opened 2019-11-01
187 Update Child model

### LINKED ISSUE #187: Update Child Model
**New Fields**:
- SASId: string | null
- birthCertificateId: string | null
- birthTown: string | null
- birthState: string | null
- foster: Boolean (default: false)

**Updated Fields**:
- birthdate: DateTime | null
- gender: Gender | null
- americanIndianOrAlaskaNative: boolean (default: false)
- asian: boolean (default: false)
- blackOrAfricanAmerican: boolean (default: false)
- nativeHawaiianOrPacificIslander: boolean (default: false)
- white: boolean (default: false)

cailyncodes:
Should close #187

### DIFF
--- a/src/Hedwig/Models/Child.cs
+++ b/src/Hedwig/Models/Child.cs
@@ -8,6 +8,8 @@ namespace Hedwig.Models
 	{
 		public Guid Id { get; set; }
 
+		public string Sasid { get; set; }
+
 		[Required]
 		[StringLength(35)]
 		public string FirstName { get; set; }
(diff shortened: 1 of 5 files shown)


src/Hedwig/Models/Child.cs
```diff
@@ -22,16 +24,24 @@ public class Child : TemporalEntity
...
-		public DateTime Birthdate { get; set; }
+		public DateTime? Birthdate { get; set; }
+
+		public string BirthTown { get; set; }
+
+		public string BirthState { get; set; }
+
+		public string BirthCertificateId { get; set; }
+
+		public bool AmericanIndianOrAlaskaNative { get; set; }
```
jlhogan: why not define default values here? even tho it didn't end up  resulting in correct generate SQL, it makes the model consistent and clear btwn DB and CLR type

cailyncodes: Oh I didn't know one could specify default values on the model. Does that just look like `public bool AmericanIndianOrAlaskaNative { get; set; } = false;` or with an annotation?

Issue #144, opened 2019-10-24
Enrollment landing page layout with step list


abbyraskin:
As a provider, I want a place to enter new enrollees into the system.

## Acceptance Criteria
- User navigates to this page by selecting `Enroll child` on Roster page; Roster is highlighted in subnav 
- When a user arrives on the page, the first step component section is in an open/active state
- Adds "Enroll child" header text
- Adds a program site dropdown (left-aligned under "Enroll child" header, note difference from design)
- Program site dropdown is pre-populated with the site selected while on the Roster page
- Adds 4 empty [step list component](https://design.cms.gov/patterns/step-list/) sections, including standard edit links
- Selecting "edit" opens the section into an active state (empty for now)
- Updates URL path on save
  - @cmgiven to decide on path names
- "Save" button per section affects front-end only for now, developed in #152 

### Dependencies
#143 - Navigation and URL path

### Related stories
#152 - Makes "Save" button functional
#146 - Child's information components
#147 - Family information components
#149 - Family income components
#150 - Enrollment and funding components
#145 - Back to roster link

### Design
[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens)

Issue #143, opened 2019-10-24
Add "enroll child" button to roster page and change URL path


abbyraskin:
As a provider, I want a way to enroll a new child in my program from the home page.

## Acceptance Criteria
- Add `Enroll child` button on the roster page, right-aligned (note this button already exists in commit history)
- Clicking on `Enroll child` button navigates user to blank enrollment landing page 
- Remove "enroll kids" from the subnavigation
- Change path to roster/enroll (Roster is highlighted in subnav when user is on enroll page)

## Design
![image](https://user-images.githubusercontent.com/30734760/67496513-20330600-f642-11e9-97b9-889e575d1c57.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens)

abbyraskin:
Note: this button exists in commit history

PR #200, opened 2019-11-01
Enrollment skeleton

cmgiven:
connects #143
connects #144 

The testing for this one is a bit weird.

The views of a child are broken up into sections. We mirror that here by creating modules each of which contain two views per section. The first view, Summary, shows an at-a-glance view of the kid, and is used on the child landing page, and during the enrollment flow to summarize completed sections. The second view, Form, is the editable view, used on the Edit pages and during the enrollment flow for the current section.

Form will handle all of the mutation directly. It does not pass data to its parent(s), except by mutating the Apollo cache.

Unfortunately, we don't have those mutations built out yet. It seems it's pretty tough to mock Apollo mutations. So enrolling a brand new child will error out on the second step of the enrollment flow once it realizes that a child record wasn't actually created by the first step.

The solution for this is to use a child that already exists, and run them through the enrollment flow a second time. As the authentication is not preserved across browser refreshes, the easiest way to do this is to edit Roster.tsx so that each row links to: `/roster/enrollments/${row.child.id}/new/child-information` instead of `/roster/enrollments/${row.child.id}/`.

At that point, clicking any child in the Roster will enter the enrollment flow. You can continue through that flow, click Finish, and arrive at the child landing page. Then click any of the Edit buttons to see the edit page.

Bit of a mongo PR, but it'll be quite exciting to put this together with the new mutations and some real form components.


cailyncodes:
> As the authentication is not preserved across browser refreshes
Are we just calling this out or would this be a needed business case to support refreshes?

cmgiven:
I think it's needed. It's not just refreshes. Right now I don't believe I could have two windows open without logging each one in? Or I'd expect to be able to close a window and reopen Hedwig and still be logged in, within a reasonable time period.

I'm guessing this is because the token is just stored in the React app's global state, and once we lose that state, we lose the token.

cailyncodes:
> I'm guessing this is because the token is just stored in the React app's global state, and once we lose that state, we lose the token.

Yes, we'll need to create a local storage or cookie value to keep on the browser. It would be best for security to only ever have it in the React memory allocation, but if it's a business requirement, we can definitely add it.

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -0,0 +1,92 @@
...
+	if (loading || error) {
+		return <div className="EnrollmentNew"></div>;
+	}
+
+	const afterSave = () => {
+		if (!data || !data.child) {
+			throw new Error('EnrollmentNew called afterSave before successful save');
+		}
+
+		if (!childId) {
```
jlhogan: comments to describe each of these conditions? i.e. 
what is the state of the form flow when `!childId` ?(enrollment has not been started yet)
 when currentIndex == sections.length -1 ? (step is final step)
else  (enrollment is underway and should proceed to next step)

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -0,0 +1,71 @@
...
+				<input
+					type="text"
+					name="lastName"
+					placeholder="Last"
+					onChange={event => updateLastName(event.target.value)}
+					value={lastName}
+				/>
+				<br />
+				<br />
+
```
jlhogan: this will get augmented with bools/enums when the checklist component lands?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+	status: () => 'complete',
+
+	Summary: ({ child }) => {
+		if (!child || !child.family) {
+			return <div></div>;
+		}
+
+		return (
+			<div className="FamilyInfoSummary">
+				<p>Care 4 Kids case number: {child.family.caseNumber ? child.family.caseNumber : ''}</p>
```
jlhogan: this will no longer exist on family once cailyns PRs land

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+			}
+		}
+	}
+`;
+
+export default function ReportSubmitForm(report: ReportQuery_report) {
+	const [accredited, setAccredited] = useState(report.accredited);
+	const [submittedAt, setSubmittedAt] = useState(report.submittedAt);
+
+	console.log(REPORT_SUBMIT_MUTATION);
```
jlhogan: nit: debug logging

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
cmgiven: Yup. Now we have a real TextInput component too, so I'll basically throw out all of the form and replace it.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo.tsx
cmgiven: Yup. It was the only thing I had access to on the existing family model. I will cut before merging to avoid conflicts.

src/Hedwig/ClientApp/src/containers/ReportDetail/ReportSubmitForm.tsx
cmgiven: I think that was you... I just renamed the file. :-p

Issue #131, opened 2019-10-21
Dropdown component


cmgiven:
**Story:** As a developer, I want to be able to build forms with Dropdown controls, per the [USWDS](https://designsystem.digital.gov/components/form-controls/#dropdown).

The [CMS Design System](https://design.cms.gov), which is implemented in React, may be a useful reference.

**AC:**

* A React component, matching the USWDS, is available in the Storybook.
* The control can be manipulated from the keyboard alone.
* If the selected option is invalid, a warning state highlights the offending control, and a warning message informs the user what needs to be done to correct the problem.

<img width="286" alt="Screen Shot 2019-10-21 at 8 09 18 AM" src="https://user-images.githubusercontent.com/5375557/67209691-70f5f500-f3dd-11e9-92c9-f472731358d2.png">

cailyncodes:
Closed by #201

PR #201, opened 2019-11-03
Create dropdown component (#131)

mmazanec22:
This component raised a few odd little problems:

- Needed to define theme image path and font path relative to USWDS files using them
- Due to a [bug](https://github.com/uswds/uswds/issues/3117), uswds-theme-components was not included by default, so those are now included in index.scss
- Defining the image path broke the hero image path, because it used image path by default but was used in a different place-- upgrade to USWDS 2.2.1 fixed this by allowing us to set hero-image-path

### DIFF
--- a/src/Hedwig/ClientApp/src/components/Dropdown/Dropdown.tsx
+++ b/src/Hedwig/ClientApp/src/components/Dropdown/Dropdown.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+
+type DropdownOption = {
+	value: string;
+	text: string;
+};
+
+type DropdownProps = {
+	options: DropdownOption[];
+	label: string;
+	unselectedOptionText: string;
+	onChange: () => any;
+	id?: string;
+	success?: boolean;
+	disabled?: boolean;
+	error?: boolean;
+	errorMessage?: string;
+};
+
+export default function Dropdown({
+	options,
+	label,
+	unselectedOptionText,
+	onChange,
+	id,
+	success,
+	disabled,
+	error,
+	errorMessage,
+}: DropdownProps) {
+	const selectId = id || label.split(' ').join('-');
+	const errorMessageId = `${selectId}-error-message`;
+
+	return (
+		<div className={`usa-form-group${error ? ' usa-form-group--error' : ''}`}>
+			<label className={`usa-label${error ? ' usa-label--error' : ''}`} htmlFor={selectId}>
+				{label}
+			</label>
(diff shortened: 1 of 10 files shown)


src/Hedwig/ClientApp/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+				}`}
+				name={selectId}
+				id={selectId}
+				onChange={onChange}
+				disabled={disabled}
+				aria-describedby={error ? errorMessageId : undefined}
+			>
+				<option value={undefined}>{unselectedOptionText}</option>
+				{options.map(option => (
+					<option value={option.value} key={option.value}>
```
cailyncodes: Nit: Do we want to use `option.value` as the key? I've generally seen the index in the array be used for the key, but am not 100% on this.

src/Hedwig/ClientApp/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,63 @@
...
+	label,
+	unselectedOptionText,
+	onChange,
+	id,
+	success,
+	disabled,
+	error,
+	errorMessage,
+}: DropdownProps) {
+	const selectId = id || label.split(' ').join('-');
```
cailyncodes: Is there a reason for constructing the `selectId` as so vs making `id` a required prop value?

src/Hedwig/ClientApp/src/components/Dropdown/Dropdown.tsx
mmazanec22: I think React yells at you if you use the index.  Option.value might not be the best either, though, considering the possibility for duplication.  Maybe a combination?

cailyncodes: I like the idea of a combination!

src/Hedwig/ClientApp/src/components/Dropdown/Dropdown.tsx
mmazanec22: Out of sheer laziness I like defaults, but can make this required if that seems more sensisble

cailyncodes: I'm not opposed to having a sensible default. I was just curious if there was motivating reason. I think you can leave this as is 😃

Issue #176, opened 2019-10-28
Family determination can be updated


jlhogan:
**Story**: As an API consumer, I want to be able to update an existing family determination

**Acceptance Criteria**:
- I should be able to make an authenticated GraphQL mutation like the following:
```
{
updateFamilyDetermination(
  id: Int!,
  numberOfPeople: Int,
  income: Number,
  determined: date
) {}
}
```
IFF my user has access to the provided determination id

- the mutation should update the provided determination with any supplied values, and leave other fields as they are

cailyncodes:
Closed by #204

PR #204, opened 2019-11-04
170: Add family determination mutations

### LINKED ISSUE #170: Family determination can be created
**Story**: As an API consumer, I want to be able to create a family determination

**Acceptance criteria**:
- I should be able to make an authenticated GraphQL mutation like the following:
```
{
createFamilyDetermination(
  numberOfPeople: Int!,
  income: Number!,
  determined: Date!,
  familyId: Int!) 
{}
}
```
IFF my user has access to the provided familyId

**Related**: #168 (for discussion of how to handle inputs for create mutations)

jlhogan:
I opted to put the null-checking logic into the repository `UpdateFamilyDetermination` function, as opposed to handling it in the mutation itself. This was easy enough to do because none of the familydetermination fields are nullable; if they have a value, use it to update, if not then nothing changes. Not sure it's what we want to do / should do / can do for objects with nullable fields, but it seemed the cleanest here. 

closes #170 #176

### DIFF
--- a/src/Hedwig/Repositories/FamilyDeterminationRepository.cs
+++ b/src/Hedwig/Repositories/FamilyDeterminationRepository.cs
@@ -20,10 +20,69 @@ namespace Hedwig.Repositories
 
 			return determinations.ToLookup(d => d.FamilyId);
 		}
+
+		public Task<FamilyDetermination> GetDeterminationByIdAsync(int id, DateTime? asOf = null)
+		{
+			return GetBaseQuery<FamilyDetermination>(asOf)
+				.FirstOrDefaultAsync(d => d.Id == id);
+		}
+
+		public FamilyDetermination CreateFamilyDetermination(
+			int numberOfPeople,
+			decimal income,
+			DateTime determined,
+			int familyId
+		)
+		{
+			var familyDetermination = new FamilyDetermination {
+				NumberOfPeople = numberOfPeople,
+				Income = income,
+				Determined = determined,
+				FamilyId = familyId
+			};
+			_context.Add(familyDetermination);
+			return familyDetermination;
+		}
+
+		public FamilyDetermination UpdateFamilyDetermination(
+			FamilyDetermination determination,
+			int? numberOfPeople,
+			decimal? income,
+			DateTime? determined
+		)
+		{
+			if(numberOfPeople.HasValue) {
+				determination.NumberOfPeople = numberOfPeople.Value;
+			}
+
+			if(income.HasValue) {
+				determination.Income = income.Value;
+			}
+
+			if(determined.HasValue) {
+				determination.Determined = determined.Value;
+			}
+
+			return determination;
+		}
 	}
 
 	public interface IFamilyDeterminationRepository
 	{
 		Task<ILookup<int, FamilyDetermination>> GetDeterminationsByFamilyIdsAsync(IEnumerable<int> familyIds, DateTime? asOf = null);
+		Task<FamilyDetermination> GetDeterminationByIdAsync(int id, DateTime? asOf = null);
+		FamilyDetermination CreateFamilyDetermination(
+			int numberOfPeople,
+			decimal income,
+			DateTime determined,
+			int familyId
+		);
+
+		FamilyDetermination UpdateFamilyDetermination(
+			FamilyDetermination determination,
+			int? numberOfPeople = null,
+			decimal? income = null,
+			DateTime? determined = null
+		);
 	}
 }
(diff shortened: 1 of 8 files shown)


src/Hedwig/Repositories/FamilyDeterminationRepository.cs
```diff
@@ -20,10 +20,69 @@ public class FamilyDeterminationRepository : TemporalRepository, IFamilyDetermin
 
 			return determinations.ToLookup(d => d.FamilyId);
 		}
+
+		public Task<FamilyDetermination> GetDeterminationByIdAsync(int id, DateTime? asOf = null)
```
cailyncodes: I've been marking these types of functions (those that return `Task<>`)  with async and adding await on the `return` line (i.e. `return await GetBaseQuery<FamilyDetermination>(asOf).FirstOrDefaultAsync(d => d.Id == id);`) But maybe I shouldn't be doing that?

src/Hedwig/Repositories/FamilyDeterminationRepository.cs
```diff
@@ -20,10 +20,69 @@ public class FamilyDeterminationRepository : TemporalRepository, IFamilyDetermin
...
+			}
+
+			if(income.HasValue) {
+				determination.Income = income.Value;
+			}
+
+			if(determined.HasValue) {
+				determination.Determined = determined.Value;
+			}
+
```
cailyncodes: 😢 My comment got deleted. Here's a replication:
I think this makes sense here. However, I wonder if this is introducing competing APIs for updating db objects. For example, in `Enrollment` we have `Entry` and `Exit` which instead of accepting `DateTime` in the mutation, accepts a string. If it is `"null"` then we clear the field; otherwise, we parse it and set it to that. If the string is `null` then we don't do anything. This wouldn't support that. Of course, it's not needed but we should come to some common pattern.

src/Hedwig/Schema/Mutations/FamilyDeterminationMutation.cs
```diff
@@ -0,0 +1,77 @@
...
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<IntGraphType>> { Name = "familyId" },
+                    new QueryArgument<NonNullGraphType<IntGraphType>> { Name = "numberOfPeople" },
+                    new QueryArgument<NonNullGraphType<DecimalGraphType>> { Name = "income" },
+                    new QueryArgument<NonNullGraphType<DateGraphType>> { Name = "determined" }
+                ),
+                resolve: async context => 
+                {
+                    var familyId = context.GetArgument<int>("familyId");
+                    var family = (Family) await families.GetFamilyByIdAsync(familyId);
```
cailyncodes: Is this explicit cast necessary?

src/Hedwig/Schema/Mutations/FamilyDeterminationMutation.cs
```diff
@@ -0,0 +1,77 @@
...
+                arguments: new QueryArguments(
+                    new QueryArgument<NonNullGraphType<IntGraphType>> { Name = "id" },
+                    new QueryArgument<IntGraphType> { Name = "numberOfPeople" },
+                    new QueryArgument<DecimalGraphType> { Name = "income" },
+                    new QueryArgument<DateGraphType> { Name = "determined" }
+                ),
+                resolve: async context => 
+                {
+                    var id = context.GetArgument<int>("id");
+                    var determination = (FamilyDetermination) await familyDeterminations.GetDeterminationByIdAsync(id);
```
cailyncodes: Likewise here?

src/Hedwig/Repositories/FamilyDeterminationRepository.cs
jlhogan: was getting at this in my PR description. Agreed that it doesn't work for objects with nullable fields. Also think we should punt on figuring out the standard approach until after weds :) will open a ticket to make sure we address

src/Hedwig/Repositories/FamilyDeterminationRepository.cs
jlhogan: yeah i dont have a hard opinion either way. That approach increases consistency, but also means writing extra code (`return await`) which could be confusing. WIll open a ticket to decide; i don't think we need to worry too much about standardizing right now.

cailyncodes: Super fair about not needing to fully standardize now. On one of my PRs I went through and made everything async/await due what I thought was a parallelization issue. But it wasn't, and I don't think it affects the actual runtime conditions.

jlhogan: i believe the difference here is just syntax styling; the actual execution will be the same either way

jlhogan:
closes #170, #176

cailyncodes:
> closes #170, #176

I could totally be wrong on this but I think I remember reading that this only works if it is in the initial description of the PR -- it can be edited in later but as a standalone comment it won't work

PR #205, opened 2019-11-04
Create checklist component (#132)

mmazanec22:


src/Hedwig/ClientApp/src/components/Checklist/Checklist.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+type ChecklistProps = {
+	options: CheckboxOptions[];
+	groupName: string;
+	onClick: (value: string) => any;
+	legend: string;
+	horizontal?: boolean;
+	showError?: boolean;
+	errorMessage?: string;
```
cmgiven: Might be an opportunity for a small refactor here. `showError` is basically redundant of the existence of `errorMessage`. I'd suggest we could key everything off of a single `error` prop.

src/Hedwig/ClientApp/src/components/Checklist/Checklist.tsx
```diff
@@ -0,0 +1,68 @@
...
+	text: string;
+  value: string;
+  checked: boolean;
+  disabled?: boolean;
+};
+
+type ChecklistProps = {
+	options: CheckboxOptions[];
+	groupName: string;
+	onClick: (value: string) => any;
```
cmgiven: Possibly we want `onClick` to be at the checkbox level? My thinking is that even though the checkboxes are grouped, generally each checkbox will correspond to a single field. So I might have:

```
const [apple, setApple] = useState(data.apple);
const [banana, setBanana] = useState(data.banana);
...
```

And then want to bind those to a checklist like this:

```
<Checklist options={[{ text: 'Apple', value: 'apple', checked: apple, onClick: setApple }, ...]} />
```

cmgiven: This is as opposed to a radio button or dropdown, which do generally correspond to a single field each.

Flip side, this might make the refactor you mentioned to combine all these components more difficult.

src/Hedwig/ClientApp/src/components/Checklist/Checklist.tsx
```diff
@@ -0,0 +1,68 @@
...
+	onClick,
+	horizontal,
+	legend,
+	showError,
+	errorMessage,
+}: ChecklistProps) {
+
+  const currentlySelectedValues: { [key: string]: any } = {};
+  options.forEach(option => currentlySelectedValues[option.value] = option.checked);
+  const [currentlySelected, setSelection] = useState(currentlySelectedValues);
```
cmgiven: Sorry, I'm closing this window now. But I think the component should probably not maintain an internal state? It could be in conflict with the `checked` props on each option.

Issue #150, opened 2019-10-24
Add CDC funding section to enrollment


abbyraskin:
**Story**: As a provider, I want to enter a child's funding information so I can enroll them in a program space.

**Acceptance Criteria**:
- Add a header below Age group field that says "Funding" (differs from comp)
- Below, add a "Funding type" dropdown, with the following options (including the standard -select- empty state):
  - CDC - full time
  - CDC - part time
  - Private pay
- If user selects CDC full time or part time as funding type, display the "First reporting period" dropdown, populated with the first reporting period inclusive of the enrollment start date entered, and the next two periods that follow (three total)
- If the user selects "Private pay" as funding type, "First reporting period" is disabled 

**Design**:
![image](https://user-images.githubusercontent.com/30734760/71018903-f7b60e00-20be-11ea-8e4a-0af7f6d2dfe8.png)

abbyraskin:
assigning to myself to get the acceptance criteria once the designs are finalized!

abbyraskin:
@atreni can you add the details for this story please? so far we've only built out the enrollment piece, but we don't have anything for funding yet (you may want to break it into multiple stories to account for the alerts, etc.), thank you!

abbyraskin:
Note: we decided to remove any indication of available spaces from the funding section of enrollment for now until we've had more time to research behavior here!

atreni:
Reporting period dates through end of 2019-2020 fiscal year:
- 10-28-2019 - 11-29-2019 (5 weeks)
- 12-2-2019 - 12-27-2019 (4 weeks)
- 12-30-2019 - 1-31-2020 (5 weeks)
- 2-3-2020 - 2-28-2020 (4 weeks)
- 3-2-2020 - 3-27-2020 (4 weeks)
- 3-30-2020 - 4-24-2020 (4 weeks)
- 4-27-2020 - 5-29-2020 (5 weeks)
- 6-1-2020 - 6-26-2020 (4 weeks)

Issue #149, opened 2019-10-24
Add enrollment step 3 components


abbyraskin:
As a provider, I want to enter information about a child's family income so I can enroll them in my program.

## Acceptance Criteria
- Adds components to the "Child's information" section in step list (third step), with the following fields under the main section label:
  - Household size (text field)
  - Annual household income (text field)
  - Date of income determination label
  - [Date field](https://design.cms.gov/components/date-field/)
- Save button

## Design
![image](https://user-images.githubusercontent.com/30734760/67503843-f5e74580-f64d-11e9-97bf-db24c4643e9c.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens/388356285)

Issue #147, opened 2019-10-24
Add enrollment step 2 components


abbyraskin:
**Story**: As a provider, I want to enter a child's family information so I can enroll them in my program.

## Acceptance Criteria
- Adds components to the "Family information" section in step list (second step)
- Address header with the following components
  - Address line 1 text field
  - Address line 2 (optional) text field
  - State dropdown, with the options: CT, MA, NY, RI
  - Town text field
- Other header with the following checkboxes
  - Child lives with foster family
  - Experiencing homelessness
- Note: Do not add "link to existing family" component shown in design (out of scope for now)

### Related stories
- (Validations)

### Design
![image](https://user-images.githubusercontent.com/30734760/67502713-f7177300-f64b-11e9-8c64-40a9ccddd16f.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens/388354420)

Issue #146, opened 2019-10-24
Add enrollment step 1 components


abbyraskin:
As a provider, I want to enter information about a child to enroll them in my program.

## Acceptance Criteria
- Adds components to the "Child's information" section in step list (first step), with the following text fields under the main section label:
  - SASID (optional) 
  - First 
  - Middle 
  - Last 
  - Suffix (optional) 
- Date of Birth header
  - Date field component
- Birth Certificate header
  - Birth certificate ID # text field
  - State dropdown
  - Town text field
- Race header, with *As identified by family* label below, and the following checkboxes:
  - American Indian or Alaska Native 
  - Asian
  - Black or African American
  - Native Hawaiian or Pacific Islander
  - White
  - Other
- Ethnicity header, with *As identified by family* label below, and the following radio buttons:
  - Not Hispanic or Latinx
  -  Hispanic or Latinx
- Gender header, with the following dropdown options:
  - -Select-
  - Female
  - Male
  - Did not specify male or female
- Save button

## Design
![image](https://user-images.githubusercontent.com/30734760/67500691-b1a57680-f648-11e9-81c3-5c3a64dbcf00.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens/388356285)

PR #222, opened 2019-11-06
[WIP] Beth Demo

cmgiven:
Connects #146, #147, #149, and #150.

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
@@ -1,26 +1,150 @@
 import React from 'react';
+import useAuthMutation from '../../../hooks/useAuthMutation';
+import {
+	CREATE_FAMILY_DETERMINATION_MUTATION,
+	UPDATE_FAMILY_DETERMINATION_MUTATION,
+	CHILD_QUERY,
+} from '../enrollmentQueries';
+import { CreateFamilyDeterminationMutation } from '../../../generated/CreateFamilyDeterminationMutation';
+import { UpdateFamilyDeterminationMutation } from '../../../generated/UpdateFamilyDeterminationMutation';
+import { ChildQuery } from '../../../generated/ChildQuery';
 import { Section } from '../enrollmentTypes';
 import Button from '../../../components/Button/Button';
+import TextInput from '../../../components/TextInput/TextInput';
+import DatePicker from '../../../components/DatePicker/DatePicker';
+import dateFormatter from '../../../utils/dateFormatter';
+import moment from 'moment';
 
 const FamilyIncome: Section = {
 	key: 'family-income',
 	name: 'Family income',
 	status: () => 'complete',
 
-	Summary: () => {
-		return <div className="FamilyIncomeSummary"></div>;
+	Summary: ({ child }) => {
+		const determination = child && child.family && child.family.determinations[0];
+		return (
+			<div className="FamilyIncomeSummary">
+				{determination ? (
+					<>
+						<p>Household size: {determination.numberOfPeople}</p>
+						<p>Annual household income: ${determination.income.toFixed(2)}</p>
+						<p>Determined on: {dateFormatter(determination.determined)}</p>
+					</>
+				) : (
+					<p>No income determination on record.</p>
+				)}
+			</div>
+		);
 	},
 
-	Form: ({ afterSave }) => {
+	Form: ({ child, afterSave }) => {
+		if (!child || !child.family) {
+			throw new Error('FamilyIncome rendered without a family');
+		}
+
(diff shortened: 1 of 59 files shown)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -1,26 +1,150 @@
...
+			<div className="FamilyIncomeForm usa-form">
+				<TextInput
+					id="numberOfPeople"
+					label="Household size"
+					defaultValue={numberOfPeople ? '' + numberOfPeople : ''}
+					onChange={event => {
+						const value = parseInt(event.target.value.replace(/[^0-9.]/g, ''), 10) || null;
+						updateNumberOfPeople(value);
+					}}
+					onBlur={event => (event.target.value = numberOfPeople ? '' + numberOfPeople : '')}
```
cailyncodes: I think there should be a double equals here? `event.target.value == numberOfPeople`

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -1,26 +1,150 @@
...
+					id="income"
+					label="Annual household income"
+					defaultValue={income ? '$' + income.toFixed(2) : ''}
+					onChange={event => {
+						const value =
+							Math.round(parseFloat(event.target.value.replace(/[^0-9.]/g, '')) * 100) / 100 ||
+							null;
+						updateIncome(value);
+					}}
+					onBlur={event => (event.target.value = income ? '$' + income.toFixed(2) : '')}
```
cailyncodes: Again here

src/Hedwig/ClientApp/src/containers/Enrollment/enrollmentQueries.tsx
```diff
@@ -1,17 +1,282 @@
...
+	${CHILD_FRAGMENT}
+`;
+
+export const UPDATE_CHILD_MUTATION = gql`
+	mutation UpdateChildMutation(
+		$id: String!
+		$sasid: String
+		$firstName: String!
+		$middleName: String
+		$lastName: String!
```
cailyncodes: I don't think these need to be required on update.

src/Hedwig/ClientApp/src/hooks/useAuthQuery.ts
```diff
@@ -18,6 +18,5 @@ export default <T>(query: DocumentNode, options: QueryHookOptions = {}): QueryRe
 				Authorization: accessToken ? `Bearer ${accessToken}` : '',
 			},
 		},
-		fetchPolicy: 'cache-and-network',
```
cailyncodes: I ran into an issue displaying data without this set. Once I checkout this branch locally I'll update on whether it still works for me.

src/Hedwig/Schema/Types/EnrollmentType.cs
```diff
@@ -7,49 +7,60 @@
...
-						loaderCacheKey,
-						(ids) => children.GetChildrenByIdsAsync(ids, asOf));
+    public class EnrollmentType : TemporalGraphType<Enrollment>, IAuthorizedGraphType
+    {
+        public EnrollmentType(IDataLoaderContextAccessor dataLoader, IChildRepository children, IFundingRepository fundings, ISiteRepository sites)
+        {
+            Field(e => e.Id);
+            Field<DateTime?>(e => e.Entry, type: typeof(DateGraphType), nullable: true);
+            Field<DateTime?>(e => e.Exit, type: typeof(DateGraphType), nullable: true);
+            Field(e => e.Age, type: typeof(AgeEnumType));
```
cailyncodes: Maybe `AgeEnumType` can't be null in which case this change is good. If not, though, I thought we were temporarily proceeding with no nulls for age?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
cmgiven: No, I'm actually doing assignment here. This is so that when I leave the field, it'll replace what's in the text box with the cleaned up number it'll try to save. So if I enter "2.5", it'll round that down to 2.

src/Hedwig/ClientApp/src/containers/Enrollment/enrollmentQueries.tsx
cmgiven: Good point.

src/Hedwig/Schema/Types/EnrollmentType.cs
cmgiven: Decision: It should be non-nullable, but needs a pseudo-null value in the enum.

PR #225, opened 2019-11-07
209: reconcile frontend with backend changes

jlhogan:
Remove fields from queries/generated types that have been removed from backend:
- funding: entry, exit,
- family: casenumber


jlhogan:
outdated; only whitespace diff but git is not recognizing that as no diff

PR #227, opened 2019-11-08
Rest api with master

cailyncodes:
This branch was created by starting a new project/solution in VS with the API template. I iteratively added Hedwig files into this new project (upgrading from dotnet core 2.2 to 3.0 in the process) until everything sans GraphQL stuff was added. I then created a new branch where I deleted all of Hedwig. I then through cherry-picking and interactive rebasing added the Hedwig git history to this new project and removed the delete everything commit. I added the test folder as a second commit and the tests (less GraphQL/integration) pass (at least one my computer).

~~I haven't taken a stab at updating the azure-pipelines nor buildspec yaml files, so it probably won't build on CI/CD.~~ Thanks to tremendous work by @speedcm, we have a building CI/CD!


docker-compose.yaml
```diff
@@ -36,19 +35,3 @@ services:
     entrypoint:
       - sh
       - /entrypoint.sh
-  winged-keys:
```
jlhogan: why is winged-keys out?

src/Hedwig/Program.cs
```diff
@@ -1,51 +1,21 @@
...
-					logger.LogError(ex, "An error occurred while seeding the database.");
-				}
-			}
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.UseStartup<Startup>();
+                });
+    }
```
cailyncodes: Need to add back in logging

Issue #230, opened 2019-11-12
API v2 Spec


jlhogan:
`/users` (GET/POST)
|---`/:id` (GET/PUT)
|---`/current` ?
`/organizations` 
|---`/:id` 
|---|---`/sites`
|---|---|---`/:id` 
|---|---|---|---`/enrollments` (GET/POST)
|---|---|---|---|---`/:id` (GET/PUT)
|---|---`/children` (GET)
|---|---|----`/:id` (GET/PUT)
|---|---`/families` (GET)
|---|---|---`/:id` (GET/PUT)
|---|---`/reports` (GET/POST)
|---|---|---`/:id` (GET/PUT)

For Pilot MVP:
- users will only create families & children via `.../enrollments`
- children and families will be updated via PUT to `.../enrollments/id`

cmgiven:
We don't have to worry about org-level access to children until we get to supporting multiple sites, but I suspect that it would be useful at that point.

Thoughts on advantages of nesting everything under organization versus a flatter schema? As a strawman, something like:

`GET /users/:id` (Name and lists of the organizations and sites I have access to.)
`GET /sites/:id/roster`, `GET /organizations/:id/roster` (Roster of a site or organization; potentially would need additional parameters to filter by date.)
`POST /children` (Create a new child.)
`GET /children/:id` (Everything about a child, including enrollments, fundings, etc.)
`PUT /children/:id` (Update a child.)
`POST /families`
`GET /families/:id`
`PUT /families/:id`
`GET /reports` (List of all reports I have access to.)
`GET /reports/:id`
`GET /reports/:id/roster` (Roster as of the time the report was submitted.)
`POST /reports/:id/submit`
`POST /reports/:id/revise`

And then, only models that can only ever be accessed through a single parent are nested.

`POST /families/:id/determinations` (Create a new determination.)
`GET /families/:id/determinations/:id`
`PUT /families/:id/determinations/:id`
`POST /children/:id/enrollments`
`GET /children/:id/enrollments/:id`
`PUT /children/:id/enrollments/:id`
`POST /children/:id/enrollments/:id/fundings/`
`GET /children/:id/enrollments/:id/fundings/:id`
`PUT /children/:id/enrollments/:id/fundings/:id`

I know that a child can only ever have one organization in our model. But as a child could be enrolled in multiple sites, or a family could have multiple children, maybe that's a reason to avoid nesting those models?

cailyncodes:
[...]

PR #247, opened 2019-11-12
233: Adds a mocked rest backend to the docker-compose project

### LINKED ISSUE #233: mock backend
**user story**: as a frontend developer, I want to continue to develop frontend features while the backend is refactored

**AC**:
- mocked backend that can serve static data at correct paths
- mimics api spec #230

jlhogan:
adds a simple json server to docker-compose project that serves static data at localhost:6000

missing: family not accessible via child route. we did not determine in API v2 Spec (#230) if we want to expose family on the child via an optional `include=family` or `includeFamily=true` qs param

closes #233

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -24,6 +24,18 @@ services:
     entrypoint:
       - sh
       - /entrypoint.sh
+  rest-backend:
+    image: clue/json-server
+    volumes:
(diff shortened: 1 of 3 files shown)


cailyncodes:
In light of #260 and it's demonstration of the power of additional field inclusion, I would support the `include[]=[objects]` over `includeFamily` boolean.

PR #260, opened 2019-11-13
Add serializing support without DTO

### LINKED ISSUE #258: Serializing EF object data
How should we serialize recursive data structures to JSON?

cailyncodes:
Should close #258 

Added Newtonsoft b/c the Dotnet Core 3 built-in is not feature-ful enough for our use case, re recursive serialization.

As part of this, I explored additional sub-field inclusion. ASPNet Core provides us great query string inspection that makes adding this super easy! Combined with EF's hydration support and we can do some fancy things!

For example, try these queries:
* `/api/organizations/1`
* `/api/organizations/1?include[]=site`
* `/api/organizations/1?include[]=site&include[]=enrollment`

### DIFF
--- a/src/Hedwig/Controllers/OrganizationController.cs
+++ b/src/Hedwig/Controllers/OrganizationController.cs
@@ -13,10 +13,17 @@ namespace Hedwig.Controllers
     public class OrganizationsController : ControllerBase
     {
         private readonly IOrganizationRepository _organizations;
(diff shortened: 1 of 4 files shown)


jlhogan:
@cmgiven  given the easy of implementing the `include` logic, would we feel OK with going this route? I think it's cleaner than default including embedded objects.

PR #261, opened 2019-11-14
232 authorization

### LINKED ISSUE #232: Authorization rules in dotnet core
Implement authorization rules leveraging dotnet core classes and paradigms. I believe authentication issues claims, so we should implement claims-based authorization https://docs.microsoft.com/en-us/aspnet/core/security/authorization/claims?view=aspnetcore-3.0

**AC**: 
- a policy exists to enforce authenticated user (default `Authorize`)
- a policy exists to check that current user has access to root organization/site
- a policy exists to check that current user has access to given user
    - option 1: user with site permissions can see users at site, user with organization permissions can see users at any site in org
    - option 2: user can only access self

jlhogan:
- Implements simple authorization requirements to enforce user access to an organization or a site; see: `Security_NEW/UserSiteAccessRequirement.cs` and `Security_New/UserOrganizationAccessRequirement.cs`
- Adds each custom requirement to an authorization policy; see: `ServiceExtensions.cs` and `Startup.cs`
- Adds a multi-policy AuthorizationHandler, called `PermissionHandler` for now copied from a microsoft example, but maybe confusing since nothing else is called "permissions"; see: `Secutiry_NEW/PermissionHandler.cs`

note: this does not explicitly add an authenticated user requirement. the requirement is implicit in the other requirements. this may not be the best design.

TODO: actually implement the PermissionRepository code for user/site and user/organization logic

closes #232

### DIFF
--- a/src/Hedwig/Security_NEW/Requirements/UserOrganizationAccessRequirement.cs
+++ b/src/Hedwig/Security_NEW/Requirements/UserOrganizationAccessRequirement.cs
@@ -1,0 +1,46 @@
+using System;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Hedwig.Repositories;
+
+namespace Hedwig.Security_NEW
+{
+    public class UserOrganizationAccessRequirement : IAuthorizationRequirement, IPermissionRequirement
+    {  
+        public const string NAME = "UserOrganizationAccess";
+
+        public bool Evaluate(
+            ClaimsPrincipal user,
+            HttpContext httpContext,
+            IPermissionRepository permissions
+        )
+        {
+            return  UserCanAccessOrganization(user, httpContext, permissions);
+        }
+
+        private bool UserCanAccessOrganization(
+            ClaimsPrincipal user,
+            HttpContext httpContext,
+            IPermissionRepository permissions
+        )
+        {
+            var userIdStr = user.FindFirst("sub")?.Value;
+            var routeData = httpContext.GetRouteData();
+            // If request controller = 'Organizations', path param for organiation id is 'id'
+            // else, path param is 'orgId'
+            var orgIdParam = (string)routeData.Values["controller"] == "Organizations" ? "id" : "orgId";
+            var orgIdStr = (string)routeData.Values[orgIdParam];
+            if(userIdStr != null && orgIdStr != null) {
+                var userId = Int32.Parse(userIdStr);
+                var orgId = Int32.Parse(orgIdStr);
+                return permissions.UserCanAccessOrganization(userId, orgId);
+            }
+
+            return false;
+
+        }
+    }
+}
--- a/src/Hedwig/Security_NEW/Requirements/UserSiteAccessRequirement.cs
+++ b/src/Hedwig/Security_NEW/Requirements/UserSiteAccessRequirement.cs
@@ -1,0 +1,46 @@
+using Microsoft.AspNetCore.Authorization;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Http;
+using System;
+using System.Security.Claims;
+using Hedwig.Repositories;
+
+namespace Hedwig.Security_NEW
+{
+    public class UserSiteAccessRequirement : IPermissionRequirement
+    {  
+        public const string NAME = "UserSiteAccess";
+
+        public bool Evaluate(
+            ClaimsPrincipal user,
+            HttpContext httpContext,
+            IPermissionRepository permissions
+        )
+        {
+            return UserCanAccessSite(user, httpContext, permissions);
+        }
+        
+        private bool UserCanAccessSite(
+            ClaimsPrincipal user,
+            HttpContext httpContext,
+            IPermissionRepository permissions 
+        )
+        {
+            var userIdStr = user.FindFirst("sub")?.Value;
+            var routeData = httpContext.GetRouteData();
+            // If request controller is 'Sites', path param for site id is 'id'
+            // else, path param is 'siteId'
+            var siteIdParam = (string)routeData.Values["controller"] == "Sites" ? "id" : "siteId";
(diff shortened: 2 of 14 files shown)


src/Hedwig/Security_NEW/PermissionHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+                        context.Succeed(req);
+                    }
+                }
+            }
+
+            return;
+        }
+
+
+        private Task<bool> UserCanAccessSite(
```
cailyncodes: Thoughts on putting these methods as a instance method on the requirement classes themselves? It feels like it's inconsistent with the "tell, don't ask" pattern (https://thoughtbot.com/blog/tell-dont-ask -- thanks to Gabriel for sharing this).

This would let us do something like
```
foreach (var req in pendingRequirements) {
  if (await req.Evaluate(context.User, _htttp[...]HttpContext)) context.Succeed(req)
}
```
and gives us better testing support.

cailyncodes: It would require us to DI the permissions repo into the requirement and probably use a service provider in the configure services startup. So I'm not 100% sure where I fall on this.

jlhogan: I split the difference and implemented the checks on the authoriztion requirements, accepting IPermissionRepository as a parameter

jlhogan:
thoughts on encapsulating access check in repository? Alternative is repo exposing something like: `GetPermittedSiteIdsByUserId` and `GetPermittedOrganizationIdsByUserId` and then doing the check in the authorizationrequirements

cailyncodes:
> 
> 
> thoughts on encapsulating access check in repository? Alternative is repo exposing something like: `GetPermittedSiteIdsByUserId` and `GetPermittedOrganizationIdsByUserId` and then doing the check in the authorizationrequirements
> (@cmgiven also maybe has thoughts as OG repo implementor?)

I guess I don't understand what the "access check in repository" would look like? The alternative you mention seems reasonable to me, and moreover I think provides transparency in the authorization requirement class what is happening.

src/Hedwig/Security_NEW/Requirements/UserOrganizationAccessRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        private bool UserCanAccessOrganization(
+            ClaimsPrincipal user,
+            HttpContext httpContext,
+            IPermissionRepository permissions
+        )
+        {
+            var userIdStr = user.FindFirst("sub")?.Value;
+            var routeData = httpContext.GetRouteData();
+            var orgIdParam = (string)routeData.Values["controller"] == "Organizations" ? "id" : "orgId";
```
cailyncodes: This line is a little funky. Can you throw a comment on this?

src/Hedwig/Security_NEW/Requirements/UserSiteAccessRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+        
+        private bool UserCanAccessSite(
+            ClaimsPrincipal user,
+            HttpContext httpContext,
+            IPermissionRepository permissions 
+        )
+        {
+            var userIdStr = user.FindFirst("sub")?.Value;
+            var routeData = httpContext.GetRouteData();
+            var siteIdParam = (string)routeData.Values["controller"] == "Sites" ? "id" : "siteId";
```
cailyncodes: Same here

cmgiven:
Yeah, I'm also unclear on "encapsulating access check in repository."

jlhogan:
> encapsulating access check in repository."

referring to the fact that the access check logic lives in the repo function. Not sure how we feel about that. If repository is just supposed to be data access, then perhaps it should simply return the list of siteIds the user has permission to, and the requirement checks that the user has access.

Also added some inline comments for discussion.

src/Hedwig/Controllers/ChildrenController.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System;
+using Hedwig.Models;
+using Hedwig.Repositories;
+using Hedwig.Security_NEW;
+
+namespace Hedwig.Controllers
+{
+    [Route("api/organizations/{orgId:int}/children")]
```
jlhogan: putting `child` under site felt weird when I came to implementing the auth checks. I know that Chris mentioned that users in a single site should not be able to see children from outside their site. I wonder if it would be acceptable to enforce this via UI flows, and only actually restrict on the backend at an organization level ?  I would like to know more about the permissioning model we want to implement, but needing to complicate it at this early stage feels meh.

src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -0,0 +1,60 @@
...
+using Hedwig.Repositories;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+using System.Linq;
+using Hedwig.Models;
+using Hedwig.Security_NEW;
+
+namespace Hedwig.Controllers
+{
+    [Route("api/organizations/{orgId:int}/sites/{siteId:int}/enrollments")]
```
jlhogan: `orgId` is useless here, and could be literally any value, and I can't figure out how it could be useful, except for API consistency. Raises the question does `sites/#/enrollments/` needs to be nested under organization since site ids are unique across the DB, not org specific?

src/Hedwig/Controllers/ChildrenController.cs
cmgiven: I'd support that/think we can justify that.

src/Hedwig/Controllers/EnrollmentsController.cs
cmgiven: If we reach a point where we can get rid of path parameters that don't serve a purpose, that seems a lot better. I'm imagining writing API documentation, and trying to describe the reason for a parameter that _may_ be used, but probably not.

cailyncodes: I'm comfortable raising sites to be top-level sibling of organizations. I don't think that that presents an issue with auth checking? A user may only have an explicit org permission but we should be able to go siteId -> orgId -> auth check?

PR #286, opened 2019-11-15
265: Adds organizationId to family

### LINKED ISSUE #265: Add organizationId to family
add organizationId to Family and Child

jlhogan:
Also, sets OnDelete = restrict for _ALL_ fks with a custom
ModelBuilderExtension method

closes #265

### DIFF
--- a/src/Hedwig/Data/DbInitializer.cs
+++ b/src/Hedwig/Data/DbInitializer.cs
@@ -61,7 +61,7 @@ namespace Hedwig.Data
 			context.SaveChanges();
 
 			var families = Enumerable.Range(1, 15)
-				.Select(i => new Family())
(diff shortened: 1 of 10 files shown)


jlhogan:
i can update the extension to set fks to restrict on a per-table basis, and leave things like fundings and determinations with cascade behavior

Making this change does not change the object relationships in their essence. They're not nullable, the deletes just don't cascade. As long as we implement clean up properly in the API (and ensure we are good stewards when doing manual data cleanup), null FKs will not happen. I think :).

cmgiven:
No strong feelings here. Just wondering what was up.

jlhogan:
opted to NOT deal with it, altho i don't like that a totally required field (orgId on family, or orgId on site) is nullable! :(

PR #289, opened 2019-11-19
252 children

### LINKED ISSUE #252: Children controller
path: `/organizations/{orgId}/sites/{siteId}/children`, `/organizations/{orgId}/sites/{siteId}/children/{id}`
methods: GET, POST, PUT
available `include[]`: family

**User Story**: As an API consumer, I want to access child entities via `/organizations/{orgId}/sites/{siteId}/children`

**AC**:
- GET is implemented for `.../children/` and `.../children/{id}`

jlhogan:
closes #252 

includes commit from PR #286 which should be merged first

### DIFF
--- a/src/Hedwig/Controllers/ChildrenController.cs
+++ b/src/Hedwig/Controllers/ChildrenController.cs
@@ -1,0 +1,89 @@
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.EntityFrameworkCore;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Linq;
+using System;
+using Hedwig.Models;
+using Hedwig.Repositories;
+using Hedwig.Security_NEW;
+
+namespace Hedwig.Controllers
+{
+    [Route("api/organizations/{orgId:int}/[controller]")]
+    [ApiController]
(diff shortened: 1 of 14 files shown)


src/Hedwig/Controllers/ChildrenController.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Threading.Tasks;
+using System.Linq;
+using System;
+using Hedwig.Models;
+using Hedwig.Repositories;
+using Hedwig.Security_NEW;
+
+namespace Hedwig.Controllers
+{
+    [Route("api/organizations/{orgId:int}/[controller]")]
```
jlhogan: site-only users can have permissions to:
- READ anything udner `/sites/{siteId}/....`
- WRITE anything under `/sites/{siteId}/...` (can create any object tree, so can create children via `/enrollments`)
- READ/UPDATE a _specific_ entity by Id under an organization your permitted sites belong to (so can see/update a child or family that you created via enrollment flow, but limits access because you have to know the id -- you can't just pull all the data for the whole org and find out who is in there)

which means  child & family entities do not need to be accessible via `organization/sites/children` or `sites/children`; they can be  retrieved and written via `sites/enrollments`

we can expose child & family entities only at the organization level

PR #290, opened 2019-11-19
Use config file for React App OpenID url

cailyncodes:
It looks like your suggestion will work! (We just need to confirm that there aren't any optimizations during the build phase that will prevent us from reading the value at runtime)

### DIFF
--- a/src/Hedwig/ClientApp/public/config.json
+++ b/src/Hedwig/ClientApp/public/config.json
@@ -1,0 +1,3 @@
+{
+	"WingedKeysUri": "https://localhost:5050"
+}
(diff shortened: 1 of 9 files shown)


speedcm:
If we use the config.json file approach, do I need to create it in the root of the installation?  Can you please test this locally first to make sure it works?

cailyncodes:
Not sure what the `manifest.json` file is? I don't see it in the `ClientApp/` directory.

cailyncodes:
> 
> 
> If we use the config.json file approach, do I need to create it in the ClientApp\build directory? Can you please test this locally first to make sure it works? We have a build once deploy many model. As a result we do not build for a specific runtime environment.

I'm checking right now if `config.json` is persisted after the build stage.

PR #291, opened 2019-11-21
234 openapi client

mmazanec22:
This incorporates @jlhogan's changes to add a mocked backend.


src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -14,62 +11,53 @@ import Button from '../../components/Button/Button';
...
+	child: Child;
+	fundings: Funding[];
+}
 
 export default function Roster() {
 	const [showPastEnrollments, toggleShowPastEnrollments] = useState(false);
 	const [dateRange, setDateRange] = useState<DateRange>(getDefaultDateRange());
 	const [byRange, setByRange] = useState(false);
+	const { data, runQuery } = useOASClient('organizationsOrganizationIdSitesSiteIdGet', {
+		// TODO: Get org and site id from authenticated user?
```
jlhogan: That seems right to me: A call to `user/current/` returns a user object that includes org id & site id(s). 

 As we extend the authorization model to allow diff functionality for diff users, we will probably want to include not just the ids but also the types of permissions the user has. But thats for a later date.

src/Hedwig/ClientApp/src/hooks/useOASClient.ts
```diff
@@ -0,0 +1,35 @@
...
+	const [user, setUser] = useState();
+	const [data, setData] = useState();
+
+	const { accessToken, withFreshToken } = useContext(LoginContext);
+	useEffect(() => {
+		withFreshToken();
+	}, []);
+
+  const api = accessToken ? new DefaultApi(new Configuration({ accessToken })) : null;
+  // TODO: NEED USER IN MOCKED DATA?
```
jlhogan: before every API call, we make a call to `user/current`? What is the flow here?

hedwig-1.0.0-OAS.yaml
```diff
@@ -0,0 +1,261 @@
...
+          type: boolean
+        gender:
+          $ref: '#/components/schemas/Gender'
+        foster:
+          type: boolean
+        family:
+          $ref: '#/components/schemas/Family'
+    Gender:
+      type: string
+      enum: [Male, Female, Unknown, Unspecified]
```
cailyncodes: Flagging as something that will need to be updated given our newest recommendation.

rest-backend/routes.json
```diff
@@ -1,7 +1,9 @@
 {
   "/organizations/1/sites/1": "/sites/1",
+  "/organizations/1/sites/1\\?include*": "/siteHydrated",
   "/organizations/1/sites/1/enrollments": "/enrollments",
   "/organizations/1/sites/1/enrollments/:id": "/enrollments/:id",
+  "/organizations/1/sites/1/enrollments/:id\\?include*": "/enrollments/:id",
```
cailyncodes: Should this by `enrollmentsHydrated`?

src/Hedwig/ClientApp/src/containers/App/App.tsx
```diff
@@ -25,6 +26,7 @@ export const APP_QUERY = gql`
 `;
 
 const App: React.FC<WithLoginPropsType> = ({ accessToken }) => {
+	useOASClient();
```
cailyncodes: Should we be assigning the result of this function call?

src/Hedwig/ClientApp/src/hooks/useOASClient.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (!query || !api) {
+      return;
+    }
+    const rval = await (api as any)[query](params);
+    setData(rval)
+  };
+
+	useEffect(() => {
+    runQuery()
+	}, [user]);
```
cailyncodes: I don't know if this is needed? We never call `setUser` so the local state should never change. As a result, this `useEffect` call will only ever be called once, so we can just call the `runQuery` function.

PR #293, opened 2019-11-21
253 enrollments

jlhogan:


src/Hedwig/Controllers/HedwigController.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class HedwigController : ControllerBase
+    {
+        public const string CHILD = "child";
+        public const string FUNDINGS = "fundings";
+        public const string FAMILY = "family";
+        public const string DETERMINATIONS = "determinations";
+        protected bool INCLUDE_FAMILY { get; private set; } = false;
+        protected bool INCLUDE_DETERMINATIONS { get; private set; } = false;
+        public bool INCLUDE_CHILD  { get; private set; }= false;
+        public bool INCLUDE_FUNDINGS { get; private set; }= false;
```
cmgiven: Why do these need to be public?

src/Hedwig/Controllers/ChildrenController.cs
```diff
@@ -0,0 +1,98 @@
...
+            _families = families;
+        }
+
+        [HttpGet]
+        public async Task<ActionResult<List<Child>>> Get(
+            int orgId,
+            [FromQuery(Name="include[]")] string[] include
+        ) 
+        {
+            ProcessInclude(include);
```
cmgiven: Nit: don't love the pattern of needing to run this at the top of each action. Concern being that the code would still run fine if this line was forgotten/deleted, because the INCLUDE_* all default to false. If going the inheritance route, possibly something like overriding OnActionExecuting in the parent class would at least make this automatic. Alternatively, could use a utility class to explicitly parse the include array, avoiding side effects.

Another nit: I think the idiomatic way to identify private variables in C# is _varName? I would assume an ALL_CAPS name indicated a constant.

src/Hedwig/Controllers/ChildrenController.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            var children = await _children.GetChildrenForOrganizationAsync(orgId);
+
+            if(INCLUDE_FAMILY) {
+                await _families.GetFamiliesByIdsAsync(
+                    children.Where(c => c.FamilyId != null).Select(c => c.FamilyId.Value).ToArray(),
```
cmgiven: This feels like a kind of hacky way to achieve what I think should be a `.ThenInclude` in the repository. Particularly when we are occasionally using parameters like INCLUDE_DETERMINATIONS below to do it the proper way.

Could we instead pass an array of includes to the repository, and do something like the following?

```C#
var enrollments = _context.Enrollments
    .Where(e => e.SiteId == siteId);

if (includes.Contains(FUNDING))
{
    enrollments = enrollments.Include(e => e.Fundings);
}

if (includes.Contains(CHILDREN))
{
    enrollments = enrollments.Include(e => e.Child);

    if (includes.Contains(FAMILY))
    {
        enrollments = enrollments.ThenInclude(c => c.Family)

        if (includes.Contains(DETERMINATIONS))
        {
            enrollments = enrollments.ThenInclude(f => f.Determinations)
        }
    }
}

return enrollments.ToListAsync();
```

src/Hedwig/Repositories/PermissionRepository.cs
```diff
@@ -0,0 +1,54 @@
...
+        public bool UserCanAccessSite(int userId, int siteId)
+        {
+            // User can access site if:
+            // - Site permission exists for user for site
+            // - Organization permission exists for user for organization that includes site 
+
+            var sitePermissions = _context.Permissions
+                .OfType<SitePermission>()
+                .Where(sp => sp.UserId == userId && sp.SiteId == siteId)
+                .FirstOrDefault();
```
cmgiven: We have an opportunity to bail out early without checking organization permissions if the site permission exists.

src/Hedwig/Repositories/PermissionRepository.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            var organizationPermissions = _context.Permissions
+                .OfType<OrganizationPermission>()
+                    .Include(op => op.Organization)
+                        .ThenInclude(o => o.Sites)
+                .Where(op => op.UserId == userId)
+                .SelectMany(op => op.Organization.Sites.Select(s => s.Id))
```
cmgiven: I think this would work?

```C#
.Where(op => op.UserId == userId & op.Organization.Sites.Any(s => s.Id == siteId))
```

src/Hedwig/Controllers/ChildrenController.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            return child;
+        }
+
+        [HttpPost]
+        public async Task<ActionResult<Child>> Post(int orgId, Child child)
+        {
+            if(child.Id != new Guid()) return BadRequest();
+            
+            // TODO validate that orgId == child.organizationId (as part of model validation?)
```
cmgiven: Don't we kind of have to do this validation here? Because `orgId` won't make it to the model.

src/Hedwig/Security_NEW/Requirements/UserOrganizationAccessRequirement.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.AspNetCore.Http;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Hedwig.Repositories;
+
+namespace Hedwig.Security_NEW
+{
+    public class UserOrganizationAccessRequirement : IAuthorizationRequirement, IPermissionRequirement
+    {  
+        public const string NAME = "UserOrganizationAccess";
```
cailyncodes: Not something we need to do now, but something to consider. By using a constant only within the context of each class, we could theoretically have a naming collision via the strings that would probably be a pain to figure out. Perhaps we consider enum-ing these instead so we have compile time enforcement of uniqueness?

src/Hedwig/Security_NEW/RequirementsHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+            var pendingRequirements = context.PendingRequirements.ToList();
+
+            foreach (var req in pendingRequirements) {
+                if (req is IPermissionRequirement pReq)
+                    if(pReq.Evaluate(
+                        context.User,
+                        _httpContextAccessor.HttpContext,
+                        _permissions
+                        )
+                    ) context.Succeed(req);
```
cailyncodes: Nit: Can we throw some brackets around the body of this "if"? I know it doesn't matter but I think it is cleaner and more consistent to always wrap "if" bodies.

test/HedwigTests/Fixtures/TestHedwigContext.cs
```diff
@@ -21,7 +21,7 @@ public TestHedwigContext(DbContextOptions<HedwigContext> options, IHttpContextAc
 
         public override int SaveChanges()   
         {
-            if(!ShouldRetainObjects || !TestEnvironmentFlags.ShouldRetainObjects()) {
+            if(!(ShouldRetainObjects || TestEnvironmentFlags.ShouldRetainObjects())) {
```
cailyncodes: This changes the semantics of this conditional -- is that intentional?

src/Hedwig/Security_NEW/RequirementsHandler.cs
jlhogan: will address this in a cleanup/comment PR after everything's merged. On the list !

test/HedwigTests/Fixtures/TestHedwigContext.cs
jlhogan: yes but then i got rid of test data clean up -- was causing deadlocks in the DB on cleanup, and thus nondeterministic test failures. Forgot to clean this up as well. Ty !

src/Hedwig/Repositories/EnrollmentRepository.cs
```diff
@@ -12,6 +12,45 @@ public class EnrollmentRepository : TemporalRepository, IEnrollmentRepository
...
+		}
+		public void AddEnrollment(Enrollment enrollment) 
+		{
+			_context.Add(enrollment);
+		}
+
+		public Task SaveChangesAsync()
+		{
+			return _context.SaveChangesAsync();
+		}
```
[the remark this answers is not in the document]
jlhogan: I don't believe there's a difference. Just syntactic. Have you proven otherwise?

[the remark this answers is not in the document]
cailyncodes: Not in the C# world, no, haha. In JS, I know the answer but C# is a mystery realm.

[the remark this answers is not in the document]
jlhogan: Yeah, the difference i have found in the ability to cast types to other types. async/awaiting the task means in that context it's type if accessible. In the controllers, we have to await the repo tasks so the type of the resolved task is available to be cast or transformed to the `ActionResult<T>` type that the controllers methods return

PR #300, opened 2019-11-25
296 user context

mmazanec22:
Thank you @cailyncodes for pairing on this!  :tada:

### DIFF
--- a/src/Hedwig/ClientApp/src/hooks/useOASClient.ts
+++ b/src/Hedwig/ClientApp/src/hooks/useOASClient.ts
@@ -2,29 +2,32 @@ import { useContext, useEffect, useState } from 'react';
 import { DefaultApi, Configuration } from '../OAS-generated';
 import LoginContext from '../contexts/Login/LoginContext';
 
-export default function useOASClient (query?: string, params?: any) {
+export default function useOASClient(query?: string, params?: any) {
 	const [data, setData] = useState();
 	const { accessToken, withFreshToken } = useContext(LoginContext);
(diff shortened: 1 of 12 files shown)


src/Hedwig/ClientApp/src/hooks/useOASClient.ts
```diff
@@ -2,29 +2,31 @@ import { useContext, useEffect, useState } from 'react';
...
+			// TODO: HANDLE ERRORS
+			console.log("These are not the error messages you're looking for: ", e);
+		}
+	};
 
 	useEffect(() => {
-    runQuery()
-	});
+		runQuery();
+	}, [query, accessToken, JSON.stringify(params)]);
```
cmgiven: Should a different accessToken prompt the query to re-run? Does this handle logouts or something?

cmgiven: This is a good first step, and this comment shouldn't affect merging. But how do we update the data if something other than the parameters changes? Like if I fire off a POST, how do I invalidate this state, or better, update it on a successful response?

Something we'll figure out this week, I suppose!

cailyncodes: My initial thoughts are to use a component-state for the response and funnel that to a application context. The application context will export an `updateData` (or other name) function that can be run after a successful POST/PUT operation. It does but the burden on the calling-client, but doesn't feel too encumbering. Totally open to other ideas as I've only partially formed this thought.

PR #306, opened 2019-11-26
no tags wahoo

jlhogan:


jlhogan:
changes captured in `mega-refactor`

PR #309, opened 2019-11-26
Prepare for WingedKeys users table

cailyncodes:
Add `WingedKeysId` to `User` model because WingedKeys expects a Guid as the `sub` claim.

### DIFF
--- a/src/Hedwig/Data/DbInitializer.cs
+++ b/src/Hedwig/Data/DbInitializer.cs
@@ -40,7 +40,11 @@ namespace Hedwig.Data
 
 			var users = new User[]
 			{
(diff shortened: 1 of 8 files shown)


cailyncodes:
> 
> 
> LGTM as well.
> 
> Should we use the Guid from winged keys as the only Id on user? or might that complicate user creation process ?

The id field in Hedwig is an int while a string in WingedKeys. I didn't want to have to bother changing the type in Hedwig to string (but we totally can). Identity Server in several places expects a string, so changing it in WingedKeys to int was sort of out of the option. I felt like this was the cleanest approach without needing to change too much code.

PR #313, opened 2019-12-02
[WIP] Mega refactor

cmgiven:


src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -0,0 +1,95 @@
+import { useContext, useEffect, useState, DependencyList } from 'react';
+import { Configuration, HedwigApi } from '../OAS-generated';
+import LoginContext from '../contexts/Login/LoginContext';
+
+export type Reducer<TData> = (data: TData, result: TData) => TData;
```
cmgiven: I'd suggest we allow for the possibility that the type of `result` might be different than the type of `data`.

src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -0,0 +1,95 @@
...
+	(TData | undefined),
+	Mutate<TData>
+];
+
+export default function useApi<TData>(
+  query: (api: HedwigApi) => Promise<TData>,
+  deps: DependencyList = [],
+	defaultValue?: TData,
+	skip: boolean = false,
+	callback?: (_: TData) => void
```
cmgiven: Perhaps we wrap these in an `opts` object? So I can do something like specify `skip` without providing a `defaultValue`.

src/Hedwig/ClientApp/src/routes.ts
```diff
@@ -40,16 +40,16 @@ export const routes: RouteConfig[] = [
 		component: EnrollmentNew,
 	},
 	{
-		path: '/roster/enrollments/:childId',
+		path: '/roster/enrollments/:enrollmentId',
```
cmgiven: Note that we'll need to be able to show multiple enrollments on this page. That doesn't rule this approach out, although there is some awkwardness around having multiple URLs (one for each enrollment) that show the same content.

src/Hedwig/ClientApp/src/utils/dateFormatter.ts
```diff
@@ -1,8 +1,10 @@
-export default function dateFormatter(date?: string | null) {
+import moment, { Moment } from 'moment';
+
+export default function dateFormatter(date?: string | null | Date | Moment) {
+  // TODO: USE MOMENT MAYBE?
```
cmgiven: We can delete this comment.

src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.tsx
```diff
@@ -1,49 +1,59 @@
...
+	const { user } = useContext(UserContext);
+	const params: ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGetRequest = {
+		id: enrollmentId ? enrollmentId : 0,
+		orgId: (user && user.orgPermissions && user.orgPermissions[0] && user.orgPermissions[0].organizationId) || 1,
+		siteId: siteId ? parseInt(siteId) : 1,
+		include: ['child', 'family', 'determinations', 'fundings'],
+	}
+	const [loading, error, enrollment, mutate] = useApi<Enrollment>(
+		(api) => api.apiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet(params),
+		[enrollmentId],
```
cmgiven: Looks like we also need `user` and `siteId` here.

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -36,47 +38,58 @@ const mapSectionsToSteps = (sections: Section[]) => {
...
+			history.push(`/roster/enrollments/${enrollment.id}/new/${nextSectionId}`);
 		}
 	};
 
+	const [loading, error, enrollment, mutate] = useApi<Enrollment>(
+		(api) => api.apiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet(params),
+		[enrollmentId],
+		undefined,
+		!enrollmentId,
+		afterSave
```
cmgiven: I'm uncertain about this pattern, where we're using a callback that is be called after anything is retrieved from the server, either the initial fetch or a mutation. Could we instead attach this callback to the promise returned by `mutate()`?

src/Hedwig/ClientApp/src/hooks/useApi.ts
cmgiven: Note that `deps` is not included in this comment. It needs to be outside of the object so that eslint can pick it up for static analysis of whether it's exhaustive.

Issue #272, opened 2019-11-15
Add "Back to reports" link to CDC report


abbyraskin:
**Story:** As a provider, I want a way to return to the reports list so I can reference a different report.

**Acceptance Criteria:**
- Adds a "Back to reports" link to the CDC report page (same icon used in enrollment)
- Clicking this link navigates the user to the reports list 

**Design:**
![image](https://user-images.githubusercontent.com/30734760/68957053-dcd63e00-078e-11ea-83bd-3cb30220ffa4.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens/390268232)

mmazanec22:
Grabbing this and the other back link ticket; making one component with rotate-able arrow

mmazanec22:
Addressed in #340

Issue #145, opened 2019-10-24
Add "back to roster" link to enrollment page


abbyraskin:
As a provider, I want a way to get back to the roster from the enrollment page.  

## Acceptance Criteria
- Adds "Back to Roster" link with icon to enrollment page, left-aligned
- Selecting "Back to Roster" navigates the user back to the roster page 

## Design
[Icon needed]

![image](https://user-images.githubusercontent.com/30734760/67499569-f7613f80-f646-11e9-8ab3-a0f8c06c45d1.png)

abbyraskin:
@atreni could you provide the .svg for the back icon when you have a chance?

mmazanec22:
Arrow SVG is included in USWDS library.

Note from the tech debt doc: "format enrollment detail; add  nav buttons (back to roster? Enroll another child? Confirm in design spec)"

mmazanec22:
Addressed by #340

PR #340, opened 2019-12-04
145 272 directional link

mmazanec22:
Addresses #145 and #272


src/Hedwig/ClientApp/src/components/DirectionalLink/DirectionalLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+// Have to have it inline or we can't inherit color
+
+type DirectionalLinkProps = {
+	direction: 'left' | 'right' | 'up' | 'down';
+	to: string;
+	text: string;
+	arrowSide?: 'left' | 'right';
+	ariaLabel?: string;
+	// We prob shouldn't use titles: https://a11yproject.com/posts/creating-valid-and-accessible-links/
+	// And we should generally write descriptive link text rather than using aria labels
```
cmgiven: Very much agree. I wonder whether you see a need for this option if the link test is being properly designed? And if not, maybe removing the option is a step toward enforcing best practices?

src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.tsx
```diff
@@ -28,27 +30,28 @@ export default function EnrollmentDetail({
...
-		orgId: (user && user.orgPermissions && user.orgPermissions[0] && user.orgPermissions[0].organizationId) || 1,
-		siteId: siteId ? parseInt(siteId) : 1,
+		orgId: (user && user.orgPermissions && user.orgPermissions[0] && user.orgPermissions[0].organizationId) || 0,
+		siteId: siteId ? parseInt(siteId) : 0,
 		include: ['child', 'family', 'determinations', 'fundings'],
 	}
 	const [loading, error, enrollment, mutate] = useApi<Enrollment>(
 		(api) => api.apiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet(params),
-		[enrollmentId],
+		[enrollmentId, user],
```
cmgiven: :+1:

src/Hedwig/ClientApp/src/components/DirectionalLink/DirectionalLink.tsx
mmazanec22: I think we could-- the reason I included this was because I saw that we're using both title and aria-label on the link for the app title in Header, even though the title itself is a string (as opposed to a logo) and having both title and aria-label might be annoying to someone using a screen reader.  So I kinda thought maybe there were use cases that could call for it?  Unless we need to change how we're doing the header?

cmgiven: That was just me copying and pasting from USWDS. I don't think I gave it much thought. If it's wrong, let's change it!

PR #344, opened 2019-12-04
328 datepicker logout fix

mmazanec22:
The main issue was that the default button type was submit in the context of a form.

### DIFF
--- a/src/Hedwig/ClientApp/src/components/Button/Button.tsx
+++ b/src/Hedwig/ClientApp/src/components/Button/Button.tsx
@@ -9,9 +9,17 @@ type ButtonProps = {
 	href?: string;
 	appearance?: ButtonAppearance;
 	disabled?: boolean;
+	type?: 'button' | 'submit' | 'reset' | undefined;
 };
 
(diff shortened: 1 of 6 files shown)


src/Hedwig/ClientApp/src/components/Button/Button.tsx
```diff
@@ -9,9 +9,17 @@ type ButtonProps = {
 	href?: string;
 	appearance?: ButtonAppearance;
 	disabled?: boolean;
+	type?: 'button' | 'submit' | 'reset' | undefined;
```
cmgiven: It doesn't look like we're using this? Do we need it?

mmazanec22: Default button type in a form is submit, and that's what was causing this issue.  We can remove this and just default to type="button" if you prefer.

cmgiven: That does make sense to me. I can't recall ever actually wanting submit. I recall lots of `e.preventDefault()`.

Issue #270, opened 2019-11-15
Add monthly utilization table to CDC report


abbyraskin:
**Story**: As a provider, I want to see how many kids were enrolled at my site on my reports to OEC.

**Acceptance Criteria:**
- The reports page shows a table with counts of the kids enrolled in an organization at the time the report was submitted (or that are currently enrolled, if the report has not been submitted).
- If an organization accepts CDC funds, it will have a number of spaces allocated to it. A space is specific to a particular age range (IT/preschool/school-age) and time (part-time/full-time). (Note that the same is true of School Readiness and other types of funding sources. Please name the model generically).
- Table header: Monthly Utilization
- Table rows
  - One row for each combination of age and time that the program has spaces.
  - Total
    - Sum of values for each column
- Table columns
  - Capacity
    - The number of spaces the organization has been allocated.
  - Utilization
    - Number of enrollments with a `Funding` that overlaps with the reporting period, disaggregated by age (set on the enrollment) and time (set on the funding).
    - Shown in red if the utilization number is lower than the capacity number
    - If the report has already been submitted, we need to use the historical data to show the number as of the submit date of the report.
  - Reimbursement Rate (Weekly)
    - Added in a separate ticket (#288) 
  - Total
    - Added in a separate ticket (#288)
  - Balance 
    - Added in a separate ticket (#288)

**Design:**
![image](https://user-images.githubusercontent.com/30734760/68957053-dcd63e00-078e-11ea-83bd-3cb30220ffa4.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens/390268232)

abbyraskin:
Chris to break out into separate stories (separating count of children from calculations)

PR #352, opened 2019-12-05
Adds API for #270

cmgiven:
Sorry for the whitespace mess.

* Adds a new endpoint, `GET api/organizations/[:orgId]/reports/[:reportId]`.
* Notably, this endpoint will use temporal data as of the report submit date if `enrollments` is included.
* `Fundings` now have entry and exit dates.
* An `Organization` can have one or more `FundingSpaces`, indicating its capacity.


src/Hedwig/Data/HedwigContext.cs
```diff
@@ -3,124 +3,128 @@
...
-
-			return currentUserId;
-		}
-	}
+      var subClaim = _httpContextAccessor.HttpContext.User.FindFirst("sub")?.Value;
+      if (subClaim == null) { return null; }
+      Guid wingedKeysId;
+      var isGuid = Guid.TryParse(subClaim, out wingedKeysId);
+      if (!isGuid) { return null; }
+      return Users.Where(user => user.WingedKeysId == wingedKeysId).Select(user => user.Id).Cast<int?>().FirstOrDefault();
```
jlhogan: :clap:

src/Hedwig/Repositories/ReportRepository.cs
```diff
@@ -7,57 +9,127 @@
...
+        .ToListAsync();
+    }
+
+    public async Task<Report> GetReportByIdAsync(int id) => await _context.Reports
+      .Include(r => r.ReportingPeriod)
+      .FirstOrDefaultAsync(r => r.Id == id);
+
+    public async Task<Report> GetReportForOrganizationAsync(int id, int orgId, string[] include = null)
+    {
+      var include_orgs = include.Contains(INCLUDE_ORGANIZATIONS);
```
jlhogan: personally, I don't find this any more clear than just calling `include.Contains(XXX)` inline in the if statement

src/Hedwig/Repositories/ReportRepository.cs
```diff
@@ -7,57 +9,127 @@
...
+      if (include_funding_spaces)
+      {
+        report = report
+          .Include(report => report.Organization)
+          .ThenInclude(organization => organization.FundingSpaces);
+      }
+
+      var reportResult = await report.FirstOrDefaultAsync();
+
+      if (include_enrollments)
```
jlhogan: comment this logic, at least the `AsNoTracking()`s if they're not gonna be obscured in some other function call

test/HedwigTests/Data/HedwigContextTests.cs
```diff
@@ -10,7 +10,7 @@ namespace HedwigTests.Data
 {
     public class HedwigContextTests
     {
-        [Fact]
+        [Fact(Skip = "Need to create a user with the same Guid as USER_CONTEXT_SUB")]
```
jlhogan: If you look at `Fixtures/Migrations/99999999999999_BootstrapTestData.cs`, you can see where i'm creating the user and should be easily able to update this to include to correct GUID if we make `USER_CONTEXT_SUB` some constant

test/HedwigTests/Helpers/FundingHelper.cs
```diff
@@ -9,7 +10,9 @@ public class FundingHelper
 			HedwigContext context,
 			FundingSource source = FundingSource.CDC,
 			FundingTime time = FundingTime.Full,
-			Enrollment enrollment = null
+			Enrollment enrollment = null,
+      string entry = "2000-01-01",
```
jlhogan: weird whitespace

test/HedwigTests/Helpers/FundingHelper.cs
```diff
@@ -18,9 +21,12 @@ public class FundingHelper
 			{
 				EnrollmentId = enrollment.Id,
 				Source = source,
-				Time = time
+				Time = time,
+        Entry = DateTime.Parse(entry)
```
jlhogan: weird whitespace

src/Hedwig/Repositories/ReportRepository.cs
cmgiven: Yeah. The reason to do it was that I use `include_enrollments` is being used multiple times, and I like consistency. But it does feel a bit boilerplate-y.

jlhogan: i would opt for repeating it once rather than defining all of these at the top, but its a preference and doens't actually matter :)

Issue #288, opened 2019-11-15
Monthly utilization table reports funding amounts


cmgiven:
**Story**: As a provider, I want to see how much OEC has paid me for a given reporting period, and if I'm underutilized, how much I will owe them.

**Acceptance Criteria:**
- All sites have a region, one of East (E), North Central (NC), Northwest (NW), South Central (SC), or Southwest (SW)
- All sites have a Title I status (boolean).
- Note that although CDC reports are at the organization level, these attributes are set at the site level. We'll address this complexity as part of addressing multiple-site organizations. Until then, we assume that all organizations only have a single site.
- Add additional columns to the table created in (#270).
  - Weekly Reimbursement Rate
    - Look up for each age/time from the below table. Use the top table if the accredited checkbox is checked on the report, bottom otherwise. Select the row based on the region of the site. Then select the cell based on the age and time, and whether the site is Title I. "wa" = part time. Ignore the "max" columns and whether school is in session; these things are not relevant to our purposes. Note that the right-most column contains a typo, it should be "pt" not "ft"
  - Total
    - The utilization times the reimbursement rate times the number of weeks in the reporting period.
  - Balance
    - The difference between the total column and the same calculation with the capacity instead of utilization.
    - If negative (utilization is less than capacity) color red and wrap in parentheses.

<img width="815" alt="Screen Shot 2019-11-15 at 4 38 50 PM" src="https://user-images.githubusercontent.com/5375557/68980841-4a05c580-07c7-11ea-959a-9b7ff31daaa8.png">

**Design:**
![image](https://user-images.githubusercontent.com/30734760/68957053-dcd63e00-078e-11ea-83bd-3cb30220ffa4.png)

[Invision prototype](https://skylight.invisionapp.com/share/QDUDSEK9MP8#/screens/390268232)

PR #367, opened 2019-12-07
Utilization table

cmgiven:
Connects #270, #288.

Still needed:

* Need to add `TitleI` and `Region` attributes to the `Site` model. They are currently hard-coded in the client.
* Need to add `Age` to the seed `Enrollments`. It's currently all null, which means that the report shows no one enrolled.
* Tests.

Screen Shot 2019-12-07 at 2 06 30 PM


src/Hedwig/ClientApp/src/containers/ReportDetail/UtilizationTable.tsx
```diff
@@ -0,0 +1,171 @@
...
+        count,
+        capacity,
+        rate,
+        total,
+        balance,
+      }
+    }));
+  });
+
+  const grandRow = rows.reduce((total, row) => {
```
jlhogan: grand ?

cmgiven: Like grand total? I don't know, probably should just be `totalRow`.

PR #368, opened 2019-12-08
346: Split roster by age, and update roster legend to match invision

jlhogan:
closes #346

Issue #307, opened 2019-11-26
Update roster legend UI


abbyraskin:
**Story:** As a provider, I want to understand what information is shown on my roster.

**Acceptance Criteria:**
- Update existing roster legend with the following (see design):
  - Count of CDC spaces filled
  - Count of children receiving Care 4 Kids
  - Missing information alert
- Note: remove border from legend shown in design

**Design:**
![image](https://user-images.githubusercontent.com/30734760/70179689-17ebd300-16a4-11ea-9bc0-8591ee6177ff.png)

[Invision prototype](https://skylight.invisionapp.com/public/share/QDUDSEK9MP8#/screens)

mmazanec22:
Planning to combine this with https://github.com/ctoec/ecis-experimental/issues/304

jlhogan:
@mmazanec22 i completed some of this in PR #369 but the styling is bad

mmazanec22:
Ok-- @jlhogan everything except the total number of spaces available is done in this one, and I also ditched the redundant SVGs: https://github.com/ctoec/ecis-experimental/tree/307-legend-ui-update

Issue #157, opened 2019-10-24
"Missing information" state on step list


abbyraskin:
As a provider, I want to know if I did not enter enrollment information that is required to submit the monthly report.

## Acceptance Criteria
- For each enrollment step, if required information is missing, display a "Missing information" save state (orange exclamation mark circle, see design below) 
  - Copy note: use sentence case for consistency, which differs from design - it should say "Missing information"
- Any required information that was not filled out should still be displayed in the save state with the missing information icon indicator (see design below)
- (Note that any blocking validations will be developed in separate stories - users can only get to this state after all blocking validations have been resolved while missing information is present)

## Design
![image](https://user-images.githubusercontent.com/30734760/71012554-ed8f1200-20b4-11ea-9cdc-920683d8e8a0.png)

[Invision prototype](https://skylight.invisionapp.com/public/share/QDUDSEK9MP8#/screens/475012311)

### Related stories
- "Completed" save state, #152

abbyraskin:
There may be some tweaks coming out of the conversation w/ Ashley today so moving this over to the backlog for now!

PR #369, opened 2019-12-08
Fake incomplete data

### LINKED ISSUE #365: Roster sorting cleanup
I think we learned that name should be sorted by last name?
I think there’s a desire to make the Funding column sortable, so I can group the kids who are in a particular set of slots.

jlhogan:
right now this simply adds the alert icons to the roster view, @victoriasuwardiman do we need to impl such that when a user goes in to update the data, the icon then goes away? or do we just want reaction to the icons in the roster?

also -- the legend looks less than great with my changes from this branch and https://github.com/ctoec/ecis-experimental/pull/368, someone please help :)
Screenshot from 2019-12-07 21-40-15

also closes #365  roster sorting cleanup

links #307

### DIFF
--- a/src/Hedwig/ClientApp/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/Hedwig/ClientApp/src/__snapshots__/storyshots.test.ts.snap
@@ -13,18 +13,25 @@ exports[`Storyshots Alert Error 1`] = `
     role="alert"
   >
     <div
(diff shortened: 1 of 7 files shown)


cailyncodes:
My commit fixes the legend spacing issue

victoriasuwardiman:
Looks good, just the icon next to a name on the roster works! (I think if we can fake an icon in the child summary page - #157, we'd be pretty much set up.)

PR #376, opened 2019-12-09
Makes reports submittable, with newly added fields on CDC Report

### LINKED ISSUE #167: Add success alert after submitting report
**Story**: As a provider, I want to be explicitly updated about the status of my report submissions

**Acceptance criteria**:
- success notification (like below) is displayed when report is successfully submitted
- a notification component is created, and used in `Report` component to achieve this

Screenshot from 2019-10-28 12-39-10

jlhogan:
Also adds some new fields to site

NOTE: theres a diff btwn this and invision, where accredited checkbox is below the utilization table. easier and cleaner to impl as part of the form fragment component that way, but if there's a hard need for it to be above we can do that.

see it live here: https://3.136.37.61:5001/ !!!!!!!!!!!!!!!!!

closes #167


src/Hedwig/ClientApp/src/swagger.json
```diff
@@ -934,32 +960,68 @@
...
                     "$ref": "#/components/schemas/Site"
                   }
+=======
+                  "$ref": "#/components/schemas/CdcReport"
+>>>>>>> [WIP] submit roster with new fields
                 }
               },
               "text/json": {
                 "schema": {
+<<<<<<< HEAD
```
cailyncodes: Eep 👀

src/Hedwig/Controllers/ReportsController.cs
```diff
@@ -46,5 +46,30 @@ int orgId
...
+        if(report.SubmittedAt == null) report.SubmittedAt = DateTime.UtcNow;
+        _reports.UpdateReport(report);
+        await _reports.SaveChangesAsync();
+      }
+      catch (DbUpdateConcurrencyException)
+      {
+        return NotFound();
+      }
+
+      return Ok(report);
```
cailyncodes: Is wrapping this in `Ok` necessary? I ask this as a very general question - I see some with it, some without

jlhogan: does not change the actual content of the response, but does change the type which makes testing cleaner.

src/Hedwig/ClientApp/src/swagger.json
jlhogan: blarg

src/Hedwig/Models/Region.cs
```diff
@@ -0,0 +1,7 @@
+namespace Hedwig.Models
+{
+  public enum Region
+  {
+    Default = 0
```
cmgiven: The regions are E, NC, NW, SC, SW. Also here: https://github.com/ctoec/ecis-experimental/blob/master/src/Hedwig/ClientApp/src/containers/ReportDetail/CdcRates.ts#L3

Should not block merging this in.

PR #379, opened 2019-12-10
Better seeding of data

cmgiven:
There's now a CSV that we can update to easily add kids, and the script itself will be easier to update going forward.


cmgiven:
(Also touches up the styling on the text fields. Just slipping that in here.)

Issue #277, opened 2019-11-15
Reports list is organized by pending and submitted reports


abbyraskin:
**Story:** As a provider, I want to easily distinguish between reports that are still in progress, and reports that I previously submitted to OEC.

**Acceptance Criteria:**
- Breaks the reports list into two sections:
  - "Pending reports" (those that have not yet been submitted or are being actively revised)
  - "Submitted reports" (reports that have been sent to OEC)
- If the pending reports section is empty, include the following copy in italics (note: use copy below, not the copy shown in design):
  - The next CDC report will be available at the end of the reporting period.  

**Design:**
![image](https://user-images.githubusercontent.com/30734760/70264143-7d4fca80-175d-11ea-9bb9-4ada6f5fd1a7.png)

abbyraskin:
Is that helper copy accurate? (Not sure we've built it this way, so just want to make sure!)

abbyraskin:
Synced w/ team on this - given that we want to generate reports as soon as the last reporting period ends, will clarify language in AC

PR #385, opened 2019-12-10
Reports list is organzed by pending and submitted reports

cmgiven:
Connects #277 

In addition:
* Updates the nav bar to use a count of unsubmitted reports, per designs
* Refreshes the nav bar after report submit
* Removes unnecessary state from ReportSubmitForm (`useApi` takes care of it)
* Unifies currency formatting and text input validations, closing out some todos left as comments


src/Hedwig/ClientApp/src/containers/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -6,123 +6,127 @@ import UserContext from '../../contexts/User/UserContext';
...
-			accredited,
-			c4KRevenue,
-			retroactiveC4KRevenue,
-			familyFeesRevenue,
-			submittedAt
-		};
-	}
+  const params: ApiOrganizationsOrgIdReportsIdPutRequest = {
+    id: report.id || 0,
+    orgId: idx(user, _ => _.orgPermissions[0].organizationId) || 0
```
jlhogan: use user id util

src/Hedwig/ClientApp/src/containers/Reports/Reports.tsx
```diff
@@ -1,96 +1,114 @@
...
+          <th scope="row">
+            <Link to={`/reports/${row.id}`} className="usa-link">
+              {monthFormatter(row.reportingPeriod.period)}
+            </Link>
+          </th>
+        ),
+        sort: row => row.reportingPeriod.period && row.reportingPeriod.period.getMonth() || -1,
+      },
+      {
+        name: 'Type',
```
mmazanec22: I'm sure there's some ticket about this and that this page isn't done, but just flagging that right now we're showing abbreviations without spelling them out in the content: https://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-located.html

cmgiven: I don't think there's a ticket for that. I was thinking about that one. I'll create.

mmazanec22: I'm doing some research rn into legend best practices-- whether we should associate it with the tables in some way, or associate the values with the labels, or just have an aria label for the tags within the table, or what.  Like many data-related questions, it's not abundantly clear in WCAG :/

PR #386, opened 2019-12-10
Add some comments and some clean up

cailyncodes:


src/Hedwig/ClientApp/src/index.tsx
```diff
@@ -20,6 +20,10 @@ const render = (Component: React.FC) => {
 				localStorageKey="hedwig-key"
 				// NOTE: "offline_access" is required in scope string to retrieve refresh tokens
 				scope="openid profile hedwig_backend offline_access"
+				extras={{
+					// NOTE: Required for refresh tokens
+					access_type: "offline"
+				}}
```
jlhogan: have we confirmed that both `offline_access` in scope AND `access_type: offline` extra are needed? I see a world in which i added `offline_access` extraneously when implementing refresh logic

cailyncodes: I know it's required in scope but am not sure on the `extras` field. Will find out though :)

cailyncodes: It looks like it is necessary (at least according to the spec)

PR #389, opened 2019-12-10
303 family income not disclosed

### LINKED ISSUE #303: Add "Family income not disclosed" checkbox to funding step of enrollment
**Story:** As a provider, I want to enter private pay children who I may not have reported income information.

**Acceptance Criteria:**
- Adds a checkbox to the family income determination step of enrollment, with the label "Family income not disclosed"
- If this checkbox is selected, collapse the additional fields in this section and display a headerless information alert, with the following copy:
  - Income information is required to enroll a child in a CDC funded space. You will not be able to assign this child to a funding space without this information.
- If the checkbox is not selected (or de-selected), return to the previous state (with all fields showing)
- If the user selects "save" and the "income not disclosed" checkbox is selected, display an alert on the next step (step 4, see design below), with the following copy:
  - Funding options not available because family income was not provided. To change, edit the previous section.

**Design:**
image

image

jlhogan:
checkbox component on it's own has no borders/margins, so it looks wrong. need style help!

current PR env url: https://3.15.227.95:5001/

closes #303


src/Hedwig/ClientApp/src/containers/Reports/Reports.tsx
```diff
@@ -9,8 +9,8 @@ import UserContext from '../../contexts/User/UserContext';
 import useApi from '../../hooks/useApi';
 import { 
 	ApiOrganizationsOrgIdReportsGetRequest,
-	Report,
-	ApiOrganizationsIdGetRequest
+	ApiOrganizationsIdGetRequest,
+	CdcReport
```
cailyncodes: I just merged into master a fix for this. I just did `import { CdcReport as Report }` and everything seemed good. Not sure if we need to relay `Cdc` report all the way down into the actual code?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -54,6 +57,19 @@ const EnrollmentFunding: Section = {
     const [entry, updateEntry] = React.useState(enrollment ? enrollment.entry : null);
     const [age, updateAge] = React.useState(enrollment ? enrollment.age : null);
 
+    const familyDeterminationDisclosed = (enrollment: Enrollment) : boolean => {
```
jlhogan: should this logic be etraxted into a familyDeterminationUtils?
also, functionally right now:
if no determination, or determination exists with notDisclosed = true => prohibit user from adding funding source
if determination exists with notDisclosed = false => allow user to add funding source, and enrollment (at least re: funding /income ) is considered valid

but possible that desired functionality is:
if no determination => allow user to add funding source, but enrollment is incomplete b/c funding exists w/out income determination
if determination exists with notDisclosed = true => prohibit user from adding funding source
if determination exists with notDisclosed = false => allow user to add funding source, and enrollment (at least re: funding /income ) is considered valid

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -7,39 +7,42 @@ import dateFormatter from '../../../utils/dateFormatter';
...
-					<>
-						<p>Household size: {determination.numberOfPeople}</p>
-						<p>Annual household income: ${idx(determination, _ => _.income.toFixed(2))}</p>
-						<p>Determined on: {dateFormatter(idx(determination, _ => _.determined))}</p>
-					</>
-				) : (
-					<p>No income determination on record.</p>
-				)}
+				{determination ? 
+					determination.notDisclosed ? (
```
jlhogan: are these chained ternaries too annoying to decipher?

cailyncodes: I was just about to leave a comment on this. I don't love it, but am not vehemently against it lol.

src/Hedwig/ClientApp/src/containers/Reports/Reports.tsx
jlhogan: we don't need to now. we will eventually, but that'llbe part of a larger work stream to handle multiple types of reports. Your change is fine by me, i'll rebase and opt for that

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -119,19 +134,29 @@ const EnrollmentFunding: Section = {
...
-					</li>
-          <li>
-            <Button appearance="unstyled" text="Add Care 4 Kids subsidy" />
-          </li>
-        </ul>
-
+        {!familyDeterminationDisclosed(enrollment) &&
+          <Alert 
+            type="info"
+            text="Funding options not available because family income was not provided. To change, edit the previous section"
```
mmazanec22: Should we add a period at the end of this sentence?  Should it say what section rather than "the previous"?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -105,39 +111,63 @@ const FamilyIncome: Section = {
...
+		  			</label>
+		  			<DatePicker
+		  				onChange={range =>
+		  					updateDetermined((range.startDate && range.startDate.toDate()) || null)
+		  				}
+		  				dateRange={{ startDate: determined ? moment(determined) : null, endDate: null }}
+		  			/>
+		  			</>
+		  		)}
+		  		<Checkbox
```
mmazanec22: I wonder if we should use a checklist rather than a singleton checkbox even though it's cumbersome?  That might fix spacing woes.  Or you could just wrap it in a usa-form-group div.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
mmazanec22: I would support this being a util.  I also wonder if we should think about organizing our utils differently-- maybe grouping them in files, and then exporting them in an index file to make the imports less chaotic?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
jlhogan: will try ty !

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cmgiven: > CDC funding requires family income to be provided. To add CDC funding, please update family income.

Note that this will be updated when new designs for funding become available.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cmgiven: Yes, I think the second description is correct.

I agree @mmazanec22 on refactoring the utils. That and breaking out the CSS into per-component files that live inside each component's  directory will be very useful clean up.

jlhogan: https://github.com/ctoec/ecis-experimental/issues/420
https://github.com/ctoec/ecis-experimental/issues/419

abbyraskin:
Could we make both of these alerts the small alert style (no header)? (We decided to remove the header from the first one after the AC were written - if it's too late we we can make the change later, no worries).

For some reason the styling of the (smaller/headerless) alert seems off - maybe there is extra spacing on the top/bottom? It feels large. This is probably not worth being nit picky about at this stage and it passes without this change, just wonder if we're using the right component. (Alert component without a heading).

And then really getting picky here, add a period to the end of the second alert please.

jlhogan:
> Could we make both of these alerts the small alert style (no header)? (We decided to remove the header from the first one after the AC were written - if it's too late we we can make the change later, no worries).

yes absolutely!  my bad. I looked in the ticket (https://github.com/ctoec/ecis-experimental/issues/303) for updated AC, and the header is still there. Is there somewhere else to be looking for most current AC? 


> For some reason the styling of the (smaller/headerless) alert seems off - maybe there is extra spacing on the top/bottom? It feels large. This is probably not worth being nit picky about at this stage and it passes without this change, just wonder if we're using the right component. (Alert component without a heading).

we only have one component currently implemented in the app, and that's an alert with a header. I didn't see a different one in USWDS, but we can open a ticket to tweak styling when header is not present. I'm just a total oaf when it comes to styles and am gonna punt that to someone else :)

jlhogan:
> The checkbox styling looks great! Approved for that. I think there is a separate ticket for this alert correct? Have some feedback on that.

@atreni not sure what you mean by separate ticket for this alert? this specific alert, or styling of the alert component in general? Not sure that another ticket for this exists, but if more work is needed here let's make sure it does

PR #406, opened 2019-12-11
134: POC for model validation

### LINKED ISSUE #134: Validation of models
**Story:** As an OEC stakeholder, I want to define rules that flag obviously bad data to providers, or that require certain information to be entered before a report can be submitted, or that require information to be updated on a regular cadence.

**Tech discussion**

**Different types of rules.** Here's an interesting scenario: Sometimes providers will want to enter a child into the system, but they don't yet have all the information. Let's say they don't have the kid's address. Now address is a "required" field. We want to ensure that any kid receiving funding has that info. But we also want someone to be able to enter the information they have, and come back to it.

This leads me to conclude we probably need a two-tier validation system.

* Some rules prevent saving information. I shouldn't be able to say a kid was born in 1915. If I try to save a kid with that birthday, I'll be barred from going any further until I correct that mistake.
* Some rules allow the information to be saved, but show a warning, and would prevent submitting a report including that information. The required address scenario above would be an example. Requiring an income determination in the last 12 months is another example.

When we run the warning rules, we should also run the hard and fast rules. This would support adding a new rule that existing data violates. The data was valid at the time it was saved, but now you'll need to correct it before you can submit any reports using it.

**Server-side vs. client-side.** Ideally, we would only write rules once. With this constraint (and the possibility of API access in the future), I think the only place to perform validation is on the server side.

The API, both when fetching data and when mutating it, should be capable of returning a list of issues with a record. This would allow the UI to flag records that need some sort of action, and during data entry flows, highlight fields with known errors.

**Developer ergonomics.** Today in ECIS, all such rules are buried in a ginormous file called BusinessRules.cs. Rules call out to other rules. It's a mess. Let's not do the same.

I'd love an interface such that each rule is its own class, lives in its own file, and is independently tested. Each rule could perhaps define whether it is a blocking rule for purposes of the two-tier validation described above. Somewhere, probably on the model, we provide a list of the rules that apply to it (or maybe this can be done automagically using the type system?). And then the API controllers need to ensure that those checks are run when mutating data, or when the client requests a list of known issues.

jlhogan:
Api response from a test family controller we whipped up to integration test that validated a family with an out-of-date determination

todo: figure out how programatically get the names of the fields on the models (i thoguht we were doing something that works, but looking now i see `DateTime` that should be `Determined`

```
{                                                                                 
    "addressLine1": null,                                                         
    "addressLine2": null,                                                         
    "author": null,                                                               
    "authorId": null,                                                             
    "children": null,                                                             
    "determinations": [                                                           
    ¦   {                                                                         
    ¦   ¦   "author": null,                                                       
    ¦   ¦   "authorId": null,                                                     
    ¦   ¦   "determined": "2019-01-01T00:00:00",                                  
    ¦   ¦   "familyId": 2906,                                                     
    ¦   ¦   "id": 2898,                                                           
    ¦   ¦   "income": 20000.0,                                                    
    ¦   ¦   "notDisclosed": false,                                                
    ¦   ¦   "numberOfPeople": 3,                                                  
    ¦   ¦   "validationErrors": [                                                 
    ¦   ¦   ¦   {                                                                 
    ¦   ¦   ¦   ¦   "field": "DateTime",                                          
    ¦   ¦   ¦   ¦   "message": "Family income must be determined in past 6 months"
    ¦   ¦   ¦   }                                                                 
    ¦   ¦   ]                                                                     
    ¦   }                                                                         
    ],                                                                            
    "homelessness": false,                                                        
    "id": 2906,                                                                   
    "organization": null,                                                         
    "organizationId": 50,                                                         
    "state": null,                                                                
    "town": null,                                                                 
    "validationErrors": [                                                         
    ¦   {                                                                         
    ¦   ¦   "field": "FamilyDetermination",                                       
    ¦   ¦   "message": "Determination has validation errors"                      
    ¦   }                                                                         
    ],                                                                            
    "zip": null                                                                   
}                                                                                 
```

closes #134


src/Hedwig/Repositories/FamilyDeterminationRepository.cs
```diff
@@ -12,11 +12,11 @@ public class FamilyDeterminationRepository : TemporalRepository, IFamilyDetermin
 	{
 		public FamilyDeterminationRepository(HedwigContext context) : base(context) {}
 
-		public Task<List<FamilyDetermination>> GetDeterminationsByFamilyIdAsync(int familyId)
+		public List<FamilyDetermination> GetDeterminationsByFamilyId(int familyId)
```
jlhogan: @cailyncodes  the simplest of fixes!!! since multi-threading db operations is not allowed anyway!

src/Hedwig/Validations/Rules/Family/DeterminationIsValid.cs
```diff
@@ -0,0 +1,45 @@
+using Hedwig.Models;
+using Hedwig.Repositories;
+using System.Linq;
+
+namespace Hedwig.Validations.Rules
+{
+  public class DeterminationIsValid: SubObjectIsValid, IValidationRule<Family>
+  {
+    //TODO: can we get around having to define private base class constructor param?
```
cmgiven: It kind of feels like the base class isn't adding much? Mostly just the constructor to stash the injected validator.

cmgiven:
What's here looks really good to me.

src/Hedwig/ClientApp/src/generated/models/ValidationError.ts
```diff
@@ -0,0 +1,63 @@
...
+     * @type {string}
+     * @memberof ValidationError
+     */
+    readonly field?: string | null;
+    /**
+     * 
+     * @type {string}
+     * @memberof ValidationError
+     */
+    readonly message?: string | null;
```
cailyncodes: Do we want to type the backend model so as to prevent these fields from being nullable?

src/Hedwig/ClientApp/src/generated/models/ValidationError.ts
```diff
@@ -24,13 +24,13 @@ export interface ValidationError {
...
      */
-    readonly field?: string | null;
+    readonly field: string | null;
     /**
      * 
      * @type {string}
      * @memberof ValidationError
      */
-    readonly message?: string | null;
+    readonly message: string | null;
```
cailyncodes: 😢 Not quite it seems

jlhogan: ugh! They're required fields in the openAPI spec, but listed as nullable, i think because string type is reference type and can be null? So something like this will be required to resolve the nullable issue: https://stackoverflow.com/questions/46576234/swashbuckle-make-non-nullable-properties-required

PR #407, opened 2019-12-12
Change the name to ECE Reporter

cmgiven:
We have a name!


jlhogan:
Happy to merge this, but there are definitely other `hedwig`s in the repo, for instance `HedwigRepository` base class etc etc.

cmgiven:
Yeah, figured it wasn't necessary to excise the codename everywhere. Maybe we want to come back to that later?

cailyncodes:
Do we want to change the repo name on Github, too?

cmgiven:
Ooh, good call.

cmgiven:
But actually, let's hold off on that until we check with @speedcm that that won't break anything.

Issue #330, opened 2019-12-03
Error handlings in form validation failures


mmazanec22:

cailyncodes:
Some musings for implementation here https://github.com/ctoec/ecis-experimental/pull/410/commits/bebdfaa1833095165e065eed7d8db01190f091da

PR #410, opened 2019-12-12
POC/Initial idea for #330

cailyncodes:
Thoughts?


cailyncodes:
I don't think we can have the save button disabled at any point. Unless we implement the rules in the front-end, we have to submit to the backend to be able to assess errors.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -212,6 +213,8 @@ const ChildInfo: Section = {
       }
     };
 
+    const disabledSave = (enrollment && enrollment.validationErrors && enrollment.validationErrors.length > 0) || false;
```
jlhogan: I don't think these validation errors should ever block save, except on submitting reports. I believe we  want to use these errors to indicate to users trying to complete enrollments to make report submittable which fields are blocking the completion and why/how

cmgiven: I think that's right, at least to start.

The exception is that we need first name and last name to save the first screen of the enrollment flow. But I think we just block that on the front-end.

jlhogan: first name and last name are REQUIRED, straight up. That'll be determinable frm the schema (once strings are non-nullable), so you won't be able to save unless those values are there b/c the post params will be invalid.

jlhogan: see https://github.com/ctoec/ecis-experimental/pull/414/files/9dda8cf1c629d71eff74d368367f4dd66f30d6bc#diff-7f26e8a7e273351f2063a941192d32a5R163

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -70,6 +70,11 @@ export default function EnrollmentNew({
 
 		const currentIndex = sections.findIndex(section => section.key === sectionId);
 
+		// If there are validation errors, do not move to the next section
+		if (enrollment.validationErrors && enrollment.validationErrors.length > 0) {
+			return;
+		}
+
```
cailyncodes: On like 4th thought, I don't think we want this. At least on the enrollment flow logic.

cailyncodes:
I added some additional logic for better UI experience (and to move away from hard coded missingInformation birth certificate logic).

There still might be some areas to address.

src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.tsx
```diff
@@ -79,9 +79,9 @@ export default function EnrollmentDetail({
 								Edit<span className="usa-sr-only"> {section.name.toLowerCase()}</span>
 							</Link>
 							{/* TODO: when we figure out the missing information logic, remove hard coding of section */}
-							{section === ChildInfo && informationIsMissing && (
+							{invalidEnrollment && enrollment.validationErrors && enrollment.validationErrors.find(error =>  section.fields && section.fields.includes(error.field ? error.field : '')) && (
```
cailyncodes: This line needs some love haha

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
cmgiven: Discussed this with @abbyraskin yesterday. The conclusion is that we probably do want to block moving to the next section, but not for missing information errors. So what's emerging there is that we want two kinds of errors.

1. Missing information. Ah well, could be intentional. We'll let 'em keep moving but mark the section as wrong.

2. The information entered is bad. We'll save it to the database, because we want to be nice about it, but we're going to ask you to correct it before moving on. If you're in the enrollment flow, we won't go to the next section, and instead will keep you on the current section with an alert and validation errors highlighted on the specific fields. If you're in edit, instead of taking you back to the detail view, we'll leave you on the edit screen with an alert and validation errors highlighted.

cmgiven: cc @jlhogan

jlhogan: why are we going to save it to the database if we're going to block continuing to the next section? i think we can/should impl blocking validations as dotnet model validations. That means we only have to maintain 2 types of validations: ours which report validationerrors, and dotnet core which result in a 400 response to your post

jlhogan:
closed in favor of https://github.com/ctoec/ecis-experimental/pull/446

PR #414, opened 2019-12-12
326: Makes strings, date-times, and uuids non-nullable

### LINKED ISSUE #326: Why are strings nullable when they should be required?
Fix this and undo some of the typing gymnastics we did to accommodate it:
* nameFormatter should only accept type Name
* rawEnrollments/enrollments type mapping in Roster

jlhogan:
closes #326


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -159,35 +160,46 @@ const ChildInfo: Section = {
...
-        birthState,
-        americanIndianOrAlaskaNative,
-        asian,
-        blackOrAfricanAmerican,
-        nativeHawaiianOrPacificIslander,
-        white,
-        hispanicOrLatinxEthnicity,
-        gender,
-      };
+    const [validArgs, setValidArgs] = useState(child);
```
jlhogan: @cailyncodes  @cmgiven  fyi: this is how i was imagining save/no save is determined.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -382,7 +394,7 @@ const ChildInfo: Section = {
           onChange={event => updateGender(genderFromString(event.target.value))}
         />
 
-        <Button text="Save" onClick={save} />
+        <Button text="Save" onClick={save} disabled={!!!validArgs}/>
```
cailyncodes: I think we can just use one `!`

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -159,35 +160,46 @@ const ChildInfo: Section = {
...
+      hispanicOrLatinxEthnicity,
+      gender,
+    };
 
-      if (enrollment) {
+    useEffect(() => {
+      // Ideally I wish this could be a run-time type check
+      // (i.e. if we can cast args to child, then its valid, otherwise not valid),
+      // but internet searches are indicating ts standard is user-defined type predicates :(
+      if(args.firstName && args.lastName) {
```
cailyncodes: I can get behind this approach (broadly on save/no save). Given this (disable save) is only ever used in a small number of cases, I think it works! I would still suggest we create a util for this (this line), something like `childCanBeSaved` or `childIsNonBlockingValid` (or similar) to provide semantics for what is happening.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -55,11 +55,13 @@ const FamilyIncome: Section = {
     const child = enrollment.child;
     const determination = idx(child, _ => _.family.determinations[0]) || undefined;
     const [numberOfPeople, updateNumberOfPeople] = React.useState(
```
cailyncodes: Can we import `useState` from react and change `React.useState` to just `useState`?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -111,7 +113,7 @@ const FamilyIncome: Section = {
           label="Household size"
           defaultValue={numberOfPeople ? '' + numberOfPeople : ''}
           onChange={event => {
-            const value = parseInt(event.target.value.replace(/[^0-9.]/g, ''), 10) || null;
+            const value = parseInt(event.target.value.replace(/[^0-9.]/g, ''), 10) || undefined;
```
cailyncodes: Should we be automatically stripping out non-number characters? Or should we add this check to the `firstName` `lastName` checks we already run above? That is, treat this as blocking save?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
cailyncodes: Is there a way to surface an error message with this approach?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
cailyncodes: I think this will allow an invalid save.

1) User enters first and last name
2) validArgs is updated to a value
3) User removes first and last name
4) setValidArgs won't be set because the check in the condition will fail
5) validArgs is still a value
6) Put/Post request can be made but then errors on server side ef validation

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -202,7 +214,7 @@ const ChildInfo: Section = {
           enrollment: {
             id: 0,
             siteId: 0,
-            child: { ...args }
+            child: validArgs
```
cmgiven: Note that if validArgs is falsey, we'll end up here, attempting to submit a POST.

jlhogan:
@cmgiven yes, not necessarily by design but that's how the not-supplied-values mappings btwn dotnet and generated typescript play out by leveraging the optional `?` in typescript to denote values on API models that are not required (which then by defn have to be nullable in the DB, there's no undefined in dotnet) 

do you think that is a problem? to me, these models make more sense and i think will definitely be easier to interact with


@ everyone -- would we be ok with merging this and opening a separate ticket to deal with the argument validation/save blocking. I just threw that in here to demonstrate what i was talking about on @cailyncodes's other PR for validation error handling. I think it would be good to get new model defns in so as we continue building more functionality we can be freed from all the null | undefined shenanigans

cmgiven:
Nope. I think that the undefined semantics work out better in a lot of ways in the Javascript. Just making sure I was understanding the theory behind what this does.

Makes sense to me to merge.

cailyncodes:
Agreed with Chris!

jlhogan:
https://github.com/ctoec/ecis-experimental/issues/418

PR #424, opened 2019-12-13
Align baselines (Issue #392)

### LINKED ISSUE #392: align `enroll child` CTA with program title

cailyncodes:
Should close #392 

URL: https://3.133.92.106:5001


cailyncodes:
Not quite sure what this error message is? Another set of eyes would be greatly appreciated!

cmgiven:
We use snapshots for testing some of the pages and components. It basically records the markup that's generated, and if anything changes, requires you to go in a verify that you actually wanted the change.

Run `./dc-client yarn test` to see what's failing, and if the changes are as you intend, run `./dc-client yarn test -u` to update the snapshots.

cailyncodes:
Sorry, yes. There is another error besides snapshot. Forgot to call that out specifically

cmgiven:
Ah, missed that. What was the error?

cailyncodes:
`TypeError: wrapper.find(...).first(...).props(...).onClick is not a function

      166 |                             .first()
      167 |                             .props()
    > 168 |                             .onClick();
          |                              ^
      169 |                     await waitForUpdate(wrapper);
      170 |             });
      171 |

      at onClick (src/containers/Roster/Roster.test.tsx:168:6)`

cmgiven:
Looks like it was intended to be capturing a button related to the previous enrollments dialog, but is now instead capturing the enroll child button. We'll want to be more specific with the selectors. The particular issue is that it's expecting to find a prop passed to the component called `onClick`, which the enroll child button doesn't have because it instead uses `href`.

cailyncodes:
Ah that makes a lot of sense! Tysm!!!

PR #426, opened 2019-12-13
Update models to actually reflect real relationships

jlhogan:
makes required Ids required, and non-nullable ids non-nullable by
removing cascade delete behavior, replacing it with restrict. Lots of
changes, and means we have to supply default values for ids on object
creation but i think we abstract all that into utils and this is better


src/Hedwig/ClientApp/src/generated/models/CdcReport.ts
```diff
@@ -67,7 +67,7 @@ export interface CdcReport {
      * @type {number}
      * @memberof CdcReport
      */
-    organizationId?: number;
+    organizationId: number;
```
cailyncodes: Do you know why this has no `?` but in child there is one?

src/Hedwig/Models/Family.cs
```diff
@@ -16,9 +19,7 @@ public class Family : TemporalEntity, IValidateable
 		public bool Homelessness { get; set; }
 		public ICollection<FamilyDetermination> Determinations { get; set; }
 
-		// Optional FK to prevent cascade delete
-		// (multiple cascade delete FKs disallowed by SQLServer due to potential for cycles)
-		public int? OrganizationId { get; set; }
+		public int OrganizationId { get; set; }
```
cailyncodes: Should this be required?

src/Hedwig/ClientApp/src/generated/models/CdcReport.ts
jlhogan: right now, assumption in creating child/family is that user does _not_ provide the orgid, and we pull it from path params; i think we want to do this for report as well, so will mark orgId _not_ required on organizationReport

src/Hedwig/Models/Family.cs
jlhogan: see above comment. either/or

src/Hedwig/ClientApp/src/generated/models/Child.ts
```diff
@@ -46,24 +46,12 @@ import {
...
-     * @memberof Child
-     */
-    organization?: Organization;
     /**
      * 
      * @type {string}
      * @memberof Child
      */
-    id?: string;
+    id: string;
```
cailyncodes: We have three different types here and I'm not sure I know the difference.

cailyncodes: id, sasid, firstName that is

src/Hedwig/Models/FundingSpace.cs
```diff
@@ -1,14 +1,17 @@
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
+using System.ComponentModel.DataAnnotations;
 
 namespace Hedwig.Models
 {
   public class FundingSpace
   {
     public int Id { get; set; }
```
cmgiven: Should this be required as well?

src/Hedwig/Repositories/ChildRepository.cs
```diff
@@ -88,7 +84,7 @@ public async Task<Child> GetChildByIdAsync_OLD(Guid id, DateTime? asOf = null)
 		public async Task<ILookup<int, Child>> GetChildrenByFamilyIdsAsync(IEnumerable<int> familyIds, DateTime? asOf = null)
 		{
 			var children = await GetBaseQuery<Child>(asOf)
-				.Where(c => c.FamilyId != null && familyIds.Contains((int) c.FamilyId))
+				.Where(c => familyIds.Contains((int) c.FamilyId))
```
cmgiven: Hallelujah.

cailyncodes:
Closed by #433

src/Hedwig/Models/FundingSpace.cs
jlhogan: yes! ah! will sneak that in somewhere

PR #436, opened 2019-12-16
434: AuthorId added to entire temporal entity object tree

### LINKED ISSUE #434: get authorId working

jlhogan:
closes #434

### DIFF
--- a/entrypoint-test.sh
+++ b/entrypoint-test.sh
@@ -1,14 +1,18 @@
+#!/bin/sh
+
 set -e
 echo "moving test migrations into app..."
 testMigrations=$(ls test/HedwigTests/Fixtures/Migrations/*)
 cp test/HedwigTests/Fixtures/Migrations/* src/Hedwig/Migrations
 
-echo "installing dotnet ef utility..."
-until dotnet tool install --global dotnet-ef --version 3.0.0; do
-    sleep 1
-done
-export PATH="$PATH:/root/.dotnet/tools"
+if [ -z "$(dotnet tool list -g | grep dotnet-ef)" ]; then
+  echo "installing dotnet ef utility..."
+  until dotnet tool install --global dotnet-ef --version 3.0.0; do
+      sleep 1
+  done
+fi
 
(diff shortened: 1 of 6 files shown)


entrypoint-test.sh
```diff
@@ -1,14 +1,18 @@
...
 echo "moving test migrations into app..."
 testMigrations=$(ls test/HedwigTests/Fixtures/Migrations/*)
 cp test/HedwigTests/Fixtures/Migrations/* src/Hedwig/Migrations
 
-echo "installing dotnet ef utility..."
-until dotnet tool install --global dotnet-ef --version 3.0.0; do
-    sleep 1
-done
-export PATH="$PATH:/root/.dotnet/tools"
+if [ -z "$(dotnet tool list -g | grep dotnet-ef)" ]; then
```
cailyncodes: Just curious: what does this do?

src/Hedwig/Data/HedwigContext.cs
```diff
@@ -100,20 +101,32 @@ private bool IsTemporalEntityType<T>()
...
+      }
     }
 
     /// <summary>
     /// Gets the current UserId from HttpContext.User,
     /// or null if no UserId could be retrieved
     /// </summary>
     /// <returns></returns>
-    protected int? GetCurrentUserId()
+    protected virtual int? GetCurrentUserId()
```
cailyncodes: Why the change here?

jlhogan: for test :(

entrypoint-test.sh
jlhogan: running the test project on my local, i was running in infinite loop of trying to install the ef tool that alraedy existed. This makes it so it will only try to install the tool if it is not installed, which means i can run this on my local.

src/Hedwig/Data/HedwigContext.cs
cailyncodes: 😢

PR #440, opened 2019-12-16
Skip family income for foster

### LINKED ISSUE #155: Income information is not required when enrolling a foster child
As a provider, I want to skip the income determination section of enrollment if I am entering a foster child because the child is automatically eligible for state funding.

## Acceptance Criteria
- If a user has selected the "Child lives with foster family" checkbox in the family information section of enrollment (step 2) and selects save, skip the family income section and navigate directly to the enrollment and funding section of the step list (and show step 4 in an edit state)
- On save, step 2 is shown in a complete state (note: this differs slightly from design shown in prototype, avoids adding a third save state)
- Step should should include the following copy:
  - Household income: This information is not required for foster children

### Design
image
[Invision](https://skylight.invisionapp.com/public/share/QDUDSEK9MP8#/screens/475071032)

cailyncodes:
Should close #155.

I wasn't 100% clear what the requirements were because of conflicts comments/comps in the issue discussion. Let me know if anything needs to change on this PR.

URL: https://18.223.188.88:5001

### DIFF
--- a/src/Hedwig/ClientApp/src/components/StepList/Step.tsx
+++ b/src/Hedwig/ClientApp/src/components/StepList/Step.tsx
@@ -3,7 +3,7 @@ import { Link } from 'react-router-dom';
 import InlineIcon from '../InlineIcon/InlineIcon';
 import ErrorBoundary from '../ErrorBoundary/ErrorBoundary';
 
-export type ExternalStepStatus = 'incomplete' | 'complete' | 'attentionNeeded';
+export type ExternalStepStatus = 'incomplete' | 'complete' | 'attentionNeeded' | 'exempt';
 
 export type InternalStepStatus = 'notStarted' | 'active' | ExternalStepStatus;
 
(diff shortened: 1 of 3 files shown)


cailyncodes:
@cmgiven Re Point 2 of @atreni's comment, do you have any ideas for intelligent scroll positioning while stepping through the enrollment flow. We could attach an afterSave handler that updates the scroll position but it would require some calculation of position for the steps which may be variable and dynamic height. I spoke with Ashley yesterday that this could be a fast-follow item but just wanted to flag it if you have ideas on how to implement it cleanly.

cmgiven:
Potentially we should consider each move to the next section as the loading of a new page. The URL updates, an item is added to history, etc. The only difference is that we preserve scroll position.

Which is all to say: perhaps what we're actually looking to do is just reset the scroll to the top of the page?

cailyncodes:
Ooh I like that -- I don't think that's for this PR (because the page bounce occurs in all instances just not as noticeably) but something to consider in the StepList

cailyncodes:
@atreni This should solve point 1 of your comment. I'm leaving point 2 as a fast-follow fix implementing Chris's suggestion.

PR #442, opened 2019-12-16
418: Implements blocking model validations in client for child

### LINKED ISSUE #418: handle blocking model validations in enrollment flow
Half-implemented in `ChildInfo` section, but should be extracted into util and made to work correctly (right now, will not update save disabled state if first/last name is removed)

also, do we want to do any styling to make it more obvious to users what fields are required to enable save? Or just lack of `(optional)` marker is sufficient?

jlhogan:
closes #418 

NOTE: birthdate format validation cannot be implemented without changing how we're doing this input. Right now, the state variable for birthdate is only updated when a user uses the datepicker, not when they type in a date. Do we want/need to use date picker here? (cc @mmazanec22 @cmgiven)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -168,26 +102,39 @@ const ChildInfo: Section = {
...
+      if(childArgsAreValid(args)) {
+        updateValidArgs(args as Child);
+      } else {
+        updateValidArgs(undefined);
+      }
+    }, [firstName, lastName, birthdate]);
+
+    const save = () => {
+      if(!validArgs) {
+        // apply styling to missing fields
```
cailyncodes: Is this a TODO?

src/Hedwig/ClientApp/src/utils/modelUtils/childUtils.ts
```diff
@@ -0,0 +1,57 @@
+import { Child } from '../../generated';
```
cailyncodes: Can this file/folder path be something like `/utils/models/child.ts` instead?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
jlhogan: yup, no tickets with AC for this yet

PR #446, opened 2019-12-16
330 cherry picked

### LINKED ISSUE #330: Error handlings in form validation failures

jlhogan:
closes #330


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -101,6 +101,7 @@ const ChildInfo: Section = {
   key: 'child-information',
   name: 'Child information',
   status: () => 'complete',
+  Objects: (enrollment: Enrollment) => [enrollment.child],
```
jlhogan: hate the name of this field, please help

src/Hedwig/ClientApp/src/containers/Enrollment/enrollmentTypes.tsx
```diff
@@ -16,4 +17,5 @@ export interface Section {
 	status: (props: SectionProps) => StepStatus;
 	Summary: React.FC<SectionProps>;
 	Form: React.FC<SectionProps>;
+	ValidationObjects: (enrollment: DeepNonUndefineable<Enrollment>) => (Validatable | Validatable[] | null)[];
```
jlhogan: not sure that this needs to return an array of validation object, or if just one will suffice. I believe in the `enrollment/funding` section, we will need to check both the enrollment and the fundings so I _think_ it is in fact required.

src/Hedwig/ClientApp/src/utils/validations/sectionHasValidationErrors.ts
```diff
@@ -0,0 +1,16 @@
+import { Validatable, hasValidationErrors } from '.';
+
+export function sectionHasValidationErrors(sectionObjects: (Validatable | Validatable[] | null)[]) {
+  let hasErrors = false;
+  sectionObjects.forEach(o => {
+    if (!o) return;
+
+    if (Array.isArray(o)) {
+      hasErrors = o.some(hasValidationErrors) && !hasErrors;
```
jlhogan: need to think about how this will work for familydeterminations[]. Either update the logic here (don't check validationErrors on any but the first determination) OR update the logic in the backend (only create validation errors for the most recent familydetermination for a give family, consider others "out of date" and thus un-error-able

enrollment
```diff
@@ -0,0 +1 @@
+{"id":160,"childId":"16c07e75-fd55-4a94-a192-08d780eaf8e0","child":{"id":"16c07e75-fd55-4a94-a192-08d780eaf8e0","sasid":null,"firstName":"Harry","middleName":null,"lastName":"Melling","suffix":null,"birthdate":"2016-10-27T00:00:00","birthTown":"Hartford","birthState":"CT","birthCertificateId":"49929444","americanIndianOrAlaskaNative":false,"asian":false,"blackOrAfricanAmerican":true,"nativeHawaiianOrPacificIslander":false,"white":false,"hispanicOrLatinxEthnicity":true,"gender":"Male","foster":false,"familyId":160,"family":null,"enrollments":[],"organizationId":4,"organization":null,"validationErrors":null,"authorId":null,"author":null},"siteId":4,"site":null,"age":"Preschool","entry":"2019-09-02T00:00:00","exit":null,"fundings":null,"validationErrors":null,"authorId":null,"author":null}
```
cailyncodes: This probably wants to be deleted

src/Hedwig/ClientApp/src/utils/validations/Validatable.ts
```diff
@@ -0,0 +1,5 @@
+import { ValidationError } from '../../generated';
+
+export type Validatable = {
+	validationErrors?: ValidationError[] | null,
+}
```
cailyncodes: Does this not get autogenerated by Swashbuckle?

src/Hedwig/ClientApp/src/utils/validations/sectionHasValidationErrors.ts
cailyncodes: I tentatively lean to the latter case. I think the former would clutter up this function too much and become a weird edge case. The latter has a greater semantic clarity.

src/Hedwig/ClientApp/src/utils/validations/sectionHasValidationErrors.ts
```diff
@@ -0,0 +1,16 @@
+import { Validatable, hasValidationErrors } from '.';
+
+export function sectionHasValidationErrors(sectionObjects: (Validatable | Validatable[] | null)[]) {
+  let hasErrors = false;
+  sectionObjects.forEach(o => {
```
cailyncodes: Can we use `object` or `sectionObj` or similar?

enrollment
jlhogan: oops!

src/Hedwig/ClientApp/src/utils/validations/Validatable.ts
jlhogan: No, the `ValidationError` type does, but interface types are not in the schema, only concrete return types

jlhogan: No, the ValidationError type does, but interface types are not in the schema, only concrete return types. (I copied this from your diff :smile: )

jlhogan:
>     1. I think the newest comps have the missing info indicator on the specific field, so we'll need to think about how to do that with the `ValidationObjects` field on Section.
> 
>     2. When walking through the Enroll flow, a section always shows `Complete` even if there is missing information. It shows up correct at the overview/summary page at the end.

RE 1: was thinking this could be addressed in a separate PR; just wanted to get the util functions in here and then we can divide & conquer
RE 2: didn't update the `EnrollmentNew` component b/c it's not immediately obvious to me what's the best way to inject the complete/incomplete logic into the `StepList` and I was hoping @cmgiven could pick that up b/c again wanted to get this baseline in

cailyncodes:
>
> RE 1: was thinking this could be addressed in a separate PR; just wanted to get the util functions in here and then we can divide & conquer
RE 2: didn't update the EnrollmentNew component b/c it's not immediately obvious to me what's the best way to inject the complete/incomplete logic into the StepList and I was hoping @cmgiven could pick that up b/c again wanted to get this baseline in

Agree to 1 -- def a separate PR

Sounds good to 2

PR #455, opened 2019-12-17
Add error boundary, make new enrollment section load wait for user

mmazanec22:
Also changed whitespace from spaces to tabs when adding error boundary to app-- sorry for that


src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -96,7 +95,8 @@ export default function EnrollmentNew({
 	};
 
 
-	if (loading || error) {
+	if (loading || error || !user) {
+		// Need tadd user here so that a refresh after partial enrollment doesn't crash
```
jlhogan: :+1:

PR #469, opened 2019-12-18
463 non blocking validations

jlhogan:
Implements some non-blocking model validations

DOES NOT implement:
- rules about dates
- rules about fundings


src/Hedwig/Validations/Rules/Child/RaceRequired.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+          "AmericanIndianOrAlaskaNative",
+          "Asian",
+          "BlackOrAfricanAmerican",
+          "NativeHawaiianOrPacificIslander",
+          "White"
+        };
+        return new ValidationError(
+          fields: fields,
+          message: $"One of {String.Join(",", fields)} must be true."
```
cmgiven: We expect to show these messages to the user, yes? I'd suggest this should just be "Race is required."

src/Hedwig/Validations/Rules/Family/AddressLine1Required.cs
```diff
@@ -0,0 +1,10 @@
+using Hedwig.Models;
+
+namespace Hedwig.Validations.Rules
+{
+  public class AddressLine1Required : FieldRequired<Family>
+  {
+    public AddressLine1Required() : base("AddressLine1")
```
cmgiven: We're probably going to want a constructor that takes a sanitized version of the field name for the message.

jlhogan:
> So. Many. Tests.

still need to add integration tests to ensure the rules are actually _applied_, since that requires registering with the DI container in `ServiceExtensions.cs`

PR #472, opened 2019-12-19
Repository cleanup

jlhogan:
also:
- removes mocked REST backend
- some controller cleanup/standardization, and tests


jlhogan:
> Haven't closely examined all of the deletions, but LGTM unless you feel the need for additional scrutiny there.

nahhhhhhh, that's the point of strictly typed languages, no????

PR #474, opened 2019-12-19
Fixup missing information roster

jlhogan:


src/Hedwig/ClientApp/src/utils/types.ts
```diff
@@ -25,6 +25,11 @@ type DeepNonUndefineableObject<T extends object> =
 
 // Helper interface for handling arrays
 // Interface is required because defined types cannot use extend
+// class DeepNonUndefineableArray<T> extends Array<DeepNonUndefineable<T>> {
```
jlhogan: my attempt :(

PR #475, opened 2019-12-20
Add ability for selecting CDC funding / private pay during enrollment

### LINKED ISSUE #150: Add CDC funding section to enrollment

cailyncodes:
* Updates backend enrollment save to support updating funding --> because the mapping for fundings <--> enrollment lives on funding, we cannot just use _context.Update(enrollment) and instead need to inspect the fundings pointing to enrollment.Id. And then ef is very opininated on the order of everything and aggressive tracking.
* Creates backend reportingperiod controller. I feel like there should be a better way for all of this reporting period stuff given it lives so far away (within the db) to funding.
* Creates reportingperiod repo
* Adds in smarter typings for removing the type predicate =/= bool problem
* A fair amount of ui wrangling -- several utils updates to summary and form and some strong-ish assumptions. /shrug/

Should close #150

URL: ....

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useState, useEffect } from 'react';
 import { Section } from '../enrollmentTypes';
 import Button from '../../../components/Button/Button';
 import DatePicker from '../../../components/DatePicker/DatePicker';
@@ -7,12 +7,16 @@ import RadioGroup from '../../../components/RadioGroup/RadioGroup';
 import dateFormatter from '../../../utils/dateFormatter';
 import moment from 'moment';
 import idx from 'idx';
-import { ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPutRequest, Age, Enrollment } from '../../../generated';
+import { ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPutRequest, Age, FundingTime, Funding, FundingSource, ReportingPeriod } from '../../../generated';
 import UserContext from '../../../contexts/User/UserContext';
 import { ageFromString, prettyAge } from '../../../utils/ageGroupUtils';
 import getIdForUser from '../../../utils/getIdForUser';
-import Alert from '../../../components/Alert/Alert';
+import { DeepNonUndefineable } from '../../../utils/types';
 import { sectionHasValidationErrors } from '../../../utils/validations';
+import { prettyFundingTime, fundingTimeFromString } from '../../../utils/fundingTimeUtils';
+import { nextNReportingPeriods } from '../../../utils/models/reportingPeriod';
+import ReportingPeriodContext from '../../../contexts/ReportingPeriod/ReportingPeriodContext';
+import { currentFunding } from '../../../utils/models';
 
 const EnrollmentFunding: Section = {
   key: 'enrollment-funding',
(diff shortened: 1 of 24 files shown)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -7,110 +7,209 @@ import RadioGroup from '../../../components/RadioGroup/RadioGroup';
...
+						</p>
+						<p>Funding:{' '}
+							{isPrivatePay ?
+								'Private pay' :
+								`CDC - ${prettyFundingTime(cdcFundings[0].time)}`
+							}
+						</p>
+						{!isPrivatePay &&
+							<p>First reporting period:{' '}
+								{/* TODO: figure out how to get reporting period from funding (and whether it is possible without refetching all reports */}
```
cmgiven: A slightly hacky way: take the funding start date, add a week, take the month of the result, and show.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -7,110 +7,209 @@ import RadioGroup from '../../../components/RadioGroup/RadioGroup';
...
+		const [reportingPeriodOptions, updateReportingPeriodOptions] = useState<ReportingPeriod[]>([]);
+
+		useEffect(() => {
+			if (reportingPeriods) {
+				const currentCdcReportingPeriod = currentReportingPeriod(reportingPeriods);
+				if (!currentCdcReportingPeriod) {
+					throw new Error("Current CDC Reporting Period does not exist");
+				}
+
+				updateCdcReportingPeriod(currentCdcReportingPeriod);
```
cmgiven: The menu should start not with the current reporting period, but the earliest possible reporting period for which the child is eligible, the period that contains their start date. This period might be in the past.

cmgiven: Also, won't this overwrite the reporting period, which might already be set?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -7,110 +7,209 @@ import RadioGroup from '../../../components/RadioGroup/RadioGroup';
...
+					) as (_: DeepNonUndefineable<Funding>) => _ is DeepNonUndefineable<Funding>
+				) || null;
+				
+				updateCdcFundingId(cdcFunding ? cdcFunding.id : null);
+				updateCdcFundingTime(cdcFunding ? cdcFunding.time : null);
+				updatePrivatePay(cdcFunding === null);
+			}
+		}, [reportingPeriods, fundings]);
+
+		const familyDeterminationNotDisclosed = (enrollment: Enrollment) : boolean => {
```
cmgiven: Possible util candidate.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -7,110 +7,209 @@ import RadioGroup from '../../../components/RadioGroup/RadioGroup';
...
+				// TODO: Consider whether this should instead set the exit date on funding to today
+				updatedFundings = [
+					...(fundings.filter<DeepNonUndefineable<Funding>>(funding => funding.id !== cdcFundingId))
+				]
+			} else if (cdcFundingId && !privatePay) {
+				// Current funding exists, update it with supplied information
+				const newCdcFunding: Funding = {
+					...(currentCdcFunding as Funding), // currentCdcFunding will always be defined but typescript doesn't infer so
+					time: cdcFundingTime ? cdcFundingTime : undefined,
+					entry: entry ? entry : undefined
```
cmgiven: This should be the start of the selected first reporting period. It's not automatically the same as the enrollment entry date. (I know it's confusing they're both called `entry`; this was done to match CEDS, and I'm regretting it.)

Occurs once more below as well.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -7,110 +7,209 @@ import RadioGroup from '../../../components/RadioGroup/RadioGroup';
...
+					enrollmentId: enrollment.id,
+					entry: entry ? entry : undefined,
+					source: FundingSource.CDC,
+					time: cdcFundingTime ? cdcFundingTime : undefined
+				}
+				updatedFundings = [
+					...fundings,
+					newCdcFunding
+				]
+			}
```
cmgiven: I think (so could be wrong) that we can simplify this to have fewer conditionals.

`updatedFundings = [...(fundings.filter<DeepNonUndefineable<Funding>>(funding => funding.id !== cdcFundingId))]` is done in both the first and second cases, and is a no-op in the third and fourth. So we can do that every time.

I don't think it's necessary to find the current funding to splat for updates. Assuming it's not private pay, we can create a funding with the cdcFundingId and send it to the server (setting ID to 0 if there's no cdcFundingId), handling both create and update cases.

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
```diff
@@ -0,0 +1,40 @@
...
+	const currentYear = now.getFullYear();
+	const currentMonth = now.getMonth();
+	const startOfMonthDate = new Date(currentYear, currentMonth);
+	return periods.find(period => period.period.getTime() === startOfMonthDate.getTime());
+}
+
+export const fundingWithinReportingPeriod = (funding: Funding, period: ReportingPeriod): boolean => {
+	const entry = funding.entry;
+	const exit = funding.exit;
+	const start = period.period;
```
cmgiven: We're looking for `periodStart` here. `period.period` is the month that the period can be described using, e.g. "September 2019". It's possible both the start and end dates do not fall in that month.

src/Hedwig/ClientApp/src/utils/types.ts
```diff
@@ -23,9 +23,16 @@ type DeepNonUndefineableObject<T extends object> =
...
+// Class for handling arrays with overriden methods to handle typescript limitations
+export class DeepNonUndefineableArray<T> extends Array<DeepNonUndefineable<T>> {
+	constructor() {
+		super();
+	}
+
+	filter<S extends DeepNonUndefineable<T>>(callbackfn: (value: DeepNonUndefineable<T>, index: number, array: DeepNonUndefineable<T>[]) => boolean, thisArg?: any): DeepNonUndefineable<T[]> {
+		return super.filter<S>(((obj, i, arr) => callbackfn(obj, i, arr)) as (_: DeepNonUndefineable<T>, i: number, a: DeepNonUndefineable<T>[]) => _ is S);
+	}
+}
```
cmgiven: I'm not going to pretend to understand this sorcery. 😬

cailyncodes:
Thank you so much @cmgiven for the comments!

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
```diff
@@ -0,0 +1,40 @@
...
+	} else if (!exit) {
+		return (entry as Date) < end;
+	} else {
+		return start < exit;
+	}
+}
+
+export const nextNReportingPeriods = (periods: ReportingPeriod[], n: number): ReportingPeriod[] => {
+	const sortedPeriods = [...periods].sort((a, b) => a.period == b.period ?
+	0 :
```
jlhogan: nit but i think this is much easier to read as a function w/ a body over the nested ternary i.e.
```
(a, b) => {
  if (a.period == b.period) return 0;
  if (a.period < b.period) return -1;
  return 1;
}
```

src/Hedwig/ClientApp/src/utils/types.ts
jlhogan: loooooooooooooooooool :boom:

src/Hedwig/Data/DbInitializer.cs
```diff
@@ -34,10 +34,13 @@ public void Seed()
       CreateOrganizationPermission(organizationId: organization.Id, userId: user.Id);
 
       var reportingPeriods = new ReportingPeriod[] {
-        CreateReportingPeriod(period: "2019-08-01", start: "2019-07-29", end: "2019-09-01", due: "2019-09-15"),
-        CreateReportingPeriod(period: "2019-09-01", start: "2019-09-02", end: "2019-09-29", due: "2019-10-15"),
+        CreateReportingPeriod(period: "2019-08-01", start: "2019-07-29", end: "2019-08-30", due: "2019-09-15"),
```
jlhogan: is this true? or do the reporting periods actually wrap weirdly into other moths?

src/Hedwig/Repositories/EnrollmentRepository.cs
```diff
@@ -16,8 +16,35 @@ public class EnrollmentRepository : TemporalRepository, IEnrollmentRepository
 
 		public void UpdateEnrollment(Enrollment enrollment)
 		{
-      _context.Update(enrollment);
+
```
jlhogan: at a previous point i'd had some similar logic in here, with some stuff extracted into the base repository. Not sure that it maps directly to what you're doing here, but it could be a nice way to standardize some of this. Interesting read: RUDE, thanks for nothing EF core) that you have to do this checking  yourself, and can't just replace an array of fundings with another array of fundings........

anyway pyutting this here for your reference incase it's useful:  https://github.com/ctoec/ecis-experimental/commit/2da0d1dbba78fd34cac5f1023cbd2339428d2cae#diff-95af9b1213a14705330bf39a4b8a64bb

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
jlhogan: now is the time to rename .... :upside_down_face:

cailyncodes: With that, can we update the `Funding` model to actually store a `firstReportingPeriod` (type `ReportingPeriod`) object and a `lastReportingPeriod` object. And just ditch the `entry` `exit` fields?

src/Hedwig/Data/DbInitializer.cs
cailyncodes: Ah yep, I need to go back and update these

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: Ah makes sense -- see my comment below about adding reporting period to funding as an alternative that also handles this

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: Ooh yeah that should work. I think that can be a tech debt clean up to this so we can try to parallelize the remaining funding tickets.

src/Hedwig/ClientApp/src/utils/types.ts
cailyncodes: I'm going to sound like a broken record to @jlhogan at this point because I've said it so much, but I just love it so much. TypeScript's type system is turing complete (https://github.com/Microsoft/TypeScript/issues/14833) and that's just the coolest imo -- the fact type systems can encode computation!

src/Hedwig/Repositories/EnrollmentRepository.cs
cailyncodes: Ooh thanks! Yes could be useful -- I'm going to punt on that abstraction for now though

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cmgiven: If coming out of this PR we think that's going to be the most straightforward, yes, let's do it. Note that we'll need to add the ability to include ReportingPeriod to all the endpoints that can include a Funding, and it will break the utilization table on the Report, although the fix there should be simple.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cmgiven: Yup. That obviates any hackiness.

src/Hedwig/Repositories/EnrollmentRepository.cs
```diff
@@ -16,8 +16,35 @@ public class EnrollmentRepository : HedwigRepository, IEnrollmentRepository
...
+					if (dbFunding != null)
+					{
+						_context.Update(funding);
+					}
+					else
+					{
+						_context.Fundings.Add(funding);
+					}
+				}
+				_context.Update(enrollment);
```
cmgiven: This is the reason for one of the broken tests. Update() is only run if fundings is not null, which it is in the broken test. Move it down one curly brace and it's back to green.

src/Hedwig/Repositories/ReportRepository.cs
```diff
@@ -84,14 +84,14 @@ public async Task<CdcReport> GetReportForOrganizationAsync(int id, int orgId, st
         var fundings = await (asOf.HasValue ? _context.Fundings.AsOf(asOf.Value) : _context.Fundings)
           .AsNoTracking()
           .Where(funding => enrollmentIds.Contains(funding.EnrollmentId))
-          .Where(funding => funding.CertificateStartDate <= reportResult.ReportingPeriod.PeriodStart)
-          .Where(funding => funding.CertificateEndDate == null || funding.CertificateEndDate >= reportResult.ReportingPeriod.PeriodEnd)
+          .Where(funding => funding.FirstReportingPeriod.PeriodStart <= reportResult.ReportingPeriod.PeriodStart)
```
cmgiven: And this is the other issue. `FundingHelper.CreateFunding()` has not been updated to use reporting periods, so this is null in the tests, and the repository finds no reports.

cmgiven: And by "finds no reports" I mean "finds no fundings". 🙃

Issue #476, opened 2019-12-20
DevOps - Cleanup - Remove Unused Build Driver File - buildspec.yml


speedcm:
The buildspec.yml file was in use when the team was prototyping with the AWS CodeBuild tool.  This tool is not going to be implemented.  As a result, its build driver file can be removed from the repository.

speedcm:
This issue is complete.  This file has been removed from the repository.

PR #477, opened 2019-12-20
devops cleanup remove unused file (#476)

speedcm:
The buildspec.yml file was in use when the team was prototyping with the AWS CodeBuild tool.  This tool is not going to be implemented.  As a result, its build driver file can be removed from the repository.


speedcm:
This PR is ready for review.

PR #485, opened 2019-12-20
Update funding model

cailyncodes:


cailyncodes:
Specifically flagging: Can we have subtypes of funding if in enrollment we include a `ICollection<Funding>` --> will the fields added on the subtype show up on serialization? @jlhogan

jlhogan:
@cailyncodes so this will work fine from a dotnet perspective, but will not work when the json gets serialized into the typescript objects. We can define the funding model ourselves, and tell the generator _not_ to generate it. but not sure how hacky/un-hacky this will end up being.

src/Hedwig/Data/DbInitializer.cs
```diff
@@ -33,6 +34,14 @@ public void Seed()
 
       CreateOrganizationPermission(organizationId: organization.Id, userId: user.Id);
 
+      // secondary organization
+      var organization2 = CreateOrganization("Hogwarts");
```
cailyncodes: Omg yes!!

PR #487, opened 2019-12-20
Add reporting period include and update db init

cailyncodes:
Prereq and placeholder for ui update


jlhogan:
I did a hacky thing to solve this and merged it; you probably want half way between what I have and what you have

PR #489, opened 2019-12-21
153 blocking validations

jlhogan:


src/Hedwig/ClientApp/src/components/TextInput/TextInput.tsx
```diff
@@ -8,7 +9,8 @@ type TextInputProps = {
   defaultValue?: string;
   disabled?: boolean;
   success?: boolean;
-  error?: string;
+  errorMessage?: string;
+  errorType?: 'error' | 'warning';
```
cmgiven: Would be nice to make these props consistent across the components.

Rather than having two props that could potentially be in conflict with each other, I'd suggest we could use a single `error` prop, which is an object with an interface that includes a `message` and a `type`.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -7,15 +7,16 @@ import dateFormatter from '../../../utils/dateFormatter';
...
-import Checkbox from '../../../components/Checklist/Checkbox';
 import Alert from '../../../components/Alert/Alert';
 import parseCurrencyFromString from '../../../utils/parseCurrencyFromString';
 import currencyFormatter from '../../../utils/currencyFormatter';
 import getIdForUser from '../../../utils/getIdForUser';
 import UserContext from '../../../contexts/User/UserContext';
-import { sectionHasValidationErrors } from '../../../utils/validations';
+import { sectionHasValidationErrors, hasValidationErrors } from '../../../utils/validations';
+import FieldSet from '../../../components/FieldSet/FieldSet';
+import { determinationArgsAreValid, proceedWithoutCreatingDetermination } from '../../../utils/models';
```
cmgiven: `proceedWithoutCreatingDetermination` seems pretty specific to the FamilyIncome view. It might be better to have it as a function in this file, as a util is unlikely to be reused?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -123,41 +134,44 @@ const FamilyIncome: Section = {
...
+		  				<TextInput
+		  					id="numberOfPeople"
+		  					label="Household size"
+		  					defaultValue={numberOfPeople ? '' + numberOfPeople : ''}
+		  					onChange={event => {
+		  						const value = parseInt(event.target.value.replace(/[^0-9.]/g, ''), 10) || null;
+		  						updateNumberOfPeople(value);
+		  					}}
+								onBlur={event => (event.target.value = numberOfPeople ? '' + numberOfPeople : '')}
+								errorType={attemptedSave && !numberOfPeople && !notDisclosed ? 'error' : undefined}
```
cmgiven: I think having this logic here is a bit cumbersome, and might get particularly gnarly once we need to also show errors from the server.

I'd love to be able to call a function like `errorsForField` and have that deal with knowing about which errors to show based on whether the model has been saved, or a save has been attempted. And if that logic could be shared across multiple forms as a custom effect, that'd be great.

This is not a blocker. But I think a good candidate for refactoring.

jlhogan: I dont see how this logic is easily abstracted -- its different for each part of the form. what's your vision ?

src/Hedwig/ClientApp/src/assets/styles/_uswds-theme-custom-styles.scss
```diff
@@ -490,6 +490,9 @@ $future-step-color: color('base-dark') !default;
 		@include u-margin-top(6);
 		@include u-margin-bottom(4);
 	}
+	.inline-block {
```
mmazanec22: We don't need this-- USWDS has utilities: https://designsystem.digital.gov/utilities/display/

src/Hedwig/ClientApp/src/components/FieldSet/FieldSet.tsx
```diff
@@ -1,19 +1,29 @@
...
+    <fieldset className={`grid-gap grid-row usa-fieldset
+      ${error ? `usa-fieldset--${error.type}` : ''}
+      ${inlineBlock ? 'inline-block' : ''}
+    `} >
       <legend className="usa-sr-only">{legend}</legend>
-      {warning && (
-        <span className="usa-warning-message" id="field-set-warning-message" role="alert">
-          <InlineIcon icon='incomplete' /> {warning}
+      {error && (
+        <span className={`usa-${error.type}-message`} id="field-set-error-message" role="alert">
```
mmazanec22: We probably shouldn't use the alert role on warnings: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

src/Hedwig/ClientApp/src/utils/models/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './gender';
```
mmazanec22: :tada:

jlhogan:
@mmazanec22 ty for fixing up :)

PR #494, opened 2019-12-27
491 date picker warning error

mmazanec22:
Make datepicker warning/error/success states match text input, make reusable error/warning/success component for form items, accessibility-related fixes to associate inputs/labels/status messages via unique id.


src/Hedwig/ClientApp/src/assets/styles/_uswds-theme-custom-styles.scss
```diff
@@ -497,10 +504,9 @@ $future-step-color: color('base-dark') !default;
 	border-left-color: color($theme-color-warning-dark);
 }
 
-.usa-warning-message {
+.usa-warning-message, .usa-success-message {
```
cmgiven: I don't think this should be bold.

Screen Shot 2019-12-30 at 7 24 42 AM

PR #513, opened 2019-12-30
Adds validation to report enrollments

jlhogan:


src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -37,39 +38,41 @@ export default function UtilizationTable(report: CdcReport) {
...
+      }
+    })
+    : [];
 
-    return ageGroups.flatMap(ageGroup => fundingTimes.flatMap(fundingTime => {
+  const rows: UtilizationTableRow[] = ageGroups.flatMap(ageGroup => 
+    fundingTimes.flatMap(fundingTime => {
       const count = enrollments
-        .filter(enrollment => enrollment.ageGroup === ageGroup && enrollment.time === fundingTime)
+        .filter(enrollment => enrollment.ageGroup == ageGroup && enrollment.time == fundingTime)
```
cmgiven: Why this change?

src/Hedwig/Repositories/ReportRepository.cs
```diff
@@ -68,43 +68,17 @@ public async Task<CdcReport> GetReportForOrganizationAsync(int id, int orgId, st
       var reportResult = await reportQuery.FirstOrDefaultAsync();
 
       // Optionally manually insert time-versioned enrollment records
-      if (include.Contains(INCLUDE_ENROLLMENTS) || include.Contains(INCLUDE_CHILD))
+      if (include.Contains(INCLUDE_ENROLLMENTS))
```
cmgiven: This change means that `INCLUDE_CHILD` is a no-op unless `INCLUDE_ENROLLMENTS` is also included, yes? Is there a reason to prefer this?

src/Hedwig/Controllers/ReportsController.cs
```diff
@@ -16,10 +17,15 @@ namespace Hedwig.Controllers
   [Route("api/organizations/{orgId:int}/[controller]")]
   public class ReportsController : ControllerBase
   {
+    private readonly INonBlockingValidator _validator;
```
cmgiven: I'd expect we'd also want to validate on PUT to prevent submitting a report that has issues?

src/Hedwig/Repositories/ReportRepository.cs
jlhogan: that's the way we implemented all the enrollment includes, but it's not necessary. fine w/ either.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
jlhogan: yeah good q.... :)

PR #515, opened 2019-12-30
No dbinit in non-dev

jlhogan:
@speedcm -- do we plan to use "Development" as the environment for staging?

### DIFF
--- a/src/Hedwig/Program.cs
+++ b/src/Hedwig/Program.cs
@@ -13,21 +13,23 @@ namespace Hedwig
 
     public static void Main(string[] args)
     {
+      var environment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
       var host = CreateHostBuilder(args).Build();
-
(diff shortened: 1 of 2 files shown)


speedcm:
As per Chris, we have a total of 3 environments (for now).  Staging is non-development and is used for CI/CD and is updated automatically when the default branch (i.e. master) is updated. UAT is non-development and is a controlled environment and updated by the UAT pipeline manually triggered by a human.  Production is non-development and is a controlled environment and updated by the Production pipeline manually triggered by a human.  
If you would like additional pipelines/environments or want to change the branching strategy, just let me know.

cmgiven:
Do we maybe want to flip these to check if we're in prod? Running the db init stuff for non-prod?

speedcm:
Maybe only run db-init in non-prod?  We may have to introduce a new parameter to app_settings.json to specify either production or non-production OR specify the environment stage?.

Issue #470, opened 2019-12-18
Validate: end of first reporting period is after or the same as enrollment date


abbyraskin:
**AC:**
- Validate that the last day of the child's first CDC reporting period is the same or after the child's enrollment date (e.g., if the start date is January 1, the reporting period can't be earlier than Jan 1-Jan 31; a child CAN enroll earlier than the funding/reporting start date)
- Error highlights enrollment date field ("Start date") with following copy:  
  - Reporting period must be after start date

cailyncodes:
I think the error message should be: "Start date cannot be later than end of first reporting period" ?

abbyraskin:
@cailyncodes i just updated the AC after thinking about this more, i think it makes sense to validate on the reporting period rather than the start date (Reporting period must be after start date), sorry for the confusion.

PR #520, opened 2020-01-01
Add backend validation for entry start date

cailyncodes:
Backend for #470.

Frontend in separate PR.


jlhogan:
should this be a blocking validation? Also, should this be inverse? where the valid values for a funding start are restricted by enrollment entry? Not practically different, but more a question of how we imagine these will be created/updated, i think

cailyncodes:
How I interpreted AC for the issue was to error on enrollment entry date but a) could have misunderstood, or b) am open to instead flagging on funding period

PR #523, opened 2020-01-02
493 datepicker input format

mmazanec22:
@atreni I had added that calendar dropdown component during the rush to the demo, and @jlhogan pointed out that it's not that great for things like birthdate.  This is an implementation of [the USWDS recommendation](https://components.designsystem.digital.gov/components/detail/date-input.html) with some validation logic added in.  I swapped it out for birthdate in the enrollment flow.

What do you think of this validation behavior?  Are there any cases besides birthdate where the calendar dropdown is weird and needs replaced?

### DIFF
--- a/src/Hedwig/ClientApp/src/__snapshots__/storyshots.test.ts.snap
+++ b/src/Hedwig/ClientApp/src/__snapshots__/storyshots.test.ts.snap
@@ -247,7 +247,6 @@ exports[`Storyshots Checklist Box disabled 1`] = `
   <fieldset
     className="grid-gap grid-row usa-fieldset 
 			
-      
     "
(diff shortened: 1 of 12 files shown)


atreni:
@mmazanec22 I think this is the only place where this type of date picker is needed. The other use cases will be for more current dates, so the calendar drop down for those is a fine solution.

The validation looks good! The only thing that made me pause is that it's different from other validation types because it's immediate. But since it's required information for enrollment I think it's fine to catch them in real time.

One request - could we add Month (MM), Date (DD), and Year (YYYY) to the subtitles or show an example under DOB label (as shown in the [USDS style](https://components.designsystem.digital.gov/components/detail/date-input.html) you shared).

PR #540, opened 2020-01-03
Update start date validation error to new ac

cailyncodes:
Backend changes to fix broken enrollment with funding | Update validation error to new #470 AC

ToDo (in new PR): Clean up ui that is probably slightly broken


src/Hedwig/Validations/Rules/Enrollment/FundingIsValid.cs
```diff
@@ -0,0 +1,39 @@
...
+      INonBlockingValidator validator,
+      IFundingRepository fundings
+    ) : base(validator)
+    {
+      _fundings = fundings;
+    }
+
+    public ValidationError Execute(Enrollment enrollment)
+    {
+      if (enrollment != null && enrollment.Fundings.Count > 0)
```
jlhogan: enrollment.Fundings could be null, if you look in other sub object validations, i was doing something like
```
var fundings = enrollment.Fundings ?? _fundings.GetFundingsForEnrollment(enrollment)
``` which i think we need ? incase the resource was fetched w/out `include[]=fundings`? 

also, can the enrollment be null here? And if so, do we need to make sure to add that check to other subobject rules?

cailyncodes: Will make the change re `var fundings = ...`

I was running into an issue where `enrollment` was null and it was throwing null reference exceptions.

jlhogan: interesting! What situation led to null enrollment? (this convo can be had later, just interested to know haha)

cailyncodes: I don't fully remember the context, I just now I kept getting an exception on this line and was like I'll just add it to the conditional lol -- but yes, something to look into more

PR #545, opened 2020-01-03
542: A hacky fix for ethnicity selection

jlhogan:
Solution: check if value is null OR undefined
issue with that: it should always be null. weird.

additional rider: Add missing info styling to birth date

### DIFF
--- a/src/Hedwig/ClientApp/src/components/DatePicker/DatePicker.tsx
+++ b/src/Hedwig/ClientApp/src/components/DatePicker/DatePicker.tsx
@@ -86,7 +86,7 @@ export const DatePicker: React.FC<DatePickerProps> = ({
 			>
 				{labelText}
 			</label>
-			{status && <FormStatus {...status} />}
+			{status && status.message && <FormStatus {...status} />}
 			<span
 				className={`oec-date-input${status ? ` oec-date-input--${status.type}` : ''}`}
 				aria-describedby={status ? status.id : undefined}
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
@@ -252,12 +252,28 @@ const ChildInfo: Section = {
 				</div>
 
 				<h3>Date of birth</h3>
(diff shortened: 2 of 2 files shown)


src/Hedwig/ClientApp/src/components/DatePicker/DatePicker.tsx
```diff
@@ -86,7 +86,7 @@ export const DatePicker: React.FC<DatePickerProps> = ({
 			>
 				{labelText}
 			</label>
-			{status && <FormStatus {...status} />}
+			{status && status.message && <FormStatus {...status} />}
```
cmgiven: Is there a reason for this check? I see that `message` is optional on FormStatus.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -377,7 +393,7 @@ const ChildInfo: Section = {
 						},
 					]}
 					selected={
-						hispanicOrLatinxEthnicity === null ? '' : hispanicOrLatinxEthnicity ? 'yes' : 'no'
+						hispanicOrLatinxEthnicity === (null || undefined) ? '' : hispanicOrLatinxEthnicity ? 'yes' : 'no'
```
cailyncodes: This definitely works! Do we want to consider updating the `useApi` typing to avoid this? (It is a little involved after looking into it more -- probably post pilot but also maybe not at all?)

src/Hedwig/ClientApp/src/components/DatePicker/DatePicker.tsx
jlhogan: message is optional for when you want to display the warning-ed field inside of a fieldset, and have the warning text at the field-set level. If there is no message, you also don't want to add the little warning icon.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
jlhogan: worth adding to a list for post pilot but....  not my top priority :)

PR #550, opened 2020-01-06
Use api responses to display errors in forms

### LINKED ISSUE #514: refactor save logic and error message handling to allow showing blocking validation errors from server to user
we probably should just always try to post, even with bad params and let the server tell us why

jlhogan:
closes #514


src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -85,10 +85,19 @@ export default function useApi<TData>(
...
-			setState({ ...state, data: reducer(data, result) })
-			return result;
-		});
+		return query(api)
+			.then((result) => {
+				setState({ ...state, data: reducer(data, result) })
+				return result;
+			})
+			.catch(async (error) => {
+				if(error.status > 400) {
```
cailyncodes: Should this be `>=`?

src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -113,7 +122,14 @@ export default function useApi<TData>(
 					callback(result);
 				}
 			})
-			.catch((error) => setState({ ...state, loading: false, error: error.toString() }));
+			.catch(async (error) => {
+				if(error.status > 400) {
```
cailyncodes: Same here

Issue #535, opened 2020-01-03
Report should jump focus to first field with error if submit fails


abbyraskin:
**AC:**
If a report is submitted and fails, jump the focus to the first field with an error

mmazanec22:
According to Deque, we should move focus to the first form control with an error: https://www.deque.com/blog/anatomy-of-accessible-forms-error-messages/

abbyraskin:
Yes, sorry, writing quickly (I was thinking we had a universal alert for the submit failing, wasn't thinking about the fields being empty!)- updated! Thanks!

PR #561, opened 2020-01-06
Jump to first input field with a blocking error

### LINKED ISSUE #534: Jump focus when there is a first name, last name validation error
Currently, if a user saves the first section of enrollment with a blocking error, the save button becomes disabled, but cannot see the validation errors to know what happened. We should change the focus so users can see the in-line errors and resolve the issue.

**AC:**
- After saving the first section of enrollment, if there is a blocking error (missing first name and/or last name), move the focus to the first field with an error

cmgiven:
Closes #534 and #535.

Also shows an error instead of a blank page on a 500, but does not fix that reports are currently unable to be submitted due to an undiagnosed 500 error.

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportDetail.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportDetail.tsx
@@ -27,8 +27,8 @@ export default function ReportDetail() {
 		api => api.apiOrganizationsOrgIdReportsIdGet(reportParams),
 		[user]
 	);
-	
-	if (loading || error || !report) {
+
+	if (loading || !report) {
 		return <div className="Report"></div>;
 	}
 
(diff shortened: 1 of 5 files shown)


src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportDetail.tsx
```diff
@@ -27,25 +27,36 @@ export default function ReportDetail() {
 		api => api.apiOrganizationsOrgIdReportsIdGet(reportParams),
 		[user]
 	);
-	
-	if (loading || error || !report) {
+
+	if (loading || !report) {
```
jlhogan: I think in the long term, the solution is to only set the `error` from API on problems like no api object in client, and pass thru all server errors into the promise rejection as we're currently doing with the < 400s but for now this works!

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -115,6 +122,11 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
             onChange={(e) => setC4KRevenue(parseCurrencyFromString(e.target.value))}
             onBlur={event => (event.target.value = c4KRevenue !== null ? currencyFormatter(c4KRevenue) : '')}
             disabled={!!report.submittedAt}
+						status={serverErrorForField(
```
jlhogan: :clap:

Issue #544, opened 2020-01-03
Spacing and styling fixes


abbyraskin:
(Pasted from the [QA list](https://docs.google.com/document/d/1qLIojSyekdx2yKJZjchPtQfjnndT1LK-k0qWVTDkGdw/edit?folder=1m5jgRSeZRdvTj-Ny8c0_tZW5YirMvp2P) - please pair with a designer if additional context is needed!)

**Spacing**
- Fix spacing for roster filters when selected
- Section 2 - too much spacing
- Section 3 & 4 - not enough spacing
- Spacing between missing/warning info is too much, generally just make other sections consistent with section 1, make spaces between headers and content consistent
- Income determination not disclosed spacing is weird
- Width seems weird and narrow generally, but sometimes isn't-- is that on purpose?

**Styling**
- Yellow for missing information icon is not correct: #C2850C - gold 40v
- Dropdown boxes are missing a right border on Windows (Chrome)-- see if this is us, if it's a USWDS bug then leave it
- Field errors (see c4k on save with missing id) should have properly spaced title (or none at all? generic?)
- Color contrast on cancel button in enrollment withdrawal

![Screen Shot 2020-01-03 at 2 28 08 PM](https://user-images.githubusercontent.com/30734760/71747447-44c30f80-2e35-11ea-9900-78f5acd864f6.png)

mmazanec22:
Addressed by #568

PR #568, opened 2020-01-07
544 spacing styling fixes

mmazanec22:
Fixes QA list issues identified in #544 and generally corrects spacing for components that use fieldset.

There's still some screwiness with fieldset child input error spacing but I figure it's better to fix most of the things and make a new ticket for that.

### DIFF
--- a/src/Hedwig/ClientApp/src/assets/styles/_uswds-theme-custom-styles.scss
+++ b/src/Hedwig/ClientApp/src/assets/styles/_uswds-theme-custom-styles.scss
@@ -386,7 +386,7 @@ $future-step-color: color('base-dark') !default;
 	}
 }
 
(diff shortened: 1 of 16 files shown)


src/Hedwig/ClientApp/src/assets/styles/_uswds-theme-custom-styles.scss
```diff
@@ -386,7 +386,7 @@ $future-step-color: color('base-dark') !default;
 	}
 }
 
-.hedwig-enrollment-details-section {
+.oec-enrollment-details-section {
```
cmgiven: Arguably this should remain hedwig, as it relates to a part of this app, not a component that we hope to break out into an OEC component library.

PR #576, opened 2020-01-07
Withdrawal fast follows II

jlhogan:
achieves:
- updates copy for withdrawal error
- adds withdrawal success alert 
  - makes roster use common container
- hides first reporting period dropdown if enrollment is not funded


src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -115,7 +116,7 @@ export default function Roster() {
 	}
 
 	return (
-		<div className="Roster">
+		<CommonContainer>
 			<section className="grid-container">
```
mmazanec22: This is probably not a problem for this PR to solve but I had ditched the section element in the other containers because I think we're using it in a funny way-- probably anything that is a container the router is using should use a main element, and then we should section things appropriately

jlhogan: will do !

PR #588, opened 2020-01-08
PrivatePay/Select default fix

### LINKED ISSUE #582: Update funding model to allow for nullable time and source

cailyncodes:
Should close #582

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -40,10 +40,7 @@ const EnrollmentFunding: Section = {
 	key: 'enrollment-funding',
 	name: 'Enrollment and funding',
 	status: ({ enrollment }) =>
-		enrollment && (
-			sectionHasValidationErrors([enrollment, enrollment.fundings]) ||
-			!!(enrollment.fundings || []).find(funding => !funding.source)
-		)
+		enrollment && sectionHasValidationErrors([enrollment, enrollment.fundings])
 			? 'incomplete'
 			: 'complete',
 
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -124,7 +124,7 @@ const EnrollmentFunding: Section = {
 			throw new Error('EnrollmentFunding rendered without an enrollment');
 		}
 
-		const initialLoad = visitedSections ? !visitedSections[EnrollmentFunding.key] : true;
+		const initialLoad = visitedSections ? !visitedSections[EnrollmentFunding.key] : false;
 
 		const { user } = useContext(UserContext);
 		const { cdcReportingPeriods: reportingPeriods } = useContext(ReportingPeriodContext);
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -30,7 +30,7 @@ import {
 import { prettyFundingTime, fundingTimeFromString } from '../../../utils/fundingTimeUtils';
 import { nextNReportingPeriods } from '../../../utils/models/reportingPeriod';
 import ReportingPeriodContext from '../../../contexts/ReportingPeriod/ReportingPeriodContext';
-import { currentFunding } from '../../../utils/models';
+import { currentFunding, updateFunding, createFunding, currentC4kFunding } from '../../../utils/models';
 import { familyDeterminationNotDisclosed } from '../../../utils/models/familyDetermination';
 import Checklist from '../../../components/Checklist/Checklist';
 import TextInput from '../../../components/TextInput/TextInput';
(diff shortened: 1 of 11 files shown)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -131,11 +122,13 @@ const EnrollmentFunding: Section = {
 		);
 	},
 
-	Form: ({ enrollment, mutate, callback }) => {
+	Form: ({ enrollment, mutate, callback, visitedSections }) => {
 		if (!enrollment) {
 			throw new Error('EnrollmentFunding rendered without an enrollment');
 		}
 
+		const initialLoad = visitedSections ? !visitedSections[EnrollmentFunding.key] : true;
```
cailyncodes: This should default to `false` otherwise I think there might be issues with private pay on an edit of an enrollment

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -40,7 +40,10 @@ const EnrollmentFunding: Section = {
 	key: 'enrollment-funding',
 	name: 'Enrollment and funding',
 	status: ({ enrollment }) =>
-		enrollment && sectionHasValidationErrors([enrollment, enrollment.fundings])
+		enrollment && (
+			sectionHasValidationErrors([enrollment, enrollment.fundings]) ||
+			!!(enrollment.fundings || []).find(funding => !funding.source)
```
jlhogan: why wouldn't this be captured by validationErrors ?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
jlhogan: if visited sections is not a thing in enrollmentEdit then yes def

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -203,49 +197,94 @@ const EnrollmentFunding: Section = {
...
-					// The funding is to be updated, so add it back with the new values
-					updatedFundings.push({
-						...(currentCdcFunding as Funding), // currentCdcFunding will always be defined but typescript doesn't infer so
-						time: cdcFundingTime ? cdcFundingTime : undefined,
-						firstReportingPeriodId: cdcReportingPeriod ? cdcReportingPeriod.id : undefined,
-						firstReportingPeriod: cdcReportingPeriod,
-					});
-				} else {
-					// The funding is to be removed
+			// CDC REGION
```
jlhogan: lol

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: I don't think we current have a validation for that -- i.e. saving without a funding doesn't cause missing info

jlhogan: saving without a funding is valid state -- private pay

src/Hedwig/Models/Funding.cs
```diff
@@ -20,7 +20,7 @@ public class Funding : TemporalEntity, INonBlockingValidatableObject
     public Enrollment Enrollment { get; set; }
 
     [JsonConverter(typeof(StringEnumConverter))]
-    public FundingSource Source { get; set; }
+    public FundingSource? Source { get; set; }
```
jlhogan: won't you also need to remove `RequiredForFundingSource` on line 38, and then add missing info validations for these fields?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
jlhogan: if you mean funding w/out source or time does not cause missing info, that's true. But those validations should be added

cailyncodes: Yeah I meant source and time -- the state of selecting `Select` has not validation errors yet (which is the state of having a funding with no source)

src/Hedwig/Models/Funding.cs
cailyncodes: I don't think so -- line 38 (funding time) should not be affected by this (?)

jlhogan: oh duh b/c its only required for cdc fundings, so if source-less can also be time-less

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
jlhogan: do you want me to add them and push to this branch ?

cailyncodes: I was thinking we could create a new PR for that -- but I'm about to hop on a call so feel free to tackle that here if you have time

jlhogan: done

PR #592, opened 2020-01-08
Browser diff date debug

mmazanec22:
I have no way of testing this in safari and am not sure it will resolve the issue :/


cmgiven:
So here's the issue: I can't actually test in Safari, because I don't have a way to ignore the self-signed cert. The new version is just way locked down. I've be developing exclusively in Chrome. I'm stumped.

jlhogan:
imo, this can't hurt. We should use date OR moment regardless of if it's what fixes this error.

PR #596, opened 2020-01-09
Add InitialLoadingErrorGuard

### LINKED ISSUE #589: Hide missing info until already been visited

cailyncodes:
Should close #589 

This uses the `visitedSections` object that #582 introduced to only show errors after a using has hit save at least once in the section.


cailyncodes:
Just realized that refreshing the page is probably going to clear out the errors :/

PR #600, opened 2020-01-09
Add cancel enrollment functionality

### LINKED ISSUE #158: Cancel enrollment
**Story:** As a provider, I want to discard an enrollment midway through adding a child.

**Acceptance Criteria:**
- Adds a Cancel button to the enrollment page (bottom, left aligned)
- If a user selects "Cancel" while enrolling a child, display a browser system dialog with the following copy:
  - Header: Cancel enrollment?  
  - Body: If you cancel this enrollment you will lose any data you entered about this child.
  - Buttons: No, Yes
- If a user selects no, close the dialog
- If a user selects yes, delete the record and navigate the user to the roster page

**Design:**
image

cailyncodes:
Should close #158 

PR URL: https://18.220.0.248:5001

I added an alert after successfully cancelling, but we can remove that if not desired.

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
@@ -12,12 +12,14 @@ import UserContext from '../../../contexts/User/UserContext';
 import {
 	Enrollment,
 	ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGetRequest,
+	ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdDeleteRequest,
 } from '../../../generated';
 import getIdForUser from '../../../utils/getIdForUser';
 import CommonContainer from '../../CommonContainer';
 import { hasValidationErrors } from '../../../utils/validations';
 import AlertContext from '../../../contexts/Alert/AlertContext';
 import nameFormatter from '../../../utils/nameFormatter';
+import { useEffect } from '@storybook/addons';
 
 type EnrollmentNewParams = {
 	history: History;
@@ -85,6 +87,33 @@ export default function EnrollmentNew({
 		}
 	);
 
+	const [cancel, updateCancel] = useState(false);
+	const processSuccessfulCancel = () => {
+		setAlerts([
+			{
+				type: 'info',
+				heading: 'Cancelled enrollment',
+				text: 'Successfully cancelled enrollment',
+			},
+		]);
+		history.push('/roster');
+	}
+	const cancelParams: ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdDeleteRequest = {
+		id: enrollmentId ? enrollmentId : 0,
+		orgId: getIdForUser(user, 'org'),
+		siteId: getIdForUser(user, 'site'),
+		enrollment: enrollment
+	};
+	const [, cancelError] = useApi(
+		api => api.apiOrganizationsOrgIdSitesSiteIdEnrollmentsIdDelete(cancelParams),
+		[enrollmentId, enrollment, user, cancel],
+		{
+			skip: !cancel,
+			callback: processSuccessfulCancel
+		}
+	);
+	
+
 	const [visitedSections, updateVisitedSections] = useState(mapSectionToVisitedStates(sections));
 
 	const visitSection = (section: Section) => {
@@ -92,6 +121,10 @@ export default function EnrollmentNew({
 			...visitedSections,
 			[section.key]: true
 		});
+	}
+
+	if (cancelError) {
+		// TODO: do something with this error
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


cailyncodes:
Okay so these tests are not happy. I am unable to get around an EF error about `Funding` and `Enrollment` coming out of sync. I think we need to use `Cascade` for our delete approach (but I cannot make it so!) and I'm at my wits end. Any and all is so immensely appreciated!

test/HedwigTests/Repositories/EnrollmentRepositoryTests.cs
```diff
@@ -40,6 +40,51 @@ public void AddEnrollment()
...
+        var enrollmentRepo = new EnrollmentRepository(context);
+        var fundingRepo = new FundingRepository(context);
+        enrollmentRepo.AddEnrollment(enrollment);
+
+        Assert.Equal(EntityState.Added, context.Entry(enrollment).State);
+
+        enrollmentRepo.DeleteEnrollment(enrollment);
+
+        Assert.Equal(EntityState.Detached, context.Entry(funding).State);
+        Assert.Equal(EntityState.Detached, context.Entry(child).State);
```
jlhogan: will the child be deleted if it's not explicitly included in the enrollment to delete? I.e. are these actions separate-able? or will every enrollment delete also delete an associated child?

src/Hedwig/Migrations/20200109203931_ChangeFundingsToCascadeDelete.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Hedwig.Migrations
+{
+    public partial class ChangeFundingsToCascadeDelete : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
```
jlhogan: did you just manually add this migration? Might be better to change how https://github.com/ctoec/ecis-experimental/blob/master/src/Hedwig/Data/ModelBuilderExtensions.cs works to exclude this relationship there, that way the intended behavior of things is clear from the code

src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -112,5 +112,29 @@ Enrollment enrollment
             _validator.Validate(enrollment);
             return Ok(enrollment);
         }
+
+        [HttpDelete("{id}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
```
jlhogan: `[ProducesResponseType(StatusCodes.Status400BadRequest)]`

test/HedwigTests/Repositories/EnrollmentRepositoryTests.cs
cailyncodes: From a business logic standpoint, I guess I'm not sure. I was imagining this though as deleting an enrollment would also delete the child. They are separable with the Restrict functionality (or SetNull) but not Cascade. We are currently using Restrict so would require an additional call I think.

src/Hedwig/Migrations/20200109203931_ChangeFundingsToCascadeDelete.cs
cailyncodes: I ran `dotnet ef migrations add NAME` without making any code changes. It's the Microsoft doc recommended way to add new migrations that cannot be expressed through annotations.

jlhogan: sorry yeah, i meant was this code driven or not. But I think it would be better to update the model builder func that sets all fks to restrict to exclude certain FKs, or to set FKs only by type and then not set for enrollment->funding. Otherwise looking at the code would indicate that the schema should be one way, but migrations and actual DB state won't match that.

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -12,12 +12,14 @@ import UserContext from '../../../contexts/User/UserContext';
...
 	Enrollment,
 	ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGetRequest,
+	ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdDeleteRequest,
 } from '../../../generated';
 import getIdForUser from '../../../utils/getIdForUser';
 import CommonContainer from '../../CommonContainer';
 import { hasValidationErrors } from '../../../utils/validations';
 import AlertContext from '../../../contexts/Alert/AlertContext';
 import nameFormatter from '../../../utils/nameFormatter';
+import { useEffect } from '@storybook/addons';
```
mmazanec22: I wish we could override this specific auto import behavior in VS Code, gets me every time

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -94,6 +123,10 @@ export default function EnrollmentNew({
 		});
 	}
 
+	if (cancelError) {
+		// TODO: do something with this error
```
mmazanec22: I guess we would want to set an alert

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
cailyncodes: @jlhogan Very confused about your three different reacts hahaha

src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -112,5 +112,29 @@ Enrollment enrollment
             _validator.Validate(enrollment);
             return Ok(enrollment);
         }
+
+        [HttpDelete("{id}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
```
jlhogan: also 404 :)

src/Hedwig/Data/HedwigContext.cs
```diff
@@ -48,8 +48,16 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
       modelBuilder.Entity<Site>().ToTable("Site");
       modelBuilder.Entity<User>().ToTable("User");
 
-      // Set all fks onDelete to restrict to enable complex fk relationships
+      // Set all default fks onDelete to restrict to enable complex fk relationships
       modelBuilder.SetAllFKsOnDelete(DeleteBehavior.Restrict);
+
+      // Override specific foreign key relationships
```
jlhogan: works for me! but makes me worried about how we'll ever delete other things w/out cascades haha. A problem for another day

src/Hedwig/Controllers/EnrollmentsController.cs
cailyncodes: Oh lol -- your comment was an addition not a change oops

PR #607, opened 2020-01-10
Right align button

mmazanec22:


src/Hedwig/ClientApp/src/components/Button/Button.tsx
```diff
@@ -9,6 +9,7 @@ type ButtonProps = {
 	href?: string;
 	appearance?: ButtonAppearance;
 	disabled?: boolean;
+	className?: string,
```
cailyncodes: Do we want to make this array and join together multiple classNames?

src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.tsx
```diff
@@ -73,7 +73,7 @@ export default function EnrollmentDetail({
 				<div className="grid-row flex-first-baseline flex-space-between">
 					<h1>{nameFormatter(child)}</h1>
 					<div className="tablet:grid-col-auto">
-						<Button text="Withdraw" href={`/roster/enrollments/${enrollment.id}/withdraw`} />
+						<Button text="Withdraw" href={`/roster/enrollments/${enrollment.id}/withdraw`} className="margin-right-0" />
```
cailyncodes: Are we instead able to wrap this in a div and then have the CSS navigate down to the underlying html button?

src/Hedwig/ClientApp/src/components/Button/Button.tsx
mmazanec22: I have no strong feelings either way.  I think this is how we're doing this elsewhere rn-- we should maybe make a general type for form components to share via intersection and standardize how it works?

PR #609, opened 2020-01-10
Component exports

mmazanec22:
Sorry for another giant PR that changes a ton of files.  What do you all think of handling component imports this way?


src/Hedwig/ClientApp/src/components/Header/Header.tsx
```diff
@@ -30,7 +30,7 @@ const setActiveStateOfNavItem = function(item: NavItemProps, index: number, path
 	return { ...item, active };
 };
 
-class Header extends React.Component<HeaderProps & RouteComponentProps, HeaderState> {
+export class HeaderWithoutRouter extends React.Component<HeaderProps & RouteComponentProps, HeaderState> {
```
cailyncodes: I don't think we need to export this one.

PR #616, opened 2020-01-13
498: A client-side solution to current funding filtering

### LINKED ISSUE #498: Roster legends/tags should only use current fundings
Fundings are time-bound; they apply only for a certain period. Currently, the roster legends and tags look only for the presence of a funding on an active enrollment. They should be looking only for active fundings, that is those fundings whose first/last funding periods overlap with the date range being shown by the roster.

**AC: **
* When the roster is showing only current enrollments, a funding with a last funding period that ends prior to today should not be shown as receiving funding. A funding with a first funding period that starts in the future should not be shown as receiving funding.
* When the roster is showing past enrollments, any enrollment that has a funding period that overlaps with the roster's date range should be shown as receiving funding.
* When a funding has no last funding period, it should be assumed to extend indefinitely into the future.

(I don't believe this should be considered pilot-critical.)

jlhogan:
punting on a decision about should fundings be filtered on the server side until we know more about other funding types and the user stories around multiple fundings and thus past fundings etc

closes #498


src/Hedwig/ClientApp/src/utils/models/funding.ts
```diff
@@ -11,15 +12,59 @@ export function generateFundingTag(funding: DeepNonUndefineable<Funding>, index?
...
+			return isCurrentC4K(funding);
+
+		default:
+			return false;
+	}
+
+}
+
+export function isCurrentCDC(funding: Funding) : boolean {
+	const now = new Date(Date.now());
```
jlhogan: Should I be using moment here?

mmazanec22: Yeah-- I think different browsers behave differently re: time zones here, whereas moment will always use the browser's time zone.

PR #630, opened 2020-01-16
Abstract enumerable child object update

### LINKED ISSUE #483: Investigate abstracting enrollment/funding/subobject save/update
at a previous point i'd had some similar logic in here, with some stuff extracted into the base repository. Not sure that it maps directly to what you're doing here, but it could be a nice way to standardize some of this. Interesting (read: RUDE, thanks for nothing EF core) that you have to do this checking  yourself, and can't just replace an array of fundings with another array of fundings........

anyway pyutting this here for your reference incase it's useful:  https://github.com/ctoec/ecis-experimental/commit/2da0d1dbba78fd34cac5f1023cbd2339428d2cae#diff-95af9b1213a14705330bf39a4b8a64bb

_Originally posted by @jlhogan in https://github.com/ctoec/ecis-experimental/pull/475_

jlhogan:
closes #483

### DIFF
--- a/src/Hedwig/Models/HedwigIdEntity.cs
+++ b/src/Hedwig/Models/HedwigIdEntity.cs
@@ -1,4 +1,8 @@
-public interface IHedwigIdEntity<T>
+namespace Hedwig.Models
 {
-  T Id { get; }
+  public interface IHedwigIdEntity<out T>
+  {
+    T Id { get; }
+  }
 }
+
--- a/src/Hedwig/Models/HedwigIdEntity.cs
+++ b/src/Hedwig/Models/HedwigIdEntity.cs
@@ -1,0 +1,4 @@
+public interface IHedwigIdEntity<T>
+{
+  T Id { get; }
+}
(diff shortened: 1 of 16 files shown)


src/Hedwig/Models/HedwigIdEntity.cs
```diff
@@ -0,0 +1,4 @@
+public interface IHedwigIdEntity<T>
```
cailyncodes: There's a sonar cloud note that I actually found very interesting to read the "why is this an issue?" link. It was so cool that C# supports that language feature

test/HedwigTests/Repositories/EnrollmentRepositoryTests.cs
```diff
@@ -15,18 +15,77 @@ namespace HedwigTests.Repositories
...
+      {
+        var enrollmentRepo = new EnrollmentRepository(context);
+        enrollmentRepo.UpdateEnrollment(enrollment);
+        
+        Assert.Equal(EntityState.Added, context.Entry(enrollment.Fundings.Last()).State);
       }
     }
+
+
+
```
cailyncodes: Super nit: extra spaces

PR #631, opened 2020-01-16
Add additional ids for linking Site to other data sources

cmgiven:
These are IDs used in other systems including ECIS, Directory Manager, and Registry, as well as miscellaneous spreadsheets.

### DIFF
--- a/src/Hedwig/Migrations/20200116141931_AddIdsToSite.cs
+++ b/src/Hedwig/Migrations/20200116141931_AddIdsToSite.cs
@@ -1,0 +1,73 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Hedwig.Migrations
+{
+    public partial class AddIdsToSite : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropForeignKey(
+                name: "FK_Funding_Enrollment_EnrollmentId",
+                table: "Funding");
+
+            migrationBuilder.AddColumn<int>(
+                name: "FacilityCode",
+                table: "Site",
+                nullable: true);
+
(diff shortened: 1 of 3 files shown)


src/Hedwig/Migrations/20200116141931_AddIdsToSite.cs
```diff
@@ -0,0 +1,73 @@
...
+                table: "Site",
+                nullable: true);
+
+            migrationBuilder.AddForeignKey(
+                name: "FK_Funding_Enrollment_EnrollmentId",
+                table: "Funding",
+                column: "EnrollmentId",
+                principalTable: "Enrollment",
+                principalColumn: "Id",
+                onDelete: ReferentialAction.Cascade);
```
cmgiven: This appears to be correct, based on what's in `HedwigContext.cs`. I'm not sure why it's not reflected in the current snapshot, or what the specific motivation for cascade delete in this case is. Possible related to the cancel enrollment feature?

jlhogan: Yes, Cailyn added this in a previous change to enable deleting fundings and enrollments, but i guess it did not get captured in a migration and we don't have test coverage

cailyncodes: It's deleted to cancel enrollment, yep! It was implemented to delete the partial enrollment when cancel is selected. I ended up configuring it in `HedwigContext.cs` by overriding a set all FKs method, but yeah not sure why it is missing in the snapshot.

jlhogan: b/c you didn't also add a migration to capture the change. This does mean we don't have test coverage of this, otherwise it would have failed

PR #632, opened 2020-01-16
620 funding current to enrollment

### LINKED ISSUE #620: update funding filtering to "current-to-enrollment" as opposed to just "current-to-right-now"
I think this is required for viewing past enrollments correctly, i.e. if the enrollment was funded at the time of the enrollment, it should show that. right now it will not show fundings if they're in the past

jlhogan:
where should I be creating unit tests for util functions?

closes #620

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -151,7 +151,7 @@ const EnrollmentFunding: Section = {
 		);
 		const [privatePay, updatePrivatePay] = useState<boolean>(!initialLoad && !sourcelessFunding && cdcFundings.length === 0);
 
-		const [cdcReportingPeriod, updateCdcReportingPeriod] = useState<ReportingPeriod>();
+		const [cdcReportingPeriod, updateCdcReportingPeriod] = useState<ReportingPeriod|undefined>(cdcFunding ? cdcFunding.firstReportingPeriod : undefined);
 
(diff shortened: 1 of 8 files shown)


cailyncodes:
I'm thinking we should restructure all of our front end testing to separate out `src/` from a `test/` directory, a la what we do on the backend

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -151,7 +151,7 @@ const EnrollmentFunding: Section = {
 		);
 		const [privatePay, updatePrivatePay] = useState<boolean>(!initialLoad && !sourcelessFunding && cdcFundings.length === 0);
 
-		const [cdcReportingPeriod, updateCdcReportingPeriod] = useState<ReportingPeriod>();
+		const [cdcReportingPeriod, updateCdcReportingPeriod] = useState<ReportingPeriod|undefined>(cdcFunding ? cdcFunding.firstReportingPeriod : undefined);
```
cailyncodes: Do you need the `|undefined`?

jlhogan: for some reason, yes otherwise I'm getting 
```
Type 'undefined' is not assignable to type 'ReportingPeriod | (() => ReportingPeriod)'.ts(2345)
```

jlhogan: for some reason, yes otherwise I'm getting
```
Type 'undefined' is not assignable to type 'ReportingPeriod | (() => ReportingPeriod'
```

PR #645, opened 2020-01-17
640: Fixes regression and correctly displays historical/multiple fund…

### LINKED ISSUE #640: enrollments w multiple fundings in roster
the change to display fundings current to a range was incorrectly implemented, resulting in duplicate fundings showing up. 

**AC**
- fundings are shown that are current to the date range being displayed, or current if current view
- funding tags are deduplicated
- in date range view, funding count is just a total count, not a ratio

jlhogan:
…ings on enrollments

closes #640 

tests coming but i just can't right now on friday at 5, but i did at least manually test that it actually works this time

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
@@ -2,11 +2,11 @@ import React from 'react';
 import idx from 'idx';
 import pluralize from 'pluralize';
 import { Link } from 'react-router-dom';
-import { Table, TableProps, InlineIcon } from '../../components';
+import { Table, TableProps, InlineIcon, DateRange } from '../../components';
 import { Enrollment, Funding, FundingSpace, FundingSource } from '../../generated';
 import { lastFirstNameFormatter } from '../../utils/nameFormatter';
 import dateFormatter from '../../utils/dateFormatter';
-import { generateFundingTag, isCurrentToRange, NO_FUNDING } from '../../utils/models';
+import { generateFundingTag, isCurrentToRange, NO_FUNDING, dedupeFundings } from '../../utils/models';
 import { DeepNonUndefineable, DeepNonUndefineableArray } from '../../utils/types';
 import { hasValidationErrors } from '../../utils/validations';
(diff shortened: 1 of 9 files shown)


src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
```diff
@@ -18,111 +18,114 @@ type AgeGroupSectionProps = {
...
+				name: 'Funding',
+				cell: ({ row }) => (
+					<td>
+						{row.fundings && dedupeFundings(
+							row.fundings.filter(funding => isCurrentToRange(funding, rosterDateRange))
+							).length > 0 ?
+							(dedupeFundings(row.fundings.filter(funding => isCurrentToRange(funding, rosterDateRange))) as DeepNonUndefineable<Funding[]>)
+							.map<React.ReactNode>((funding: DeepNonUndefineable<Funding>, index: number) =>
+								generateFundingTag(funding, index)
+							) :
```
cailyncodes: Can we abstract this out into a separate method? It's somewhat difficult to understand what's going on here.

src/Hedwig/ClientApp/src/utils/models/funding.ts
```diff
@@ -1,8 +1,25 @@
...
+import moment from "moment";
 
+/**
+ * naive-ly deduplicate fundings based on source & time
+ * @param fundings
+ */
+export function dedupeFundings(fundings: Funding[]) {
+	const uniqueFundings: {[key: string]: Funding }  = {};
+
+	fundings.map(funding => {
```
cailyncodes: Do we need to sort the input any way because it looks like fundings get overwritten by the last value of the same key?

cailyncodes:
I guess the sonar cloud thing should be fixed, but I don't think it is as important in JS land (as compared to C)

src/Hedwig/ClientApp/src/utils/models/funding.ts
jlhogan: if this was being used for some other purpose maybe, but for now it's only used to display funding tags in the roster once they've already been filtered by time. This might be clearer when i impl your comment above and wrap together all this logic in a func

PR #647, opened 2020-01-21
Fix object assign order

mmazanec22:


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -370,7 +375,7 @@ const ChildInfo: Section = {
 					onChange={(_, selected) => {
 						updateChildRace(
 							childRace.map(raceObj =>
-								Object.assign({}, { selected: selected.includes(raceObj.value) }, raceObj)
+								Object.assign({}, raceObj, { selected: selected.includes(raceObj.value) })
```
cailyncodes: Let's do `raceObj => {...raceObj, { selected: selected.includes(raceObj.value) }}` to conform to ES2015.

Issue #634, opened 2020-01-16
front end tests for utils


jlhogan:
https://create-react-app.dev/docs/running-tests/#filename-conventions 

so, going to leave them in `src` where they live for now

PR #649, opened 2020-01-23
634: Starts adding more tests for utils

jlhogan:
starts on #634 more to come

### DIFF
--- a/src/Hedwig/ClientApp/src/utils/models/child.ts
+++ b/src/Hedwig/ClientApp/src/utils/models/child.ts
@@ -56,10 +56,3 @@ export function getSummaryLine(val: any) {
 
   return InlineIcon({ icon: 'incomplete' });
 }
-
-// TODO: validate birthdate
-export function childArgsAreValid(args: any) {
-  // required fields
-  if (!args.firstName || !args.lastName) return false;
-  return true;
-}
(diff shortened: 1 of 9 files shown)


src/Hedwig/ClientApp/src/utils/models/child.ts
```diff
@@ -55,11 +55,4 @@ export function getSummaryLine(val: any) {
   if(val) return val;
 
   return InlineIcon({ icon: 'incomplete' });
-}
-
-// TODO: validate birthdate
-export function childArgsAreValid(args: any) {
```
mmazanec22: Glad we're cleaning things out

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    ...baseReportingPeriod,
+    id: firstId - 1,
+    periodStart: new Date(beforeEndDate.add(-30, 'days').toLocaleString()),
+    periodEnd: new Date(beforeEndDate.toLocaleString())
+  };
+
+  const first = {
+    ...baseReportingPeriod,
+    id: firstId,
+    periodStart: new Date(moment(startDate).add(-30, 'days').toLocaleString()),
```
mmazanec22: Moment has some nice formatting utils that might make this simpler-- I think it would be something like moment(startDate).add(-3, 'days').format('whatever')

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    periodEnd: startDate
+  };
+
+  const afterEndDate = moment(startDate).add(3, 'months');
+  const after = {
+    ...baseReportingPeriod,
+    id: firstId + 1,
+    periodStart: new Date(afterEndDate.toLocaleString()),
```
mmazanec22: Similarly, I don't think we need the new Date thing here-- we could just use the afterEndDate.format('whatever format')

jlhogan: ahhhhh `.toDate()` ! ty

PR #650, opened 2020-01-27
322 report tests

mmazanec22:


src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -104,9 +104,11 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
 			<UtilizationTable {...{ ...report, accredited }} />
 			<form className="usa-form" onSubmit={onSubmit}>
 				<fieldset className="usa-fieldset">
+					{/* TODO: REPLACE WITH FIELDSET COMPONENT */}
 					<legend>
 						<h2 className="margin-bottom-0 margin-top-2">Other Revenue</h2>
 					</legend>
+					{/* TODO: is this actually required? error isn't happening.  We should mark it as optional if it isn't */}
```
jlhogan: requirement for `c4krevenue` does not exist in CdcReport backend model, nor for `familyFees`. Should they ? @cmgiven @abbyraskin

cmgiven: Some programs do not accept C4K. So I don't think it should be required (although we could force people to enter zeros). Long-term, we probably want some sort of validation where if you have C4K kids, you have to enter it (even if zero), or maybe where we can hide it. But because C4K payments can be so delayed and weird, I'm not sure how to write that logic.

cmgiven: `familyFees`, though, should probably be required.

jlhogan: @mmazanec22 let's leave this as is now, making them optional if that's what's needed to get this to be bug-free & ready to merge. I have opened a ticket to update the backend model to make familyfees required, and do necessary front end changes as well : https://github.com/ctoec/ecis-experimental/issues/674 can you add any details about what should be un-commented / changed RE optional status in there?

PR #653, opened 2020-01-28
Correct choicelist bug and make family state a dropdown

mmazanec22:


mmazanec22:
@cailyncodes the bug was a really awesome one where event.target.value for select was defaulting to the text if the value was undefined, so if you didn't select a state, it would be listed as "Hartford, -Select-" and not show an error :joy:

PR #660, opened 2020-01-28
482 fundings

### LINKED ISSUE #482: Simplify enrollment fundings add/update funding conditional
I think (so could be wrong) that we can simplify this to have fewer conditionals.

`updatedFundings = [...(fundings.filter<DeepNonUndefineable<Funding>>(funding => funding.id !== cdcFundingId))]` is done in both the first and second cases, and is a no-op in the third and fourth. So we can do that every time.

I don't think it's necessary to find the current funding to splat for updates. Assuming it's not private pay, we can create a funding with the cdcFundingId and send it to the server (setting ID to 0 if there's no cdcFundingId), handling both create and update cases.

_Originally posted by @cmgiven in https://github.com/ctoec/ecis-experimental/diffs/1_

cailyncodes:
This is a big one -- let me know if you have any questions.

Make sure to check `EnrollmentFunding.tsx ` -- it doesn't show by default b/c of too many lines edited

Test might fail -- forgot to check locally

Should close #482


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -186,139 +183,70 @@ const EnrollmentFunding: Section = {
 			}
 			const periods = cdcReportingPeriod ? [cdcReportingPeriod, ...nextPeriodsExcludingCurrent] : nextPeriodsExcludingCurrent;
 			updateReportingPeriodOptions([...periods].sort(periodSorter));
-		}, [enrollment.entry, entry, reportingPeriods, cdcReportingPeriod]);
+		}, [enrollment.entry, entry, reportingPeriods, /*cdcReportingPeriod*/]);
```
cailyncodes: Oops should fix this

src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
```diff
@@ -132,19 +133,57 @@ describe('EnrollmentEdit', () => {
...
+			expect(initialFundingType.value).toBe('Full');
+
+			const saveBtn = wrapper.find('.EnrollmentFundingForm button');
+			await act(async () => {
+				saveBtn
+					.props()
+					.onClick();
+			});
+
+			console.log(wrapper.find('.oec-enrollment-details-section a'));
```
jlhogan: console log

src/Hedwig/ClientApp/src/utils/models/funding.ts
```diff
@@ -203,4 +205,18 @@ export function updateFunding({
 	}
 }
 
+export function hasFundings(enrollment: DeepNonUndefineable<Enrollment>): DeepNonUndefineable<Funding[]> | undefined {
```
jlhogan: functions named with `has` indicate boolean checks to me. Since these return the value, should they be prefixed with something like `get`?

src/Hedwig/ClientApp/src/utils/forms/form.ts
```diff
@@ -0,0 +1,79 @@
+import produce from "immer";
```
jlhogan: can we get some comments on this ?

PR #670, opened 2020-01-29
Update forms with submit and no autocomplete

### LINKED ISSUE #668: Disable browser auto-fill for all form fields
Chrome and other browsers will suggest previously used addresses and other information when enrolling or editing.

cmgiven:
Closes #668.

Also updates all the forms to allow for saving by hitting the Enter key. `noValidate` is needed in order to prevent the browser default behavior of preventing submit if a required field is blank.


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -214,11 +214,13 @@ const ChildInfo: Section = {
...
 				throw new Error('Something impossible happened');
 			}
+
+			event.preventDefault();
 		};
 
 		// TODO: should gender be radio buttons as recommended by USWDS rather than select?
 		return (
-			<div className="ChildInfoForm usa-form">
+			<form className="ChildInfoForm usa-form" onSubmit={save} noValidate autoComplete="off">
```
jlhogan: worth creating a custom form component that defaults to have `noValidate autoComplete="off"` ? not a blocker, just a thought

cmgiven: Maybe. On the other hand, if this is truly a generic OEC component library (the jury's still out), we probably wouldn't want it to have either of those props set that way by default. Our use of required not-required is pretty specific to this tool. And although maybe more OEC products would want to disable auto-complete than would not, I think it's better to err toward explicitness when disabling browser behavior. Even `.usa-form` we're not using consistently as best suits the whims of the CSS gods.

Issue #652, opened 2020-01-28
Multi-site roster view


jlhogan:
User story: As a provider from a multi-site organization, I want to be able to optionally view a multi-site roster, or view a site-specific roster

**AC**:
- can view all enrollments for all sites (default)
- can view enrollments for a single site by selecting the count of sites link (see design), which displays a dropdown of sites belonging to the program organization
- the single-site view includes a "Back to program roster" link (note: program should not be capitalized) to return to the combined multi-site roster 
  - back button in browser also returns user to combined roster view

**Design:**

## Multi-site roster view (combined)
<img width="1067" alt="Screen Shot 2020-02-18 at 4 50 34 PM" src="https://user-images.githubusercontent.com/30734760/74781084-dc0de600-526e-11ea-8d4b-65ac3e0ef72c.png">

## Dropdown with list of sites
<img width="910" alt="Screen Shot 2020-02-18 at 4 55 11 PM" src="https://user-images.githubusercontent.com/30734760/74781416-70784880-526f-11ea-9d34-5de37e90fcd6.png">

## "Back to roster" roster view after selecting individual site
<img width="576" alt="Screen Shot 2020-02-20 at 6 04 41 PM" src="https://user-images.githubusercontent.com/30734760/74988296-7b67e000-540b-11ea-98ba-33e2609ddbcf.png">

abbyraskin:
@cailyncodes small AC change incoming (affects the second step) - will update after storytime momentarily.

cailyncodes:
Sounds good! There are some other things I need to figure out first before I really dig into implementation

abbyraskin:
Up to date!

PR #678, opened 2020-01-30
Initial steps for multi-site (Issue #654; #655)

### LINKED ISSUE #654: enrollment flow changes for multi-site
in multi-site world, siteId needs to be parameter passed into each enrollment section, instead of pulled from user context (since each user could have mulitple sites in their permissions object)

**AC**: 
- siteId is a path param in EnrollmentEdit and EnrollmentDetail 
- sections take siteId as a prop

### LINKED ISSUE #655: user permission util changes for multi-site
in multi-site world, siteId cannot be inferred from user permissions object as users will have permissions for multiple sites, either via a specific site permission, or an org permission for the site-owning org. `getIdForUser` must be updated to take in a specific site id, and check for access to that site. 

**AC**:
- `getIdForUser` takes specific siteId as parameter
- `getIdForUser` returns the id if the user has permissions, or some other false-y value (maybe 0?) if not

open question: If the user does not have access to the site, should the client error, or should the client make the request and have the server error? I am leaning towards server error for consistency

cailyncodes:
Add siteId to sections and containers
Update paths to include siteId
Update getIdForUser to only use org
Add validatePermissions to check for access to site id

Should close #654, close #655


src/Hedwig/ClientApp/src/routes.ts
```diff
@@ -41,20 +41,20 @@ export const routes: RouteConfig[] = [
 		component: EnrollmentNew,
 	},
 	{
-		path: '/roster/enrollments/:enrollmentId',
+		path: '/roster/sites/:siteId/enrollments/:enrollmentId',
```
mmazanec22: If a user belongs to an organization and has permission to view more than one site, would /roster eventually show all of those sites for the org?  If yes then our current routing scheme makes sense-- but if a roster is something that exists on a site per site basis then maybe /sites/:id is the view of the roster for that site and /roster doesn't need to exist?

cailyncodes: I thinks that will be addressed here: https://github.com/ctoec/ecis-experimental/issues/652

PR #679, opened 2020-01-30
661: Adds UpdatedAt to temporal entities

### LINKED ISSUE #661: Expose updated on timestamp
temporal entities have `AuthorId` field and two hidden timestamps, `SysStartTime` and `SysEndTime`. `SysEndTime` ends up being the updated at value. However I wonder if it might be better to leave those hidden fields as they are to be used for temporal versioning, and add an explicit `UpdatedAt` timestamp field to all temporal entities. 

**AC**
- Add `UpdatedAt` timestamp field to `TemporalEntity` with default value and on update value `getutcdate()`
- create migration to upadte all temporal tables to have this field

jlhogan:
closes #661

### DIFF
--- a/src/Hedwig/Data/HedwigContext.cs
+++ b/src/Hedwig/Data/HedwigContext.cs
@@ -73,6 +73,7 @@ namespace Hedwig.Data
       {
         ProtectReadOnlyEntities();
       }
+      UpdateTemporalEntities();
       return base.SaveChangesAsync();
     }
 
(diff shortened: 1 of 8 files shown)


src/Hedwig/Data/HedwigContext.cs
```diff
@@ -111,12 +113,14 @@ private bool IsTemporalEntityType<T>()
     }
 
     /// <summary>
-    /// Recursively adds author UserId to AuthorId field on temporal entities
+    /// Recursively adds author UserId to AuthorId field on temporal entities,
+    /// and set UpdatedAt timestamp
     /// </summary>
     /// <param name="entity"></param>s
-    protected void AddAuthorToTemporalEntity(TemporalEntity entity, int? currentUserId)
+    protected void AddAuthorToTemporalEntity(TemporalEntity entity, int? currentUserId, DateTime now)
```
cmgiven: Why do we do this recursively?

jlhogan: to make sure the info trickles down to the entire object tree -- at this point the backend can't determine if the state of the individual sub-objects under the enrollment have changed, and they're all considered updated on every post. This is unavoidable due to the fact that we don't do partial PUTs. We're re-inforcing this here by blindly adding updatedAt/AuthorId to every temporal entity in the tree, and it might semantically make more sense, and end up being more correct in other instances, to do this by adding the info to all Added/Updated entities before save instead of recursing into the object tree to supplement with this info. But the end result is effectively the same.

jlhogan:
@cmgiven gave this a refresh with new approach inspired by the work from #700 if you're interested in re-reviewing since a lot has changed

PR #689, opened 2020-01-31
674: Makes familyFeesrequired on CdcReport

### LINKED ISSUE #658: Accredited checkbox should be disabled on submitted report

### LINKED ISSUE #659: Report submit form should use button and fieldset components
Right now it's using plain ol HTML elements

### LINKED ISSUE #674: require FamilyFees on CdcReport
**AC**
- FamilyFees field is required on backend model
- UI correctly displays server errors if post fails due to missing familyfees

jlhogan:
also:
- uses fieldset in reportSubmitForm
- fixes server error for familyfees field
- adds disabled to checkbox directly in uber-choicelist component

closes #674, closes #659 , closes #658 

seeking copy feedback on: ( cc @atreni  / @abbyraskin)
Screenshot from 2020-01-31 13-36-28


src/Hedwig/ClientApp/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -123,6 +123,7 @@ export function ChoiceList({
 					name={name || ''}
 					onChange={changeEvent}
 					selected={selectedItems.includes(option.value)}
+					disabled={disabled}
```
jlhogan: @mmazanec22 do you think this is the correct way to accomplish this?

mmazanec22: We could disable at the fieldset level I guess-- but I'm not sure whether it makes a difference in usability.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -155,13 +154,12 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
 						}
 						disabled={!!report.submittedAt}
 						status={serverErrorForField(
-							'report.familyfeesrevenue',
+							'familyfeesrevenue',
 							apiError,
-							'This information is required for the report'
+							'This information is required to submit the report'
```
cmgiven: I'd suggest just "This information is required". It happens when you try to submit, so I don't think we need to specify like we do with missing information, and the shorter length avoids the awkward line break.

src/Hedwig/Models/Reports/CdcReport.cs
```diff
@@ -13,6 +13,7 @@ public class CdcReport : OrganizationReport
 		public bool RetroactiveC4KRevenue { get; set; } = false;
 
 		[Column(TypeName = "decimal(18,2)")]
-		public decimal FamilyFeesRevenue { get; set; }
+		[Required]
+		public decimal? FamilyFeesRevenue { get; set; }
```
cmgiven: Just checking: does this require a migration or regenerating the API specs?

jlhogan: since field went from not required but not explicitly nullable to required but nullable, nothing in the schema/api defn needs to change... the defn in the api _will_ change from `familyFees?: number` to `familyFees: number | null` but these are effectively the same wrapped in `DeepNonUndefineable<>`.

I'll check the changes in here b/c might as well keep everything correct and up to date!

PR #691, opened 2020-02-03
Fix for enrollment/funding section not saving

cmgiven:


jlhogan:
this is a partial fix -- there's backend issues as well. Wasn't clear to me if this was in the larger PR @mmazanec22 is working on. But it would make sense to get the fix in separate from the other work in that ticket. Can we add it here? cc @cailyncodes

src/Hedwig/Repositories/HedwigRepository.cs
```diff
@@ -28,44 +28,27 @@ public Task SaveChangesAsync()
...
-				}
-
-				if(update.Id is int iid && iid == 0)
-				{
-					_context.Add(update);
-					break;
-				}
-
-				_context.Update(update);
-			}
```
cailyncodes: It works removing this?

jlhogan:
merged via https://github.com/ctoec/ecis-experimental/pull/692

PR #696, opened 2020-02-03
429 alert overutilization

### LINKED ISSUE #429: Overutilization alert during enrollment
**AC:**
- If a provider has filled all slots for a given age group in the current reporting period, display an alert letting the user know they will have used more spaces than are available.
- The alert logic is based on spaces allotted per age group (e.g., if the site has 12 preschool spaces, this alert shows if there are currently 12 children enrolled in a preschool space while a user tries to enroll a new child in a preschool space)
- Copy (note differs from design): 
  - This space type is already fully utilized for the [report period, e.g., January 2019] reporting period.

**Design:**
Screen Shot 2019-12-17 at 2 18 21 PM

mmazanec22:
Closes #429

### DIFF
--- a/src/Hedwig/ClientApp/src/utils/fundingSelectionUtils.ts
+++ b/src/Hedwig/ClientApp/src/utils/fundingSelectionUtils.ts
@@ -24,12 +24,10 @@ export function fundingSelectionToString(fundingSelection: FundingSelection): st
 }
 
 export function fundingSelectionFromString(value: string): FundingSelection {
+	// prettier-ignore
 	const source =
-		value === 'privatePay'
-			? FundingType.PRIVATE_PAY
-			: value === ''
-			? FundingType.UNSELECTED
-			: FundingType.CDC;
+		(value === 'privatePay' ?
+			FundingType.PRIVATE_PAY : (value === '' ? FundingType.UNSELECTED : FundingType.CDC));
 	if (source !== FundingType.CDC) {
 		return { source };
 	}
--- a/src/Hedwig/ClientApp/src/utils/fundingSelectionUtils.ts
+++ b/src/Hedwig/ClientApp/src/utils/fundingSelectionUtils.ts
@@ -1,32 +1,38 @@
 import { FundingTime } from '../generated/';
 import { fundingTimeFromString } from './models/fundingTime';
 
-export enum FundingSelection {
-	CDC_FULL,
-	CDC_PART,
+export enum FundingType {
+	CDC,
 	PRIVATE_PAY,
-	UNSELECTED
+	UNSELECTED,
 }
 
-export function fundingSelectionToString(fundingSelection: FundingSelection) {
-	switch (fundingSelection) {
-		case FundingSelection.UNSELECTED:
+export type FundingSelection = {
+	source: FundingType;
+	time?: FundingTime;
+};
+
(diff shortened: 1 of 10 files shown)


src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
```diff
@@ -20,7 +20,8 @@ export const currentReportingPeriod = (periods: ReportingPeriod[]): ReportingPer
 	const now = moment();
 	const startOfMonthDate = moment(`${now.format('YYYY-MM')}-01`);
 	// Assumes that ReportingPeriod.Period will always be on the first day of the month
-	return periods.find(period => period.period.getTime() === startOfMonthDate.valueOf());
+	// Compare with dates, not times to avoid timezone ridiculousness
+	return periods.find(period => moment(period.period).format('YYYY-MM-DD') === startOfMonthDate.format('YYYY-MM-DD'));
```
jlhogan: can leverage `moment.date` as opposed to using format

src/Hedwig/Repositories/HedwigRepository.cs
```diff
@@ -34,9 +34,8 @@ public Task SaveChangesAsync()
 		/// <param name="updates"></param>
 		/// <param name="currents"></param>
 		/// <typeparam name="T"></typeparam>
-		public void UpdateEnumerableChildObjects<T>(IEnumerable<IHedwigIdEntity<T>> updates, IEnumerable<IHedwigIdEntity<T>> currents)
+		public void UpdateEnumerableChildObjects<T>(IEnumerable<IHedwigIdEntity<T>> updates, IEnumerable<IHedwigIdEntity<T>> currents) where T : IEquatable<T>
```
jlhogan: line 46 changed to
```
if(!updates.Any(u => u.Id.Equals(current.Id)))
```
which may mean the `where T : IEquatable<T>` is unnecessary

src/Hedwig/ClientApp/src/utils/fundingSelectionUtils.ts
```diff
@@ -1,32 +1,38 @@
...
-		FundingSelection.PRIVATE_PAY :
-		value === '' ?
-			FundingSelection.UNSELECTED :
-			fundingTimeFromString(value) === FundingTime.Full ?
-				FundingSelection.CDC_FULL :
-				FundingSelection.CDC_PART;
-}
\ No newline at end of file
+	const source =
+		value === 'privatePay'
```
jlhogan: I find the staggered indented on nested ternarys extremely helpful in readability. Was this an auto change or manual? if auto, can we update a linting rule to leave these be? Not a blocker for this PR, just my opinion

cailyncodes: I second the value of nested formatting here!

mmazanec22: I agree that it's more readable; this was an auto update.  I will check and see if there's a linter rule that will preserve it!

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
mmazanec22: I think that particular method will not be what we want here: https://momentjs.com/docs/#/get-set/date/

PR #699, opened 2020-02-03
Backend and frontend changes for initial multi-site support

### LINKED ISSUE #688: Update frontend to make multiple site queries for enrollments
After playing around with implementing [organization-level enrollments], I think it may be better to just make the change in the frontend to make multiple simultaneous queries for per-site enrollments... not the most optimized but keeps the access/permissioning and the dotnet niceness with routes simple

_Originally posted by @jlhogan in https://github.com/ctoec/ecis-experimental/issues/651#issuecomment-580406769_

cailyncodes:
Should close #688 and (old issue I closed by actually is needed and will update this soon)

Boat load of changes:
* Update permission handling to follow Microsoft pattern (reqs empty classes, multiple handlers, policy as wrapper to req).
* Add endpoint to enrollment controller for multiple sites
* Update frontend to make roster request for all sites

Outstanding: Tests for MultipleSiteHandler (question in code), getting enrollments by org permissioning, getting enrollments when only org (no sites)


cailyncodes:
Welp I have no idea why those front end tests are failing 😭

src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -9,132 +9,164 @@
 using Hedwig.Validations;
 using Hedwig.Security;
 using System;
+using System.Linq;
 
 namespace Hedwig.Controllers
 {
-    [ApiController]
```
jlhogan: :( boo at this diff

src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -9,132 +9,164 @@
...
+			Enrollment enrollment
+		)
+		{
+			// Creating enrollment with Id not allowed
+			if (enrollment.Id != 0) return BadRequest();
+
+			_enrollments.AddEnrollment(enrollment);
+			await _enrollments.SaveChangesAsync();
+
+			return CreatedAtAction(
```
jlhogan: should this enrollment also be validated before being returned?

src/Hedwig/Security/SiteAccess/MultipleSiteAccessHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+				if (userIdStr != null && siteIdStr != null && siteIdStr != "")
+				{
+					var userId = Guid.Parse(userIdStr);
+					var siteId = Int32.Parse(siteIdStr);
+					return _permissions.UserCanAccessSite(userId, siteId);
+				}
+				return false;
+			});
+
+			// TODO: Allow for only organization level access
```
jlhogan: interested in what this TODO means

test/HedwigTests/Security/OrganizationAccess/OrganizationAccessHandlerTests.cs
```diff
@@ -106,11 +126,16 @@ public void User_Does_Not_Have_Organization_Access_Evaluate_Returns_False()
...
+                var req = new OrganizationAccessRequirement();
+
+                var authContext = new AuthorizationHandlerContext(new List<IAuthorizationRequirement> { req }, userClaim, new object());
+
+                // When requirement handler handles authorization context
+                reqHandler.HandleAsync(authContext);
 
-                // Then it should evaluate to False
-                Assert.False(res);
+                // Then authorization context will have succeeded
```
jlhogan: update this comment

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -35,26 +35,27 @@ export default function Roster() {
 	};
 
 	const { user } = useContext(UserContext);
-	const siteId = idx(user, _ => _.orgPermissions[0].organization.sites[0].id) || 0;
+	const siteIds = (idx(user, _ => _.orgPermissions[0].organization.sites) || []).map(site => site.id);
```
jlhogan: this should probably get wrapped in a util b/c we probably want to do an intersection of org permissions and site permissions

cailyncodes: This is a placeholder to allow the site to still work while pending ac on roster site select. Not sure there is value in making a util just yet for this

src/Hedwig/Controllers/EnrollmentsController.cs
cailyncodes: I think it's the tab <--> space issue

src/Hedwig/Controllers/EnrollmentsController.cs
cailyncodes: Ooh probably -- I don't think we were doing that before but yeah we should add that

src/Hedwig/Security/SiteAccess/MultipleSiteAccessHandler.cs
cailyncodes: The code as written would require a user with full org permissions to also have all the child site permissions. So this needs to be updated to return all enrollments in all sites in a given org

test/HedwigTests/Security/OrganizationAccess/OrganizationAccessHandlerTests.cs
jlhogan: should just say context will NOT have succeeded :)

cailyncodes: Oh yeah -- that eyes emoji was a reminder to me to address this haha

src/Hedwig/Security/SiteAccess/MultipleSiteAccessHandler.cs
jlhogan: `UserCanAccessSite` checks for site or owning org access, so this shouldn't require user to have individual site permissions.

> So this needs to be updated to return all enrollments in all sites in a given org
does this mean a different behavior for if you make the request w/out explicit siteIds? Seems right now siteIds are required

cailyncodes:
I think I know why tests are failing! It might be how we are mocking the `useApi` hook in the tests file

src/Hedwig/Security/SiteAccess/MultipleSiteAccessHandler.cs
cailyncodes: Are we able to always get all the sites from an org in the front end? If so, then the front end just needs to be updated. My first thought was to make siteIds optional and then only check org permissions

jlhogan: user objects always include all org permissions, and fully hydrated org object including sites, and site permissions
https://github.com/ctoec/ecis-experimental/blob/39d32a2280b888061cf71bca0d063a30920b9a8e/src/Hedwig/Repositories/UserRepository.cs#L25

so, seems they will always be available in the front end! not sure if there's a separate reason to need/want a site-id-less version of this endpoint... thinking on it...

cailyncodes:
> results from the handlers are ORed where site access policy is applied?

Yes :)

PR #702, opened 2020-02-03
On route change in new enrollment, scroll to top

### LINKED ISSUE #481: StepList section as functionally separate pages with scroll fix
Potentially we should consider each move to the next section as the loading of a new page. The URL updates, an item is added to history, etc. The only difference is that we preserve scroll position.

Which is all to say: perhaps what we're actually looking to do is just reset the scroll to the top of the page?

_Originally posted by @cmgiven in https://github.com/ctoec/ecis-experimental/pull/440#issuecomment-567940531_

mmazanec22:
Closes #481

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
@@ -19,7 +19,7 @@ import CommonContainer from '../../CommonContainer';
 import { hasValidationErrors } from '../../../utils/validations';
 import AlertContext from '../../../contexts/Alert/AlertContext';
 import nameFormatter from '../../../utils/nameFormatter';
-import onRouteChange from '../../../hooks/onRouteChange';
+import useRouteChange from '../../../hooks/useRouteChange';
 
 type EnrollmentNewParams = {
 	history: History;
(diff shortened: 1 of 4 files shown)


src/Hedwig/ClientApp/src/hooks/onRouteChange.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+
+export default function onRouteChange(callback: () => any) {
+	const { pathname } = useLocation();
+	useEffect(() => {
+		window.scrollTo(0, 0);
```
cmgiven: We're doing this in the callback, so this is redundant and possibly an unexpected side effect here, yeah?

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.tsx
```diff
@@ -92,31 +93,32 @@ export default function EnrollmentNew({
...
 	const visitSection = (section: Section) => {
 		updateVisitedSections({
 			...visitedSections,
-			[section.key]: true
+			[section.key]: true,
 		});
-	}
+	};
+
+	onRouteChange(() => window.scroll(0, 0));
```
cmgiven: This is nice.

src/Hedwig/ClientApp/src/hooks/onRouteChange.ts
mmazanec22: Whoops forgot to delete this after I decided to go the other direction

PR #742, opened 2020-02-10
Fix foster skipping and add test

### LINKED ISSUE #737: Enrolling a foster child does not skip income section
Expected behavior: When enrolling a child, I check the "lives with foster family" checkbox and click "Save." I expect to skip the Income Determination section. This no longer happens, appears to be a regression.

cailyncodes:
Should close #737 

Added a test for regression catching (which took way too long to add smh)

Bug was introduced in #603

### DIFF
--- a/src/Hedwig/ClientApp/src/components/StepList/StepList.tsx
+++ b/src/Hedwig/ClientApp/src/components/StepList/StepList.tsx
@@ -33,7 +33,6 @@ const mapStepsToInternalProps = function<T>(steps: StepProps<T>[], activeStep: s
 		} else {
 			status = externalStep.status(props);
 		}
-
 		const step: InternalStepProps<T> = { ...externalStep, props, status };
 
 		return step;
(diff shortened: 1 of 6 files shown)


src/Hedwig/ClientApp/src/components/StepList/StepList.tsx
```diff
@@ -33,7 +33,6 @@ const mapStepsToInternalProps = function<T>(steps: StepProps<T>[], activeStep: s
 		} else {
 			status = externalStep.status(props);
 		}
-
```
cmgiven: Taking your aggression out on innocent whitespace, Cailyn?

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+		const saveBtn = getByText(/Save/i);
+		fireEvent.click(saveBtn);
+
+		await wait();
+
+		expect(history.location.pathname).toMatch(/family-income/i)
+	});
+
+	it('skips family income section when lives with foster family is selected', async () => {
```
cmgiven: This is real nice.

src/Hedwig/ClientApp/src/components/StepList/StepList.tsx
cailyncodes: Oh haha - I went digging for the bug and must have deleted this line after adding a console.log, oops

Issue #745, opened 2020-02-11
Test results do not report on test failure


speedcm:
In the azure pipeline, test results are not published on failure.

May be due to the test task exiting with exit code from container. This was implemented to ensure that if the application build orchestrated by docker before running the tests failed, the whole pipeline failed.

AC:

test results are published on success or failure
somewhere in the pipeline, a failed app build causes pipeline failure

Linked with: https://github.com/skylight-hq/ctoec-devops/issues/15

speedcm:
The azure-pipelines.yml file was updated to always run the PublishTestResults task by adding the conditional succeededOrFailed() to the task. Now, the PublishTestResults will execute even if a previous dependency has failed, unless the run was canceled.

PR #746, opened 2020-02-11
publish test on success or failure (#745)

speedcm:
In the azure pipeline, test results are not published on failure.

May be due to the test task exiting with exit code from container. This was implemented to ensure that if the application build orchestrated by docker before running the tests failed, the whole pipeline failed.

AC:

test results are published on success or failure
somewhere in the pipeline, a failed app build causes pipeline failure

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -40,6 +40,7 @@ jobs:
 
   - task: PublishTestResults@2
     displayName: "Publish test results"
+    condition: succeededOrFailed()
     inputs:
       testResultsFormat: "VSTest"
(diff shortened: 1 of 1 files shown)


speedcm:
The azure-pipelines.yml file was updated to always run the publish test results task by adding the conditional succeededOrFailed().  Now, the publish test task will execute even if a previous dependency has failed, unless the run was canceled.   For more information on Azure Pipelines conditions, see 
* https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions

speedcm:
This PR is ready for review.

PR #749, opened 2020-02-11
Add selenium and some example tests

### LINKED ISSUE #729: Add Selenium
Set up Hedwig to run selenium and create an first example test

cailyncodes:
Should close #729

Selenium is added!!! Took a while to get Docker to play nice with everything (NOTE: needed to update WingedKeys) (https://github.com/ctoec/winged-keys/pull/42).

I have added docker containers for the latest firefox and chrome versions.

I instrumented the test code so that `yarn test` runs our unit/rtl tests. `yarn test:e2e` runs our Selenium integration tests. This makes life easier for the developer and once we set up devops in Azure for this.

Presently, running `yarn test:e2e` will only run the chrome browser (it is about 67% faster than firefox). To run under firefox, run `yarn test:e2e:firefox`.

`yarn test:e2e` will fail on windows unless a medley of things are installed and configured.

~~Oh also: I didn't know if our staging username and password are sensitive data, so I replaced that with `''` strings. So in order for log in test to pass, update those values. Welp on the second commit I forgot to remove them -- so they are no longer sensitive data -- will change. Let me know if I should change them back to these values.~~ Just ignore all of this. I now remember that it is public in winged keys so this is all a moot point. Carry on!


mmazanec22:
Considering that IE and Edge are the PITA browsers that we definitely need to support, and most of us are developing in Chrome and FF, maybe we should default to IE rather than Chrome?

cmgiven:
Do we want a `yarn test:unit` so that `yarn test` can continue to run everything? Or are you suggesting that we should wait for that until we get Azure squared away.

cailyncodes:
> Considering that IE and Edge are the PITA browsers that we definitely need to support, and most of us are developing in Chrome and FF, maybe we should default to IE rather than Chrome?

There are no docker containers with IE/Edge (https://github.com/SeleniumHQ/docker-selenium) so they would need to be installed locally.

cailyncodes:
> Do we want a `yarn test:unit` so that `yarn test` can continue to run everything? Or are you suggesting that we should wait for that until we get Azure squared away.

Yes, I think we should wait for Azure

src/Hedwig/ClientApp/src/e2e/App.e2e.test.ts
```diff
@@ -0,0 +1,50 @@
+import { DriverHelper } from './helpers';
+import { app, clickLogin, waitForLoginWelcome, waitForHeader } from './pages/App';
```
cmgiven: I'm torn on whether these should be broken out into a separate file. Although nice for readability, and maybe preventing some duplication down the road (?), I wonder if inlining would help make clear exactly what is being tested. It's not particularly cumbersome, and I like the comments that you've left in `pages/App`.

cailyncodes: I reorganized the code and added some RTL-like utilities -- let me know what you think of this new approach

PR #755, opened 2020-02-13
Add c4k count to report

### LINKED ISSUE #444: Add count to Care 4 Kids field on report
As a provider, I want to see a count of Care 4 Kids recipients on my report to verify Care 4 Kid funding information was entered correctly before submitting my report.

**Acceptance Criteria:**
- Show a count of the total number of children enrolled for a given reporting period next to the Care 4 Kids field in the "Other revenue" portion of the report (see design below)

**Design:**
image

cailyncodes:
Should close #444 

Let me know if this is the desired functionality

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
@@ -41,26 +41,29 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
 
 	const childrenParams: ApiOrganizationsOrgIdChildrenGetRequest = {
 		orgId: getIdForUser(user, 'org'),
-		include: ['fundings']
+		include: ['fundings'],
+		startDate: report.reportingPeriod.periodStart,
+		endDate: report.reportingPeriod.periodEnd
 	}
 	const [, , childMappedEnrollments] = useApi(
 		(api) => api.apiOrganizationsOrgIdChildrenGet(childrenParams),
-		[user]
+		[user, report]
 	);
 	const [care4KidsCount, setCare4KidsCount] = useState(0);
 
 	useEffect(() => {
 		if (childMappedEnrollments) {
-			const c4kChildren = Object.values(childMappedEnrollments).filter(enrollmentGroup => {
-				const currentEnrollment = enrollmentGroup.find(enrollment => {
-					return enrollment.exit === null;
+			const c4kChildren = Object.values(childMappedEnrollments)
+				.filter(enrollmentGroup => {
+					const currentEnrollment = enrollmentGroup.find(enrollment => {
+						return enrollment.exit === null;
+					});
+					if (!currentEnrollment) {
+						return false;
+					}
+					const fundings = currentEnrollment.fundings;
+					return currentC4kFunding(fundings) !== undefined;
 				});
-				if (!currentEnrollment) {
-					return false;
-				}
-				const fundings = currentEnrollment.fundings;
-				return currentC4kFunding(fundings) !== undefined;
-			});
 			setCare4KidsCount(c4kChildren.length)
 		}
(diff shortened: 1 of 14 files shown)


src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -1,12 +1,12 @@
-import React, { useState, FormEvent, useContext } from 'react';
-import { CdcReport, ApiOrganizationsOrgIdReportsIdPutRequest } from '../../../generated';
-import { Mutate } from '../../../hooks/useApi';
+import React, { useState, useEffect, useContext } from 'react';
+import { CdcReport, ApiOrganizationsOrgIdReportsIdPutRequest, ApiOrganizationsOrgIdChildrenGetRequest, Enrollment } from '../../../generated';
+import useApi, { Mutate } from '../../../hooks/useApi';
```
cmgiven: I had no idea you could use this syntax to import default and named exports at the same time. Mind blown.

jlhogan:
my biggest `??` here is about performance. Wondering how this would perform for a large org, and if it would be worth it then

to Chris's point about 1 or 2 endpoints, i think it depends on if we consider these kids/this funding part of the Cdc report that's being submitted (which it seems we do, since this #is on the report). If that's the case, it makes more sense to my mind to include this in the same endpoint.

cmgiven:
It's such an oddity.

Reasons to consider it part of the report:

* We ask for it on the report.
* We want a forcing function to keep more kids' C4K status up-to-date.

Reasons to consider it something else:

* The kids being counted won't appear on the report (either on the PSR or in ECE Reporter).
* OEC's reasons for asking this question have to do with wanting a holistic picture of the program's fiscal condition. It's kind of unrelated to the enrollment.

I think I might be starting to err toward it being something else, and thus maybe more properly its own endpoint. I also wonder about performance, put also note that if we break it out, that would create the opportunity to progressively load this information into the page.

cailyncodes:
I am also team it's own thing. I think the performance question will be present (in different forms) regardless of whether it is in the report end point (or front end filtering) or it's own. And I think it is semantically clearer. Seconding Chris's points for separate and adding that "We ask for it on the report" seems purely syntactical.

I am seeing the next step as `asOf` before merging. Anything else that should be addressed in this PR?

cmgiven:
`asOf` is all I've got.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -38,6 +39,36 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
...
+			const c4kChildren = Object.values(childMappedEnrollments)
+				.filter(enrollmentGroup => {
+					const currentEnrollment = enrollmentGroup.find(enrollment => {
+						return enrollment.exit === null;
+					});
+					if (!currentEnrollment) {
+						return false;
+					}
+					const fundings = currentEnrollment.fundings;
+					return currentC4kFunding(fundings) !== undefined;
```
cmgiven: Does this need to be updated for `asOf`? We're not necessarily looking for the current enrollment/funding anymore.

src/Hedwig/Controllers/ChildrenController.cs
```diff
@@ -0,0 +1,46 @@
...
+		)
+		{
+			_validator = validator;
+			_children = children;
+		}
+
+		[HttpGet]
+		[ProducesResponseType(StatusCodes.Status200OK)]
+		[ProducesResponseType(StatusCodes.Status204NoContent)]
+		public async Task<ActionResult<IDictionary<Guid, ICollection<Enrollment>>>> Get(
```
cmgiven: Is this easier if we just request the enrollments instead of the kids? Is there a reason to base this around Child?

Issue #757, opened 2020-02-13
enrollment update bug


jlhogan:
Updating an enrollment more than once results in 500 due to:
```
System.InvalidOperationException: The instance of entity type 'User' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached. Consider using 'DbContextOptionsBuilder.EnableSensitiveDataLogging' to see the conflicting key values.
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.ThrowIdentityConflict(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry, Boolean updateDuplicate)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges, Boolean modifyProperties)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.PaintAction(EntityEntryGraphNode`1 node)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph[TState](EntityEntryGraphNode`1 node, Func`2 handleNode)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph[TState](EntityEntryGraphNode`1 node, Func`2 handleNode)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph[TState](EntityEntryGraphNode`1 node, Func`2 handleNode)
   at Microsoft.EntityFrameworkCore.DbContext.SetEntityState[TEntity](TEntity entity, EntityState entityState)
   at Microsoft.EntityFrameworkCore.DbContext.Update[TEntity](TEntity entity)
   at Hedwig.Repositories.EnrollmentRepository.UpdateEnrollment(Enrollment enrollment) in /app/src/Hedwig/Repositories/EnrollmentRepository.cs:line 27
[...]

PR #765, opened 2020-02-17
Attempt read-only properties

jlhogan:
This is by no means a done deal. Interested in all yalls thoughts about this. Also there is a small mystery: 
When updating enrollment with duplicative  references to author (enrollment.Author = enrollment.Child.Author), I was seeing the bug described in #757 
HOWEVER, when trying to recreate the bug by having a funding point to the same reporting period twice (funding.FirstReportingPeriod = funding.LastReportingPeriod), I could not repro the error
So, not sure what's different between these two cases that leads to one causing a dbContext error and the other NOT.... But this feels like a more robust fix than https://github.com/ctoec/ecis-experimental/pull/759, and allows us to build programmatic processes to create Users/ReportingPeriods via Db context

### DIFF
--- a/docs/backend/2020_02_05_read-only-entities.md
+++ b/docs/backend/2020_02_05_read-only-entities.md
@@ -20,9 +20,9 @@ EF Core Attribute.
 
 ## Status
 * Proposed
-* __Accepted__
+* Accepted
 * Rejected
-* Superceded
+* __Superceded__
 * Accepted (Partially superceded)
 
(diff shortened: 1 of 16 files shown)


jlhogan:
won't implement. https://github.com/ctoec/ecis-experimental/issues/777 leading us to decide to implement DTOs, which will enable us to protect read-only properties at the DTO<-> entity translation layer

PR #769, opened 2020-02-17
710 uncaught api error

### LINKED ISSUE #710: Alert 400s that are not shown on page
If you take #709 as an example, the server returns a 400 for a failed validation when trying to edit any information. However, as that validation is not shown on the current screen, the effect for a user is a silent fail. We should have a way of catching 400 validation errors that do not correspond to any field on the page, both to alert the user that something has gone wrong (probably using the validation message), as well as to throw some sort of exception for us so we know when this happens, as 400s will be a pretty common and expected error.

**AC:**
* When an enrollment validation fails but cannot be corrected by a field on the current page (i.e. no red text appears next to any field), the user should see an error.
  * Headline: "Something went wrong"
  * Body:
    > We couldn't save your changes because we detected a potential inconsistency in the data. This shouldn't have happened. Our team has been notified of the problem.
    >
    > You may still be able to update other children on your roster. Please wait to submit a report to OEC until the problem is fixed. For updates, contact oec-data-pilot@skylight.digital.
* An error is thrown and captured by Sentry. The error includes the validation that failed and the page on which the error occurred. It does not include any data from the record itself, which would by PII.

jlhogan:
Repurposes the `error` entity of `ApiState` from `useApi` to pass around errors for pages using the associated apistate entity and api mutate function for that entity. 

Adds handling for un-UI-handled errors 

closes #710

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/App/App.tsx
+++ b/src/Hedwig/ClientApp/src/containers/App/App.tsx
@@ -27,10 +27,13 @@ const App: React.FC = () => {
 		orgId: getIdForUser(user, 'org'),
 	};
 
-	const [loading, error, reports] = useApi(api => api.apiOrganizationsOrgIdReportsGet(params), [
-		user,
-		cacheInvalidator,
-	]);
+	const [loading, error, reports] = useApi(
+		api => api.apiOrganizationsOrgIdReportsGet(params),
+		[user, cacheInvalidator],
+		{
+			skip: !user,
+		}
+	);
 
 	const pendingReportsCount =
 		!loading &&
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import ChildInfo from './ChildInfo';
+import { DeepNonUndefineable } from '../../../utils/types';
+import { Enrollment } from '../../../generated';
+
+jest.mock('../../../hooks/useApi');
+
+describe('enrollment sections', () => {
+	describe('ChildInfo', () => {
+		it('shows an error if rendered without a child first name', async () => {
+			const { findByText } = render(
+				<ChildInfo.Form
+					siteId={1}
+					enrollment={{} as DeepNonUndefineable<Enrollment>}
+					mutate={async () => {}}
+					error={{
+						errors: { 'Child.FirstName': ['error'] },
+						status: 400,
+					}}
+				/>
+			);
+
+			// TODO: the id needs to be updated-- this will probably mess things up
+			const firstNameErr = await findByText('This information is required for enrollment');
+			expect(firstNameErr.id).toBe('child-firstname-error');
+		});
+
+		it('shows an error if rendered without a child last name', async () => {
+			const { findByText } = render(
+				<ChildInfo.Form
+					siteId={1}
+					enrollment={{} as DeepNonUndefineable<Enrollment>}
+					mutate={async () => {}}
+					error={{
+						errors: { 'Child.LastName': ['error'] },
+						status: 400,
+					}}
+				/>
+			);
+
+			const lastNameErr = await findByText('This information is required for enrollment');
+			expect(lastNameErr.id).toBe('child-lastname-error');
+		});
+	});
+});
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
@@ -21,9 +21,8 @@ describe('enrollment sections', () => {
 				/>
 			);
 
-			// TODO: the id needs to be updated-- this will probably mess things up
-			const firstNameErr = await findByText('This information is required for enrollment');
-			expect(firstNameErr.id).toBe('child-firstname-error');
+			const firstNameInput = (await findByText('First name')).closest('div');
+			expect(firstNameInput).toHaveTextContent('This information is required for enrollment');
 		});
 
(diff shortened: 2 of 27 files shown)


src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -73,6 +76,27 @@ export default function useApi<TData>(
...
 
+  // Create error parsing function
+	const handleError = async (error: any) => {
+		let jsonResponse: Object | null = null;
+		try {
+			jsonResponse = JSON.parse(await error.text());
+		} catch {}
+
+		let apiError: ApiError | null = null;
+		if(jsonResponse) {
```
jlhogan: this means that if the error thrown is not json-parse-able, the error obj will remain null

since we have an API spec, the only valid errors returned from the API are the types defined in the generated client models, which come from Json.

right now the only non-json error bodies we get back are the empty 403s, which happen in dev and staging, and I think are best ignored.

However i'm not sure that totally ignoring non-json errors is the right move here. Maybe we want to throw an error?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import ChildInfo from './ChildInfo';
+import { DeepNonUndefineable } from '../../../utils/types';
+import { Enrollment } from '../../../generated';
+
+jest.mock('../../../hooks/useApi');
+
+describe('enrollment sections', () => {
```
jlhogan: Stopped here (only doing these two section tests) b/c I don't know if this is the right way to be testing. So many random strings being matched. Is there an easy way to do this at a higher level? Visual snapshots on e2e tests?

src/Hedwig/ClientApp/src/containers/App/App.tsx
```diff
@@ -27,10 +27,13 @@ const App: React.FC = () => {
...
 
-	const [loading, error, reports] = useApi(api => api.apiOrganizationsOrgIdReportsGet(params), [
-		user,
-		cacheInvalidator,
-	]);
+	const [loading, error, reports] = useApi(
+		api => api.apiOrganizationsOrgIdReportsGet(params),
+		[user, cacheInvalidator],
+		{
+			skip: !user,
```
cmgiven: Yes!

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
cmgiven: I think this style works? Something like `Child.FirstName` seems fine to me. Small suggestion on the assert.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+					error={{
+						errors: { 'Child.FirstName': ['error'] },
+						status: 400,
+					}}
+				/>
+			);
+
+			// TODO: the id needs to be updated-- this will probably mess things up
+			const firstNameErr = await findByText('This information is required for enrollment');
+			expect(firstNameErr.id).toBe('child-firstname-error');
```
cmgiven: Checking the ID of the result seems indirect. I think this is closer to RTL idioms:

```suggestion
			const firstNameInput = getByLabelText('First name').closest('div');
			expect(firstNameInput).toHaveTextContent('This information is required for enrollment');
```

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -58,12 +59,32 @@ const ChildInfo: Section = {
...
+		const [hasAlertedOnError, setHasAlertedOnError] = useState(false);
+		useFocusFirstError([error]);
+		useEffect(() => {
+			if (error && !hasAlertedOnError) {
+				if (!isBlockingValidationError(error)) {
+					throw new Error(error.title || 'Unknown api error');
+				}
+				setAlerts([validationErrorAlert]);
+			}
+		}, [error, hasAlertedOnError]);
```
cmgiven: Took me too long to reason through this, not realizing that effects only happen _after_ render. Makes sense now.

src/Hedwig/ClientApp/src/utils/validations/errorForField.ts
```diff
@@ -30,15 +33,23 @@ export function warningForField<T extends Validatable>(
...
-	const fieldError = processBlockingValidationErrors(fieldId, error.errors);
+	const fieldError = processBlockingValidationErrors(
+		fieldId,
+		(error as ValidationProblemDetails).errors
+	);
 
 	if (fieldError) {
+		if (!hasAlertedOnError) {
+			setHasAlertedOnError(true);
+		}
```
cmgiven: This seems unnecessary. Is there a reason not to just call `setHasAlertedOnError(true)`?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
jlhogan: I think this should be abstracted into a common form or apiForm element for the app, where we can make descriptive function names and/or add more in-depth commenting, but didn't want to further beef up these already long files

src/Hedwig/ClientApp/src/utils/validations/errorForField.ts
jlhogan: I thought i was getting an infinite rendering loop without it, but must have been from something else! will clean up.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
cmgiven: Yup, that makes sense.

PR #774, opened 2020-02-20
Add site column

### LINKED ISSUE #764: Add a site column to the roster for multi-site organizations
As a provider at a multi-site organization, I want to easily sort by site in my roster.

**AC:**
- For multi-site organizations, add a site column to the roster displaying the child's enrollment site
  - If a user is viewing the entire roster, or a single site, this column should persist
  - If the organization only has one site, do not show this column (as is today)
- Determine a solution that prevents awkward breaks - right now we manually set the width of columns in the table; we should test long child and site name combinations
- Selecting the sort function on this column sorts the column alphabetically by site name (within each age group section)

**Design:**
Screen Shot 2020-02-18 at 2 47 29 PM

cailyncodes:
Should close #764

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
@@ -34,84 +34,83 @@ export default function AgeGroupSection({
 }: AgeGroupSectionProps) {
 	if (!enrollments.length) return null;
 
-	let columns: Column<DeepNonUndefineable<Enrollment>>[] = [];
-	columns = [
-		...columns, 
-		...[
-			{
-				name: 'Name',
-				cell: ({ row }: { row: DeepNonUndefineable<Enrollment> }) => (
-					<th scope="row">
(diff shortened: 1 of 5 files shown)


src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
```diff
@@ -77,22 +76,48 @@ export default function AgeGroupSection({
...
-				width: "20%",
-			},
-		],
+				sort: (row: DeepNonUndefineable<Enrollment>) => (row.site.name || '').toLowerCase(),
+				width: '10%'
+			}
+		]
+	}
+	columns = [
+		...columns,
```
mmazanec22: If we want the site col to be penultimate maybe we should include enrollment date above, and then add site specifically at the index where we want it, maybe with a comment about why?  Otherwise this feels like it's at a high risk of getting lost or ignored in a refactor.

cailyncodes: I went with a third approach that I think adds the clarity here without needing to use magic numbers

PR #776, opened 2020-02-21
Add console and storybook axe and remove flag preventing accessibility related tests from running in CI

mmazanec22:


src/Hedwig/ClientApp/src/index.tsx
```diff
@@ -13,6 +13,12 @@ import App from './containers/App/App';
 import 'react-dates/lib/css/_datepicker.css';
 import './assets/styles/index.scss';
 
+// Is there a better place for this?
```
cmgiven: Sometimes I've seen things broken out into a devtools file. But this doesn't bug me.

src/Hedwig/ClientApp/src/utils/stringFormatters/nameFormatter.ts
```diff
@@ -21,6 +21,7 @@ export function lastFirstNameFormatter(name?: Child) {
 		return '';
 	}
 
+	// TODO: IF FIRST NAME IS UNDEFINED THIS FORMATTING IS SILLY
```
cmgiven: First name should never be undefined. We block saving without at least a first and last name.

PR #786, opened 2020-02-24
Formats all c sharp code with dotnet-formatter

jlhogan:
also:
- adds format check step to pipeline
- Updates docker-compose file to expose debug ports in e2e containers by
default
- Updates docker-compose to have dependency chain so that
`docker-compose up client` will bring up the whole app but not selenium
contains

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -17,6 +17,8 @@ services:
     volumes:
       - ./src/Hedwig:/app/src/Hedwig:cached
       - ./test/HedwigTests:/app/test/HedwigTests:cached
+      - ./Hedwig.sln:/app/Hedwig.sln
+      - ./.editorconfig:/app/.editorconfig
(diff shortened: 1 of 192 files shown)


docker-compose.yaml
```diff
@@ -99,29 +109,15 @@ services:
...
+    ports:
+      - 5901:5900
   e2e-firefox:
     image: selenium/node-firefox
     depends_on:
       - selenium-hub
     environment:
       - HUB_HOST=selenium-hub
       - HUB_PORT=4444
-  # Uncomment these (and comment the above e2e-* containers) to allow for VNC connection to the automated browser
```
cailyncodes: We should leave these in. They are needed if you want to see the actual interaction Selenium does.

jlhogan: ah sorry missed that they were diff containers! will undo this!

PR #788, opened 2020-02-24
Update top bar header

### LINKED ISSUE #775: Add program name to top bar
This ticket is an enhancement to the top bar to provide a more personalized experience.

**AC:**
- Include "ECE Reporter" in top bar (note: do not include the word "OEC" shown in design screenshot)
- Add the program (top level organization) name to the top bar in bold

## Design
Screen Shot 2020-02-20 at 6 08 45 PM

cailyncodes:
Should close #775 

Adds program name to header, uses h1, h2 for accessibility (should fix an axe issue)

Fixes broken e2e test


src/Hedwig/ClientApp/src/components/Header/Header.tsx
```diff
@@ -83,11 +85,16 @@ class HeaderWithoutRouter extends React.Component<HeaderProps & RouteComponentPr
...
-				<header className={cx('usa-header', 'usa-header--extended')} role="banner">
+				<header className={cx('usa-header', 'usa-header--extended', 'usa-header--oec-default')} role="banner">
 					<div className={cx('usa-navbar')}>
 						<div className={cx('usa-logo')} id="extended-logo">
-							<Link to="/" aria-label={`${title} home`}>
-								<span className={cx('usa-logo__text')}>{title}</span>
+							<Link to="/" aria-label={`${primaryTitle} home`}>
+								<div className={cx('usa-logo__text', 'display-flex', 'flex-align-center')}>
+									<div>
+										<h1>{primaryTitle}</h1>
```
mmazanec22: It's a common practice (and even some accessibility-related tutorials show this structure), but header elements in the nav bar might be bad usability for people using screen readers:

[Deque's recommendation](https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one):
"Usually the best practice is to start the main content of a web page with a level 1 heading (h1), with no other headings before this high-level heading. [...] If there is no h1, or if the h1 appears somewhere other than at the start of the main content, screen reader users must listen to more of the web page to understand its structure, wasting valuable time."

[The Paciello group would flag it as a problem](https://github.com/ThePacielloGroup/ARC-Tracker) (no. 362, "h1InHeader").

"If documents are created properly, they can often form an outline of headings, which serves not only as a way to skip past the navigation but also lets screen reader users "scan" the main ideas of a document without having to read the whole thing. Defining your primary document heading at the beginning of your main content as an <h1> can provide very quick access to the beginning of the main content." [(source)](https://webaim.org/techniques/skipnav/)

mmazanec22:
Realized that my comment could have been more helpful: I suspect the axe issue you saw is caused by our using empty divs instead of any kind of loading screen (which would ideally contain at least an h1 and probably more stuff), and also not having any kind of log in screen yet

src/Hedwig/ClientApp/src/e2e/App.e2e.test.ts
```diff
@@ -30,7 +30,7 @@ describe('Smoke screen', () => {
 		try {
 			const root = await load(driver, appUrl);
 			const { getByLocator } = render(root);
-			const header = await getByLocator({ css: 'header em' });
+			const header = await getByLocator({ css: 'header h1' });
```
mmazanec22: This should probably be changed to accommodate reversion to divs

cailyncodes: Good catch!

PR #789, opened 2020-02-25
Add comment text box

### LINKED ISSUE #628: Add a comment text box to the CDC report
As a provider, I can include a note to the OEC with my monthly report, so I don't need to send an email or call to communicate a unique circumstance/issue.

**Acceptance Criteria:**
- Adds an optional text box field to the CDC report detail page,
- 400 character limit (cannot submit report if longer, button remains disabled - _open to implementation ideas here, if it makes sense to have a post-submit validation error)_
- Include "optional" tag, and allow for report submission if field is blank
- Label (note: slightly different from design screenshot): 
     > **Anything to share with the Office of Early Childhood about your report?** (optional) 
     > You can include up to 400 characters in your note. 

**Design:**
Screen Shot 2020-02-12 at 10 20 59 AM

### LINKED ISSUE #682: Convert old tests from Enzyme to RTL
Tests:
- [x] Component folder (@cailyncodes) (PR: #701)
- [x] App (@cailyncodes) (PR: #705)
- [x] Roster (@cailyncodes) (PR:  #706)
- [x] Enrollment[Edit|Detail] (@mmazanec22 ) (PR: #711)
- [x] ReportsSummary (@cmgiven) (PR: #714)
- [x] ReportDetail folder (@cmgiven) (PR: #714)

cailyncodes:
Should close #628 

Adds a migration for a text box on reports

Adds a clientError for the text if it is more than 400 characters

Updates TextInput to allow for textarea


cailyncodes:
There's a failure on the dotnet formatting in Azure pipelines. Flagging for @jlhogan if she has any ideas?

jlhogan:
To format the backend code, run `make dc-dotnet-format` or execute the `dotnet format` command inside the backend container!

Dockerfile
```diff
@@ -2,6 +2,6 @@ FROM mcr.microsoft.com/dotnet/core/sdk:3.0
 RUN apt-get update \
 	&& apt-get install -y --no-install-recommends unzip \
 	&& curl -sSL https://aka.ms/getvsdbgsh | bash /dev/stdin -v latest -l /vsdbg \
-	&& dotnet tool install --global dotnet-ef --version 3.0.0
-    && dotnet tool install --global dotnet-format
+	&& dotnet tool install --global dotnet-ef --version 3.0.0 \
+	&& dotnet tool install --global dotnet-format
```
jlhogan: :( sorry !

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -116,6 +118,8 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
 	}
 	const { isExecuting: isMutating, setExecuting: onSubmit } = usePromiseExecution(_onSubmit);
 
+	const invalidData = comment ? comment.length > 400 : false;
```
jlhogan: should we do this validation on the backend to keep all validations in one place? Eventually we should deprecate & clean up `clientErrorForField`

src/Hedwig/Models/Reports/CdcReport.cs
```diff
@@ -15,5 +15,7 @@ public class CdcReport : OrganizationReport
 	[Column(TypeName = "decimal(18,2)")]
 	[Required]
 	public decimal? FamilyFeesRevenue { get; set; }
+
+	public string Comment { get; set; }
```
jlhogan: enforce length validation here?

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
cailyncodes: I'm not opposed in theory -- AC asks for the submit button to be blocked but perhaps that isn't as important as cleaner validations

Dockerfile
cailyncodes: No worries!

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
jlhogan: while overall my inclination is to keep all the validations in the backend I also see the merit in giving a user immediate feedback on when their input is invalid ... writing a chunk of text and then not knowing when it's short enough to submit is probably a VERY annoying user experience, and maybe worth deviating from the norm for.

cailyncodes: I think it's valuable to also have a backend validation but will add back a frontend one -- should we use `clientError` or deprecate that and go a different route?

jlhogan: use `clientError` for sure. we can figure out if something else is more approp later

src/Hedwig/ClientApp/src/components/TextInput/TextInput.tsx
```diff
@@ -34,7 +48,59 @@ export function TextInput({
...
 	inputProps,
 	inline,
-}: TextInputProps) {
+}: TextInputHTMLInputElementProps | TextInputHTMLTextAreaElementProps) {
+	let inputElement;
+	switch (type) {
+		case 'textarea':
+			inputElement = (
+				<textarea
+					className={cx('usa-textarea')}
```
mmazanec22: We can probably abstract shared props into an object so this will be less prone to inconsistency

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -218,6 +218,7 @@ const ChildInfo: Section = {
 				<div className="grid-row grid-gap">
 					<div className="mobile-lg:grid-col-12">
 						<TextInput
+							type="input"
```
mmazanec22: Maybe since we use inputs a lot more we can default to this type and specify the text area type only when we need it?

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -196,6 +199,24 @@ export default function ReportSubmitForm({ report, mutate, canSubmit }: ReportSu
...
+								Anything to share with the Office of Early Childhood about your report?
+							</span>
+						}
+						defaultValue={comment || ''}
+						onChange={e => setComment(e.target.value)}
+						disabled={!!report.submittedAt}
+						status={serverErrorForField(
+							'comment',
+							apiError,
+							'You can enter at most 400 characters.'
```
mmazanec22: If we have a character limit maybe we should use a hint or something to indicate that?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
cailyncodes: I tried but typescript got mad -- but it was perhaps just the way I was going about it

cailyncodes: I tried but typescript got mad -- but it was perhaps just the way I was going about it

cmgiven:
Question: Why are we limiting the length of the text? It's only available to authenticated program users, so I don't have much concern about abuse. And the size of the text box will naturally restrict.

@abbyraskin

cmgiven:
(There probably should be an upper limit short of the half-million unicode characters that the database supports. But 400 characters is something you're actually likely to hit, as opposed to something like 16,000 characters.)

cailyncodes:
Okay pending the char limit, I'm going to work on extracting Julia's additions so we can get those in

abbyraskin:
The goal was to keep these submissions short so as not to overwhelm OEC. But I see your point - let's just nix the character limit for this first go around and keep an eye on it (We should also cut the "You can include up to 400 characters in your note" copy). @cailyncodes @cmgiven

abbyraskin:
The goal was to keep these submissions short so as not to overwhelm OEC. But I see your point - let's just nix the character limit for this first go around and keep an eye on it (We should also cut the "You can include up to 400 characters in your note" copy). @cailyncodes @cmgiven

cailyncodes:
We eventually got there team!

PR #792, opened 2020-02-25
fixup formatting to use only tabs

jlhogan:
also ensure backend formatting runs over test code as well
(build failure is just a warning, and does not break to formatting)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ dotnet-format:
 	dotnet format
 
 dc-dotnet-format:
-	./dc-backend dotnet format
+	docker-compose exec backend sh -c "cd /app && dotnet format"
 
 api-generate:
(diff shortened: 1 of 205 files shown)


Makefile
```diff
@@ -57,7 +57,7 @@ dotnet-format:
 	dotnet format
 
 dc-dotnet-format:
-	./dc-backend dotnet format
```
cailyncodes: This old approach should work?

We should also probably use the `./dc` helper if we are changing it though

jlhogan: the `working_dir` for the backend container is `app/src/Hedwig`; we need to run this commend at the root `/app` level to get it to apply to all files. 

updating to use `.dc` helper tho!

cailyncodes: We can also probably fix the docker-compose file?

jlhogan: what should be fixed in docker-compose file?

PR #795, opened 2020-02-25
Prevent withdraw for missing info

### LINKED ISSUE #556: Prevent withdrawal of child with missing information
To encourage complete/high quality data entry, and to prevent a user from a situation where they can't submit a report because they're missing information for a kid who's no longer on their roster, we should alert users to update missing information before withdrawing a kid.

**AC:**
- If a user tries to withdraw a funded kid with missing information (selects "Withdraw" button), display an alert at the top of the child's summary page, blocking the user from withdrawing any incomplete enrollments.
- This alert replaces (and encompasses the intent of) the current alert, which notifies users that they need to enter an enrollment start date.
- Red alert 
  - Header: Information needed to withdraw child
  - Body: 
     > To withdraw a child from a funded space in your program, they cannot have any missing information. Please enter all missing information indicated below to withdraw this child.

cailyncodes:
Should close #556

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
@@ -29,6 +29,7 @@ import {
 	clientErrorForField,
 	serverErrorForField,
 	useFocusFirstError,
+	hasValidationErrors,
 } from '../../utils/validations';
 import ReportingPeriodContext from '../../contexts/ReportingPeriod/ReportingPeriodContext';
 import { processBlockingValidationErrors } from '../../utils/validations/processBlockingValidationErrors';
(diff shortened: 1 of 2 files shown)


src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -89,26 +90,28 @@ export default function Withdrawal({
...
 	useEffect(() => {
-		if (cannotWithdraw) {
+		if (isMissingInformation) {
 			setAlerts([
 				{
 					type: 'error',
 					heading: 'Information needed to withdraw child',
 					text:
-						'To withdraw a child from a funded space in your program, "first reporting period" cannot be blank. Please update this information for the child before withdrawing',
+						'To withdraw a child from a funded space in your program, they cannot have any missing information. Please enter all missing information indicated below to withdraw this child.',
```
mmazanec22: There's a file with a bunch of other alert copy-- might be worth putting this in there

PR #802, opened 2020-02-26
Fix missing info that was over zealous and dropdown bugs on income not disclosed

### LINKED ISSUE #622: Missing information icon shows in summary on complete enrollment/funding section
But only when there is a CDC funding (private pay not affected)

cailyncodes:
Should close #622

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -66,16 +66,10 @@ const EnrollmentFunding: Section = {
 	key: 'enrollment-funding',
 	name: 'Enrollment and funding',
 	status: ({ enrollment }) =>
-		enrollment && (
-			sectionHasValidationErrors([enrollment.fundings]) ||
-			processValidationError(
-				'ageGroup',
-				enrollment ? enrollment.validationErrors : null
-			) ||
-			processValidationError(
-				'entry',
-				enrollment ? enrollment.validationErrors : null
-			))
+		enrollment &&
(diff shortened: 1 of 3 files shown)


cmgiven:
PR reviews appear to be broken at the moment, but here's my comment:

> Maybe some tests here? Throw some errors at the status and check that it only responds to fundings, ageGroup, and entry?
> 
> Good use of a comment to explain populating the funding type options.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -321,7 +325,18 @@ const EnrollmentFunding: Section = {
 		const [fundingTypeOpts, setFundingTypeOpts] = useState<{ value: string; text: string }[]>([]);
 		useEffect(() => {
 			const orgFundingSpaces = idx(site, _ => _.organization.fundingSpaces);
-			if (!orgFundingSpaces || familyDeterminationNotDisclosed(enrollment)) return;
+			// If the organization has no funding spaces
+			// Or if the family income is not disclosed and no previous cdc funding had been entered
+			// We do not want to automatically delete data if it has already been entered
```
jlhogan: This comment is not totally clear -- is ```we do not want to automatically delete data if it has already been entered``` and explanation for why `!cdcFunding` is part of the condition? If so, i think that could be made more explicit. I think my brain is expecting this comment to include the description of what we do when this condition is met, not just explanation of the condition itself. something liek:
```
// If the organization has no funding spaces
// or the family income is not disclosed and no previous cdcFunding exists (need preserve prior funding state even if incorrect)
// then the only valid funding type is private pay 
```

Issue #812, opened 2020-02-27
devops - pr system - increase disk space for PR system for docker


speedcm:
Increase diskspace for the PR system generated by Terraform for docker build

cailyncodes:
Closed by #813

PR #813, opened 2020-02-27
increase pr system disk size (#812)

speedcm:
This PR is to increase the disk size of the PR system to provide sufficient diskspace for the docker build process.


speedcm:
This issue is now resolved.  This PR was verified will pull request system: https://18.188.196.134:5001.   The resolution was to change the docker command to: docker-compose up --build client in the startup.sh script.  The startup.sh script is executed when a EC2 system is spun up via Terraform.

Issue #767, opened 2020-02-17
Form component


jlhogan:
At this point theres enough repeated code in the form components that it would likely be beneficial to abstract it into a form component

things to abstract:
- Form state initialization
- Form component with defn
- submit button
.... (please add!)

cailyncodes:
Related to #823

PR #823, opened 2020-02-28
Proof of concept general-use form component

cailyncodes:
Related to #767


cailyncodes:
cc @mmazanec22 @jlhogan 

Just some thoughts for how to do the form component.

`Form` would also take in the `enrollment` object and call `useReducer` inside.

`FormField` would also take as a prop a function for accessing the value on `enrollment` similar to what `idx` does

jlhogan:
I think if we're implementing the form as a component, it should take the backing data model (in this case entity) as a param, then we can potentially also use for report !

cailyncodes:
Closed by #923

PR #826, opened 2020-03-02
Context for non-blocking validations

jlhogan:
to avoid hydrating entities for validation with other entities in the hierarchy
makes a "parent entity" object available. Right now only necessary in `Rules.Fundings.IfCDC_AndEnrollmentEntry_FirstReportingPeriodIdRequired`


old
```diff
@@ -0,0 +1,2856 @@
+[
```
cailyncodes: This file and `new` should probably be delete?

PR #832, opened 2020-03-03
Refactor tests

cailyncodes:
Simplifies `CommonContextMockProvider` and `useApi` mock. Mostly addresses the issue that Chris identified in the report tests -- I think this is much cleaner.

Does have the downside that import order matters but I am comfortable with that tradeoff


cailyncodes:
Hmmmmmmm I update the snap shot but still is failing :/

PR #835, opened 2020-03-04
Use form reducer

mmazanec22:
Converted report, withdrawal, and most of enrollment flow.  IMO refactoring funding and updateData and making a form component make sense as a separate PR.

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
@@ -231,7 +231,7 @@ const ChildInfo: Section = {
 							id="sasid"
 							label="SASID"
 							name="child.sasid"
-							onChange={updateFormData(newSasid => newSasid)}
+							onChange={updateFormData()}
 							defaultValue={sasid || ''}
 							optional
 						/>
@@ -243,7 +243,7 @@ const ChildInfo: Section = {
 							label="First name"
 							defaultValue={firstName || ''}
 							name="child.firstName"
-							onChange={updateFormData(newFirstName => newFirstName)}
+							onChange={updateFormData()}
 							status={initialLoadErrorGuard(
 								initialLoad,
 								serverErrorForField(
@@ -263,7 +263,7 @@ const ChildInfo: Section = {
 							label="Middle name"
 							defaultValue={middleName || ''}
 							name="child.middleName"
-							onChange={updateFormData(newMiddleName => newMiddleName)}
+							onChange={updateFormData()}
 							optional
 						/>
 					</div>
@@ -275,7 +275,7 @@ const ChildInfo: Section = {
 								label="Last name"
 								defaultValue={lastName || ''}
 								name="child.lastName"
-								onChange={updateFormData(newLastName => newLastName)}
+								onChange={updateFormData()}
 								status={initialLoadErrorGuard(
 									initialLoad,
 									serverErrorForField(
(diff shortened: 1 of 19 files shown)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -355,6 +355,7 @@ const EnrollmentFunding: Section = {
 						updatedFundings.push(
 							updateFunding({
 								currentFunding: sourcelessFunding,
+								// Why are we setting CDC as the source by default?  Is this the cause of our overenrollment weirdness?
```
jlhogan: if you look up on line 344, this is all happening in a switch against `fundingSelection.source` match on `FundingType.CDC` case, so the user has selected CDC funding. If the user has not selected CDC funding, it will be handled above on line 333-ish. if this makes sense, this comment should be deleted.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo.tsx
```diff
@@ -64,71 +69,65 @@ const FamilyInfo: Section = {
 		// set up form state
 		const { setAlerts } = useContext(AlertContext);
 		const initialLoad = visitedSections ? !visitedSections[FamilyInfo.key] : false;
-		const [hasAlertedOnError, setHasAlertedOnError] = useState(false);
+		// const [hasAlertedOnError, setHasAlertedOnError] = useState(false);
+		// We're not setting has alerted on error anywhere because there are no server errors here?
```
jlhogan: yes, there are no blocking validations on this object so we will never get a 400 response that we want to use for a field visible on this page. However, I put the code here to make it match the other section forms so it would be clearer to see what shared logic should be extracted when creating a form component. Fine with removing it as it doesn't do anything.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo.tsx
```diff
@@ -64,71 +69,65 @@ const FamilyInfo: Section = {
...
+			FormReducer<DeepNonUndefineable<Enrollment>>
+		>(formReducer, enrollment);
+		const updateFormData = updateData<DeepNonUndefineable<Enrollment>>(updateEnrollment);
 
-		const child = enrollment.child;
 		const { user } = useContext(UserContext);
+
+		if (!_enrollment.child.family) {
+			// If there isn't a family, create one-- otherwise user can save section without making one
+			updateFormData(e => e.value)({
```
jlhogan: again this should be an object that a family util creates

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo.tsx
```diff
@@ -199,9 +201,11 @@ const FamilyInfo: Section = {
 							type="select"
 							id="state"
 							label="State"
+							name="child.family.state"
 							options={['CT', 'MA', 'NY', 'RI'].map(_state => ({ text: _state, value: _state }))}
 							selected={state ? [state] : undefined}
-							onChange={(event, selectedValues) => updateState(selectedValues[0])}
+							// onChange={(event, selectedValues) => updateState(selectedValues[0])}
```
jlhogan: nit: delete

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -230,8 +226,9 @@ const ChildInfo: Section = {
 							type="input"
 							id="sasid"
 							label="SASID"
+							name="child.sasid"
```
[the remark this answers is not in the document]
cailyncodes: In #823 there is a sample of how we might do it

PR #839, opened 2020-03-05
Fix withdrawn bug

mmazanec22:


mmazanec22:
@jlhogan I think maybe the way we're doing this elsewhere isn't working the way we think it is.  If you try to save a new enrollment without entering any child info and console log error on line 212 of ChildInfo, you can see that it's null too, and that what's preventing save there is that res is undefined.

PR #848, opened 2020-03-09
847: Update some application models to use read-only property; get ri…

### LINKED ISSUE #846: controller filter tests

### LINKED ISSUE #847: update read-only entities  --> readonly properties

jlhogan:
…d of read-only entity

closes #847 
closes #846


test/HedwigTests/Data/HedwigContextTests.cs
```diff
@@ -76,97 +76,5 @@ public async Task UpdatedTemporalEntity_IsUpdated_OnSaveChangesAsync()
 				Assert.NotEqual(updatedAt, family.UpdatedAt);
 			}
 		}
-
-		[Fact]
```
cailyncodes: Should we have an equivalent for read only properties?

jlhogan: it doesn't happen at the DB context level; need to write tests for the filter

PR #850, opened 2020-03-09
Update mutate error return

### LINKED ISSUE #834: When you try to save child info without entering any info, then enter just first and last name, can’t save

mmazanec22:
@jlhogan you were right that this would fix #834 -- the error check in the child info promise was getting the old error from the first save attempt.

### DIFF
--- a/src/Hedwig/ClientApp/src/hooks/useApi.ts
+++ b/src/Hedwig/ClientApp/src/hooks/useApi.ts
@@ -17,6 +17,7 @@ export type Mutate<TData> = (
 	query: Query<TData>,
 	reducer?: Reducer<TData | undefined>
 ) => Promise<TData | null>;
+// Null is the return value if the mutation fails
 
 interface ApiParamOpts<T> {
 	defaultValue?: T;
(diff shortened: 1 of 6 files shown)


src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -120,6 +120,7 @@ export default function useApi<TData>(
 				})
 				.catch(async apiError => {
 					await handleError(apiError);
+					return null;
```
jlhogan: can we capture somewhere (doc comment on `mutate` function?) how this works. Think for posterity it is useful to explicitly state somewhere: returns an object if successful, or `null` if an error

but saying that out loud, i realize that might not always work -- for instance, DELETE requests. There's no response body on those, is there?

jlhogan:
to my point above -- we may need to switch back to a world where failed API requests reject a promise

cailyncodes:
Maybe I'm doing something wrong but I can repo #834 off of this PR. It looks like the issue lives in the initialization of the default enrollment object in ChildInfo.

Re the returning null issue, I have a useApi cleanup that I think solves the issue on the page erroring out on first render

cailyncodes:
I found the specific issue -- in `enrollment.ts` in `modules` we are using the `user` object from our test data. As opposed to passing it in the `emptyEnrollment` function

PR #851, opened 2020-03-10
Use API cleanup

cailyncodes:


src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -116,7 +121,7 @@ export default function useApi<TData>(
 					setState(_state => {
 						return { ..._state, loading: false, error: null, data: reducer(data, result) };
 					});
-					return result;
+					return result; // is there a reason we return the result?
```
jlhogan: if we don't return the result, does `data`  in apiState reliably have the correct value? If so, then we would expect `error` to as well. But we don't see that (as evidenced by the thing melanie is trying to work out in the other PR). Like, the way we expected apistate to be available in components does not seem to be true so now we have some half way fixes throughout (like this return).

PR #856, opened 2020-03-11
829 income determination validation

### LINKED ISSUE #829: Income determination must be within the last 365 days, is missing info otherwise
**User story:** As an OEC program manager, I need to ensure that every family's income is determined at least once a year. As a provider, I need to see when a determination is out-of-date and thus no longer valid.

**AC:**
* Create a new validation rule that requires that an enrolled, publicly-funded child must have a income determination within the last 365 days. It's allowed to have older income determinations, as long as the most recent is no more than 365 days old.
* When a publicly-funded child has an income determination that is more than 365 days old, that income determination is treated as though it is empty, and the child is marked as "Missing information" on the roster and on the child summary view.
* But! This should not block submitting a report if determination was valid as of the end of the reporting period. Only if the determination expires prior to the end of the reporting period would the child be marked as "Missing information" for purposes of that report.

jlhogan:
Updates the validationContext to have a dict of all parent objects from the first inital call to `Validate` on down (the validator is scoped, and validate should be called 1x per request, so the context can be expected to contain only the entity chain for the given validation)

This enabled me to move the familyDetermination validations that require enrollment info (i.e. fundings) down to the family determination level.

implemented the AC from #829 with new validation rule `IfEnrollmentFunded_DeterminationDateValid` which checks that the determination date is either:
- less than one year old 
OR
- if the context contains a report entity (indicating the validation is occuring in the context of a report), then is less than one year from end of report reportingPeriod

closes #829


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome.tsx
```diff
@@ -243,7 +248,11 @@ const FamilyIncome: Section = {
 									onChange={updateFormData(newDate => (newDate ? newDate.toDate() : null))}
 									status={initialLoadErrorGuard(
 										initialLoad,
-										warningForField('determintionDate', determination ? determination : null, '')
+										warningForField(
+											'determinationDate',
+											determination ? determination : null,
```
cmgiven: Can `entity` accept undefined so we don't need to map that to null?

src/Hedwig/Validations/Rules/FamilyDetermination/IfEnrollmentFunded_DeterminationDateValid.cs
```diff
@@ -0,0 +1,50 @@
...
+
+			var fundings = enrollment.Fundings ?? _fundings.GetFundingsByEnrollmentId(enrollment.Id);
+			if (fundings.Any(f => f.Source == FundingSource.CDC))
+			{
+				// Generally, DeterminationDate must be < 1 year from today
+				var compareDate = DateTime.Now.Date;
+
+				// However, if validating in the context of report enrollments,
+				// determination date must be < 1 year from end of reporting period
+				var report = context.GetParentEntity<CdcReport>();
```
cmgiven: I like this so much. Very nice solution.

jlhogan:
Should we update some of the `missing information` copy to be `missing or invalid information` ?

Issue #869, opened 2020-03-16
azure build pipeline - specify triggers in YAML for workaround for automated triggers


speedcm:
Microsoft has broken its own product via an update.  As a result, the PR and CI builds are not automatically triggering.  As a workaround, they must be manually started.   To resolve the issue Microsoft recommends to explicitly identify the triggers in the YAML file.  

For more information, see https://developercommunity.visualstudio.com/content/problem/948871/github-is-not-able-to-trigger-ado-pipeline-builds.html

speedcm:
This issue is now resolved.

PR #870, opened 2020-03-16
fix pull request auto-trigger issue (#869)

speedcm:
Microsoft has broken its own product via an update. As a result, the PR and CI builds are not automatically triggering. As a workaround, they must be manually started. To resolve the issue Microsoft recommends to explicitly identify the triggers in the YAML file.

For more information, see https://developercommunity.visualstudio.com/content/problem/948871/github-is-not-able-to-trigger-ado-pipeline-builds.html

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,6 +5,11 @@
 trigger:
   - master
 
+pr:
+  branches:
+    include:
+    - '*'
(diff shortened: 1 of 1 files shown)


speedcm:
This is is ready for review.  The PR build is now automatically started via this fix.

PR #880, opened 2020-03-17
875 validation action filter

### LINKED ISSUE #875: Make validation into an action filter
Wrap validation step in an action filter `OnActionExecuting` / `OnActionExecuted`

- needs to be wrapped in `ServiceFilter` to get DI services
- should also get constructor param with incoming/outgoing flag to signal when to do the validation

when complete, revert https://github.com/ctoec/ecis-experimental/pull/871

jlhogan:
closes #875

### DIFF
--- a/src/Hedwig/Controllers/ReportsController.cs
+++ b/src/Hedwig/Controllers/ReportsController.cs
@@ -8,7 +8,6 @@ using System;
 using Hedwig.Models;
 using Hedwig.Repositories;
 using Hedwig.Security;
-using Hedwig.Validations;
 using Hedwig.Filters;
 
 namespace Hedwig.Controllers
@@ -18,15 +17,12 @@ namespace Hedwig.Controllers
 	[Route("api/organizations/{orgId:int}/[controller]")]
 	public class ReportsController : ControllerBase
 	{
-		private readonly INonBlockingValidator _validator;
 		private readonly IReportRepository _reports;
 
 		public ReportsController(
-			INonBlockingValidator validator,
 			IReportRepository reports
 		)
 		{
-			_validator = validator;
 			_reports = reports;
 		}
 
@@ -34,13 +30,13 @@ namespace Hedwig.Controllers
 		[HttpGet]
 		[ProducesResponseType(StatusCodes.Status200OK)]
 		[ProducesResponseType(StatusCodes.Status204NoContent)]
-		[TransformEntityFilter]
+		[TypeFilter(typeof(ValidateEntityFilterAttribute), Order = 1)]
+		[TransformEntityFilter(Order = 2)]
 		public async Task<ActionResult<List<CdcReport>>> Get(
 			int orgId
 		)
 		{
 			var reports = await _reports.GetReportsForOrganizationAsync(orgId);
-			_validator.Validate(reports);
 			return Ok(reports);
 		}
 
@@ -48,7 +44,8 @@ namespace Hedwig.Controllers
 		[HttpGet("{id:int}")]
 		[ProducesResponseType(StatusCodes.Status200OK)]
 		[ProducesResponseType(StatusCodes.Status404NotFound)]
-		[TransformEntityFilter]
+		[TypeFilter(typeof(ValidateEntityFilterAttribute), Order = 1)]
+		[TransformEntityFilter(Order = 2)]
 		public async Task<ActionResult<CdcReport>> Get(
 			int id,
 			int orgId,
@@ -58,7 +55,6 @@ namespace Hedwig.Controllers
 			var report = await _reports.GetReportForOrganizationAsync(id, orgId, include);
 			if (report == null) return NotFound();
 
-			_validator.Validate(report);
 			return Ok(report);
 		}
 
@@ -66,7 +62,8 @@ namespace Hedwig.Controllers
 		[ProducesResponseType(StatusCodes.Status200OK)]
 		[ProducesResponseType(typeof(ValidationProblemDetails), StatusCodes.Status400BadRequest)]
 		[ProducesResponseType(StatusCodes.Status404NotFound)]
-		[TransformEntityFilter]
+		[TypeFilter(typeof(ValidateEntityFilterAttribute), Arguments = new object[] { true }, Order = 1)]
+		[TransformEntityFilter(Order = 2)]
 		public async Task<ActionResult<CdcReport>> Put(
 			int id,
 			int orgId,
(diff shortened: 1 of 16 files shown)


test/HedwigTests/Filters/ValidateEntityFilterAttributeTests.cs
```diff
@@ -0,0 +1,112 @@
...
+				RouteData = routeData.Object,
+				ActionDescriptor = actionDescriptor.Object
+			};
+
+			var filterMock = new Mock<IList<IFilterMetadata>>();
+
+			// and an ActionExecutingContext exists with the ActionContext
+			// and with an ActionArgument member
+			// with an entry that is an application model
+			var applicationModel = new Hedwig.Models.ApplicationModel();
```
cailyncodes: Is there a reason for using namespace scoping as opposed to using the `Hedwig.Models`?

src/Hedwig/Controllers/ReportsController.cs
```diff
@@ -58,15 +56,15 @@ int orgId
...
-			_validator.Validate(report);
 			return Ok(report);
 		}
 
 		[HttpPut("{id:int}")]
 		[ProducesResponseType(StatusCodes.Status200OK)]
 		[ProducesResponseType(typeof(ValidationProblemDetails), StatusCodes.Status400BadRequest)]
 		[ProducesResponseType(StatusCodes.Status404NotFound)]
-		[TransformEntityFilter]
+		[TypeFilter(typeof(ValidateEntityFilterAttribute), Arguments = new object[] { true }, Order = 1)]
```
cailyncodes: Oh is this setting the `onExecuting` constructor param?

src/Hedwig/Filters/ValidateEntityFilterAttribute.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+
+namespace Hedwig.Filters
+{
+	public class ValidateEntityFilterAttribute : ActionFilterAttribute, IActionFilter
+	{
+		private readonly INonBlockingValidator _validator;
+		private readonly bool _onExecuting;
+
+		public ValidateEntityFilterAttribute(INonBlockingValidator validator, bool onExecuting = false)
```
cailyncodes: I'm curious as to how the instantiation from TypeFilter knows to use DI for validator but use the arguments param for onExecuting ?

src/Hedwig/Filters/ValidateEntityFilterAttribute.cs
```diff
@@ -0,0 +1,54 @@
...
+			}
+		}
+		public override void OnActionExecuted(ActionExecutedContext context)
+		{
+			if (_onExecuting) return;
+
+			var responseEntity = (context.Result as ObjectResult).Value;
+
+			ValidateEntity(responseEntity);
+		}
```
cailyncodes: These feel like very similar methods about validation. However, I don't know the semantic difference

src/Hedwig/Filters/ValidateEntityFilterAttribute.cs
jlhogan: no clue :)

src/Hedwig/Filters/ValidateEntityFilterAttribute.cs
jlhogan: https://docs.microsoft.com/en-us/aspnet/mvc/overview/older-versions-1/controllers-and-routing/understanding-action-filters-cs#the-base-actionfilterattribute-class
```

    OnActionExecuting – This method is called before a controller action is executed.
    OnActionExecuted – This method is called after a controller action is executed.
    OnResultExecuting – This method is called before a controller action result is executed.
    OnResultExecuted – This method is called after a controller action result is executed.
```
don't fully grok the difference between action vs result, but this seems to work

src/Hedwig/Controllers/ReportsController.cs
jlhogan: yes, wasn't sure how to make this clear.. this is just ... how this works in dotnet. didn't want to repeat the same comment every time this filter is applied. not sure how to best clarify this for code readers. v open to suggestions

src/Hedwig/Validations/NonBlockingValidator.cs
```diff
@@ -23,13 +23,15 @@ public NonBlockingValidator(IServiceProvider serviceProvider)
...
-			var rules = _serviceProvider.GetServices<IValidationRule<T>>();
+			var ruleType = typeof(IValidationRule<>).MakeGenericType(entity.GetType());
+			var rules = _serviceProvider.GetServices(ruleType);
 			var errors = new List<ValidationError>();
 
 			_validationContext.AddParentEntity(parentEntity);
 			foreach (var rule in rules)
 			{
-				var error = rule.Execute(entity, _validationContext);
+				var executeFunc = rule.GetType().GetMethod("Execute");
```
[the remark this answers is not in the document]
jlhogan: because the `rules` returned on line 27 are type `IEnumerable<object>` even tho the service lookup ensures they'll be of type `IEnumerable<{ruleType}>`... you cannot cast to a type variable in dotnet

Reflection is necessary to work around the compile time errors generated by attempting to just call `Execute` on the rule which has compile-time type object. 

can add a comment

src/Hedwig/Filters/TypeExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+	public static class TypeExtensions
+	{
+		/// <summary>
+		/// Returns true if given `type` is an enumerable collection of items
+		/// that is not a string.
+		/// </summary>
+		/// <param name="type"></param>
+		/// <returns></returns>
+		public static bool IsNonStringEnumerable(this Type type)
```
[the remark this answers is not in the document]
jlhogan: ah yeah this name is bad and was supposed to change before this ended up in PR. Pretty much, this is not necessarily fully complete, but for our purposes all iterable "collections" either implement ICollection or List. Strings implement IEnumerable, but not ICollection, so this only returns true when the type in question is something like `IEnumerable<T>` or `List<T>` but not string, which implements IEnumerable<char>. Better name (or better way of doing this) welcome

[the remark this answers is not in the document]
cailyncodes: Ahh I gotcha -- okay hmm, what about a string type guard:

```
{
  if (type == typeof(string)) return false;
  return type.IsAssignableFrom(typeof(IEnumerable<>));
}
``` 
?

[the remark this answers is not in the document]
jlhogan: yeah that's pretty much where I ended up! :)

test/HedwigTests/Filters/ValidateEntityFilterAttributeTests.cs
jlhogan: ah this is copy pasta from the transform entity filter tests. I defined a test model class to use in that test file. Gna change up how i do this to make it more clear and to keep the test model out of the real hedwig.Models namespace

PR #882, opened 2020-03-17
841 refactor use api

mmazanec22:
WIP, some outstanding questions, but I think this is in a place where it could use feedback


mmazanec22:
Intentionally did not update snapshots, as we don't intend to keep withdraw-new as the link, and I wanted to make that gets changed before merging

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
```diff
@@ -0,0 +1,113 @@
...
+				const _error = await parseError(apiError);
+				setState({ ...state, data: null, error: _error });
+			});
+	}, [accessToken, skip]);
+
+	useEffect(() => {
+		if (callback) callback();
+	}, [state]);
+
+	// No loading because loading will never be true at this point
```
cailyncodes: Can you elaborate on this? I am only on my first skim so maybe this reveals itself later

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
```diff
@@ -0,0 +1,113 @@
...
+	const [state, setState] = useState<ApiState<TData>>({
+		error: null,
+		data: null,
+	});
+	const { error, data } = state;
+
+	// Run query
+	useEffect(() => {
+		if (skip) {
+			setState({ ...state });
```
cailyncodes: Do we need to reset the state here? Or just return?

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
```diff
@@ -0,0 +1,113 @@
...
+
+	useEffect(() => {
+		if (callback) callback();
+	}, [state]);
+
+	// No loading because loading will never be true at this point
+	return [error, data as DeepNonUndefineable<TData>];
+}
+
+const constructApi: (_accessToken: string | null) => HedwigApi | null = (
```
cailyncodes: Do we need the `_` on `accessToken`?

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
cailyncodes: I'm not sure this is true. On first paint, won't all the effects be registered and then our first return would have loading be true, because no network request will have been made yet. Not to mention any latency in the request?

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
```diff
@@ -0,0 +1,113 @@
...
+		// make API query
+		query(api)
+			.then(apiResult => {
+				setState(_state => ({ ..._state, error: null, data: apiResult }));
+			})
+			.catch(async apiError => {
+				const _error = await parseError(apiError);
+				setState({ ...state, data: null, error: _error });
+			});
+	}, [accessToken, skip]);
```
cailyncodes: I can't remember -- did we end up deciding on using async promises for data fetching (like this does) or to use react effects in its render cycle?

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+	history,
+	match: {
+		params: { siteId, enrollmentId },
+	},
+}: WithdrawalProps) {
+	const { user } = useContext(UserContext);
+	const { setAlerts } = useContext(AlertContext);
+	const { cdcReportingPeriods: reportingPeriods } = useContext(ReportingPeriodContext);
+
+	// Set request params as  a state var so that we only attempt to run queries once we have them
```
cailyncodes: I'm currently on the fence about the request params as a combined state -- will follow up after digging in more

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+		}
+		setRequestParams({
+			// These are used in get and put
+			id: enrollmentId,
+			orgId: getIdForUser(user, 'org'),
+			siteId: siteId,
+		});
+	}, [enrollmentId, user, siteId]);
+
+	const [getError, getData] = useNewUseApi<Enrollment>(
```
cailyncodes: `getError` and `getData` suggest to me these are methods, which I don't think is the case ?

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+
+	// set up PUT request to be triggered on save attempt
+	const [putError, putData] = useNewUseApi<Enrollment>(
+		api =>
+			api.apiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPut({
+				...requestParams,
+				enrollment,
+			}),
+		{
+			skip: !attemptingSave || !requestParams || !enrollmentEndDate,
```
cailyncodes: Is there a reason we are including the `enrollmentEndDate`?

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+			// Only need to mess with this the first time
+			setAttemptedSave(true);
+		}
+
+		// TODO: should we separate this from the fundings logic in case the enrollment state update doesn't finish in time?
+		if (enrollmentEndDate) {
+			// Only do this if we're actually going to hit the server-- otherwise it doesn't get unset when we show the client error
+			setAttemptingSave(true);
+		}
+	};
```
cailyncodes: Confused by this -- Is the concern that a user could supply an end date and hit save but the data would be stale and cause an error for missing info?

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+	// set up convenience derived variables
+	const isMissingInformation = hasValidationErrors(enrollment);
+	const fundings =
+		enrollment && enrollment.fundings
+			? enrollment.fundings
+			: ([] as DeepNonUndefineableArray<Funding>);
+	const cdcFunding = currentCdcFunding(fundings);
+
+	// set up PUT request to be triggered on save attempt
+	const [putError, putData] = useNewUseApi<Enrollment>(
```
cailyncodes: This is a departure from what we discussed on the call but an idea I have for puts/posts would be to return a function here that accepts the new data as a param. So instead of updating an object with `setState` (e.g. enrollment), we pass that object in to this returned function. This would remove the risk of stale values and also allow us to trigger the put/post in the save method while also being able to register the response as a state variable like we do here (e.g. putData)

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
cailyncodes: Nevermind -- I don't think that matters too much there as this is an implementation detail that doesn't get surfaced to the client and also is kinda forced on us by the autogenerated fetch code

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
jlhogan: does this `setState` actually do anything?

jlhogan: does this ever actually change state?

src/Hedwig/ClientApp/src/hooks/newUseApi.ts
```diff
@@ -0,0 +1,113 @@
...
+	// Run query
+	useEffect(() => {
+		if (skip) {
+			setState({ ...state });
+			return;
+		}
+
+		const api = constructApi(accessToken);
+		if (!api) {
+			setState({ ...state, error: { detail: 'API not found' } });
```
jlhogan: should this instead throw an exception? that way apiState.error is scoped to only errors resulting from API responses

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+				...requestParams,
+				enrollment,
+			}),
+		{
+			skip: !attemptingSave || !requestParams || !enrollmentEndDate,
+			callback: () => setAttemptingSave(false),
+		}
+	);
+
+	useEffect(() => {
```
jlhogan: would it be possible to combine this useEffect and the next into a single hook to handle the put response? for my brain that would maybe improve readability and this still feels like a super long file

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
```diff
@@ -0,0 +1,310 @@
...
+			fundings: updatedFundings as DeepNonUndefineableArray<Funding>,
+		});
+
+		if (!attemptedSave) {
+			// Only need to mess with this the first time
+			setAttemptedSave(true);
+		}
+
+		// TODO: should we separate this from the fundings logic in case the enrollment state update doesn't finish in time?
+		if (enrollmentEndDate) {
```
jlhogan: should we add a comment to make it clear (here or on line 126 when we defined the put skip or both) that the reason this validation lives in the client is because we're enforcing a specific kind of PUT request  -- a withdrawal for the enrollment. 

As a reader of this code, I forgot that^ and was like hm why do we have this random client side validation what's going on

mmazanec22: Yeah I wasn't 100% sure of that myself until I was like "wtf why does this update and not error" and compared files.  Wondering if this is the best way to do things since the UX is a bit funky-- like first it just yells at you for not filling out one field, then it's like "jk these are required too."  For now will add a note though.

src/Hedwig/ClientApp/src/containers/Withdrawal/NewWithdrawal.tsx
mmazanec22: The enrollment end date is verified on the client side because it's how we know we're attempting a withdrawal (see Julia's comment).  If we set the attempting save variable here, then it won't get unset, because the request isn't actually hitting the server, and the callback isn't called

PR #883, opened 2020-03-17
Add back in author reference on enrollment

jlhogan:
Brings back author reference on temporal entities, which brings back the enrollment update metadata infomation on enrollment detail view

### DIFF
--- a/src/Hedwig/Filters/TransformEntityFilterAttribute.cs
+++ b/src/Hedwig/Filters/TransformEntityFilterAttribute.cs
@@ -153,7 +153,11 @@ namespace Hedwig.Filters
 				}
 
 				var propValue = prop.GetValue(entity);
-				if (IsApplicationModelType(type))
+
+				// Recursively remove entities of same type as prop if:
+				// - prop is application model type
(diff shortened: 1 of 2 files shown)


src/Hedwig/Filters/TransformEntityFilterAttribute.cs
```diff
@@ -153,7 +153,11 @@ private void UnsetTypeSubEntities(object entity, IEnumerable<Type> entityTypes)
 				}
 
 				var propValue = prop.GetValue(entity);
-				if (IsApplicationModelType(type))
+
+				// Recursively remove entities of same type as prop if:
+				// - prop is application model type
+				// - prop is not read-only (this allows all objects in the tree to retain Author and Reporting Period references)
+				if (IsApplicationModelType(type) && !ReadOnlyAttribute.IsReadOnly(prop))
```
cailyncodes: Does another PR remove the `IsApplicationModelType` helper and replace it with an extension method?

jlhogan: yeah, that's in the validation as filter PR i have out

PR #886, opened 2020-03-17
Switch C4K funding

### LINKED ISSUE #866: c4K certificate data schema changes
**AC**: 
- new model:
```
C4KCertificate {
  int Id;
  Guid ChildId;
  DateTime StartDate;
  DateTime EndDate; // is this necessary ? it will always be 1 year from start date, right?
  int FamilyId; // C4K family Id not to be confused with Family.Id and maybe should be renamed
}
```
- add reference to `Child` model:
```
Child {
...
  List<C4KCertificate> C4KCertificates { get; set; }
...
}
```
- clean up C4K fields on `Funding` model
- clean up C4K validations
- remove C4K from FundingSource enum
- Add data transformation to DB MIgration to turn c4K fundings into c4kCertificates for the appropriate child
  - should we attempt to create multiple records from single fundings that span multiple years? or should we do that in post ? 

- front end changes to map c4k to/from new modeling

cailyncodes:
Backend and model changes, front end restructure

Should close #866


cailyncodes:
I would particularly appreciate a second set of eyes on the custom sql data transform

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -545,12 +559,22 @@ const EnrollmentFunding: Section = {
 								defaultValue={c4kFamilyId ? '' + c4kFamilyId : ''}
 								// TODO: USE REDUCER HERE
 								onChange={event => updateC4kFamilyId(parseInt(event.target.value))}
+								// status={initialLoadErrorGuard(
```
cailyncodes: Should these be blocking or non blocking errors?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -371,19 +374,26 @@ const EnrollmentFunding: Section = {
 					break;
 			}
 
-			updatedFundings = [...updatedFundings].filter(
-				funding => funding.id !== (c4kFunding && c4kFunding.id)
-			);
+			const c4kCertificates: C4KCertificate[] = [
+				...(enrollment.child.c4KCertificates || []),
+			].filter(cert => cert.id !== (c4kFunding && c4kFunding.id));
+			console.log(c4kCertificates, receivesC4k);
```
jlhogan: nit console log

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -371,19 +374,26 @@ const EnrollmentFunding: Section = {
...
-					source: FundingSource.C4K,
-					certificateStartDate: c4kCertificateStartDate ? c4kCertificateStartDate : undefined,
-					familyId: c4kFamilyId,
-				});
-			} else {
-				// do nothing
+				if (c4kFunding) {
+					c4kCertificates.push({
+						...c4kFunding,
+						startDate: c4kCertificateStartDate,
```
jlhogan: truly tiny nit but can startDate, familyCertificateId be in the same order here and in defn on line 389-395?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
jlhogan: this is a product question. Should a user be allowed to add c4k funding without providing the cert startdate & family id, and then they get missing information warnings, OR should the user be blocked from proceeding without these values? 
 cc @victoriasuwardiman  @abbyraskin @cmgiven

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -103,34 +103,15 @@ export default function Roster() {
...
-		legendItems.push({
-			text: fundingSourceDetails[source].legendTextFormatter(
-				fundingSourceDetails[source].fullTitle,
-				enrolledForFunding,
-				capacityForFunding,
-				showPastEnrollments
-			),
+	const legendItems: LegendItem[] = Object.values(legendDisplayDetails).map(
+		({ colorToken, shortTitle, legendTextFormatter, hidden }) => ({
+			text: legendTextFormatter(organization, enrollments, showPastEnrollments),
```
jlhogan: :tada:

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -160,15 +160,14 @@ export default function Withdrawal({
...
+		const c4KFunding = currentC4kCertificate(enrollment);
 		if (c4KFunding) {
-			updatedFundings = [
-				...updatedFundings.filter(funding => funding.id !== c4KFunding.id),
+			c4KCertificates = [
+				...c4KCertificates.filter(cert => cert.id !== c4KFunding.id),
 				{
 					...c4KFunding,
-					certificateEndDate: enrollmentEndDate,
+					endDate: enrollmentEndDate,
```
jlhogan: is this still the correct thing to do here: should we be implicitly updating C4K Certificate end dates when we withdraw? cc @abbyraskin  @victoriasuwardiman

src/Hedwig/Migrations/20200313194854_TransformC4KFundingsToC4KCertificates.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Hedwig.Migrations
+{
+	public partial class TransformC4KFundingsToC4KCertificates : Migration
+	{
+		protected override void Up(MigrationBuilder migrationBuilder)
+		{
+			migrationBuilder.Sql(@"
+                INSERT INTO [C4KCertificate] ([ChildId], [FamilyCertificateId], [StartDate], [EndDate])
```
jlhogan: lgtm!!!!!

src/Hedwig/Models/C4KCertificate.cs
```diff
@@ -0,0 +1,28 @@
...
+		[Required]
+		public Guid ChildId { get; set; }
+
+		[Required]
+		public DateTime? StartDate { get; set; }
+
+		public DateTime? EndDate { get; set; }
+
+		[Required]
+		public int? FamilyCertificateId { get; set; }
```
jlhogan: would like to get product sign off  on variable name cc @abbyraskin  @victoriasuwardiman

src/Hedwig/Repositories/EnrollmentRepository.cs
```diff
@@ -24,6 +24,13 @@ public void UpdateEnrollment(Enrollment enrollment)
 			var newFundings = enrollment.Fundings.AsEnumerable().Cast<IHedwigIdEntity<int>>();
 			var oldFundings = currentFundings.AsEnumerable().Cast<IHedwigIdEntity<int>>();
 			UpdateEnumerableChildObjects(newFundings, oldFundings);
+
+			var currentC4kCertificates = _context.C4KCertificates.AsNoTracking().Where(c => c.ChildId == enrollment.ChildId).ToList();
```
jlhogan: i dont love that we have to do this. works for me for now, but would love to continue to think about improvements to make this more implicit ?

jlhogan:
might be nice to run the transformation on a backup of the data before pushing this out. would that be hard given data security concerns @cmgiven ?

cmgiven:
Maybe we can find a way such that the data doesn't leave our prod VPC? Something like we spin up an instance of a snapshot and connect to it from a hedwig instance running on a developer machine?

jlhogan:
> Maybe we can find a way such that the data doesn't leave our prod VPC? Something like we spin up an instance of a snapshot and connect to it from a hedwig instance running on a developer machine?

I don't think we'd even need to run the SQL via hedwig -- we could take the created SQL scripts and run them against a duplicate of the data. As long as we can create a copy of the SQL server and access it, we should be able to confirm the SQL

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
victoriasuwardiman: I think this is right? The certificate ends/moves to a new program when a child is withdrawn. But I would defer to @abbyraskin

src/Hedwig/Models/C4KCertificate.cs
victoriasuwardiman: This is probably a requirements question; I'm not sure if we were intending to capture Family/Case Number or Certificate Number with each record. If it's the Family/Case Number, that actually lives with the child (or family), so it would be the same for each certificate. However, if that's what we're referring to here, I might name it FamilyCaseId. (I don't recommend requiring Certificate Number). Deferring to @abbyraskin @atreni !

abbyraskin: Yep, it's the Case Number (it used to be called "Family ID" so we kept both in the label initially), no the certificate, which doesn't change when the certificate is renewed.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: Reflagging this @victoriasuwardiman @abbyraskin @cmgiven

src/Hedwig/Models/C4KCertificate.cs
cailyncodes: @jlhogan Thoughts based on this thread of moving C4KCaseNumber (formerly FamilyCertificateId) to the Child model? I don't see any issue with this (does require small ui changes) but want to double check and not sure which one I am leaning towards supporting?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cmgiven: I'd prefer the missing information approach (non-blocking). This is also helpful when migrating data, as we'd often be missing one or both pieces of information. If it's blocking, our only choice to migrate without causing issues is to drop the C4K funding altogether.

src/Hedwig/Models/C4KCertificate.cs
jlhogan: if c4k case number is the same for all certs for a child and does not change, then that makes sense to me! however, I'm not the person who knows if that's true

src/Hedwig/ClientApp/src/components/Legend/Legend.tsx
```diff
@@ -20,16 +21,20 @@ const defaultSymbol = (
 export function Legend({ items }: LegendProps) {
 	return (
 		<div className="grid-row flex-wrap margin-y-2 grid-gap oec-legend">
-			{items.map((item, index) => (
-				<div key={index} className="margin-right-1">
-					<div className={`oec-legend__symbol ${item.symbolClass}`}>
-						{item.symbol || defaultSymbol}
+			{items
+				.filter(item => !item.hidden)
```
mmazanec22: We could also do this with the hidden attribute-- any idea about advantages/disadvantages to either approach?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -64,21 +66,22 @@ const EnrollmentFunding: Section = {
 	name: 'Enrollment and funding',
 	status: ({ enrollment }) =>
 		enrollment &&
-		(sectionHasValidationErrors([enrollment.fundings]) ||
+		(hasValidationErrors(enrollment, ['fundings']) ||
```
jlhogan: it is pretty crazy seeing all of `hasValidationErrors`, `processValidationError`, and `sectionHasValidationErrors`  used here, and not having a clear understanding of what the differences are, or why they need to be different. But, that's a problem for another day https://github.com/ctoec/ecis-experimental/issues/915

src/Hedwig/ClientApp/src/containers/Roster/AgeGroupSection.tsx
```diff
@@ -66,23 +71,24 @@ export default function AgeGroupSection({
...
-				) : (
-					<span className="text-italic text-base">{NO_FUNDING}</span>
-				)}
-			</td>
-		),
+		cell: ({ row }: { row: DeepNonUndefineable<Enrollment> }) => {
+			const fundings = (row.fundings || []).map(funding => ({ ...funding, type: 'CDC' as 'CDC' }));
+			const certificates = (row.child.c4KCertificates || []).map(certificate => ({
+				...certificate,
+				type: 'C4K' as 'C4K',
```
jlhogan: weird

src/Hedwig/ClientApp/src/components/Legend/Legend.tsx
```diff
@@ -20,16 +21,20 @@ const defaultSymbol = (
...
+				.map((item, index) => (
+					<div key={index} className="margin-right-1">
```
mmazanec22: The react docs say we should only use index as the key as a last resort: https://reactjs.org/docs/lists-and-keys.html#keys

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -391,27 +397,33 @@ const EnrollmentFunding: Section = {
 		}, [fundingSelection, cdcReportingPeriod]);
 
 		// *** C4K ***
-		const inputC4kFunding = currentC4kFunding(fundings);
-		const [c4kFunding, updateC4kFunding] = useState<DeepNonUndefineable<Funding>>(
+		const inputC4kFunding = currentC4kCertificate(enrollment);
```
mmazanec22: Maybe we should use the already created c4kFunding variable here?

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -103,31 +102,15 @@ export default function Roster() {
...
-			text: fundingSourceDetails[source].legendTextFormatter(
-				fundingSourceDetails[source].fullTitle,
-				enrolledForFunding,
-				capacityForFunding,
-				showPastEnrollments
+	const legendItems: LegendItem[] = Object.values(legendDisplayDetails).map(
+		({ colorToken, shortTitle, legendTextFormatter, hidden }) => ({
+			text: legendTextFormatter(organization, enrollments, showPastEnrollments),
+			symbol: (
+				<Tag text={shortTitle} color={colorToken} className="position-relative top-neg-2px" />
```
mmazanec22: This is a regression from using the generate funding tag function here

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -219,7 +241,7 @@ export default function Withdrawal({
 					</div>
 					{cdcFunding && (
 						<div className="mobile-lg:grid-col-6">
-							<p>{generateFundingTag(cdcFunding)}</p>
+							<p>{generateFundingTypeTag({ ...cdcFunding, type: 'CDC' })}</p>
```
mmazanec22: I think having 'CDC' here should be unnecessary?

src/Hedwig/ClientApp/src/utils/fundingType.tsx
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return '';
+}
+
+export function generateFundingTypeTag(fundingType: FundingType, index?: any): JSX.Element {
+	const color = fundingType.type ? getDisplayColorForFundingType(fundingType.type) : 'gray-90';
+	let key, text;
+	switch (fundingType.type) {
+		case 'CDC':
+			key = `${fundingType.source}-${fundingType.time}`;
```
mmazanec22: This is a regression: https://github.com/ctoec/ecis-experimental/commit/0d503fdb828617dd4b7c09181a05238a289d295c

src/Hedwig/ClientApp/src/utils/models/funding.ts
```diff
@@ -37,48 +20,17 @@ function dedupeFundings(fundings: Funding[]) {
...
-}
-
 /**
  * Returns true if the funding overlaps with the provided range.
  * If no range is provided, defaults to {now(), now()}
  * @param funding
  * @param range
  */
-export function isCurrentToRange(funding: Funding, range?: DateRange): boolean {
+export function isCurrentFundingToRange(funding: Funding, range?: DateRange): boolean {
```
mmazanec22: Is it necessary to have both this and `isCurrentToRangeCDC` now?

src/Hedwig/ClientApp/src/utils/fundingType.tsx
cailyncodes: Seeing that from the snapshot failures -- I fixed it

PR #904, opened 2020-03-20
Azure e2e test

cailyncodes:


jlhogan:
also, looking at the output here, prob want to limit the containers that we're seeing log output from

PR #905, opened 2020-03-23
Add CDC ReportingPeriodId validation and clean up ui warnings/errors

### LINKED ISSUE #885: make FirstReportingPeriod required for fundings with Source = CDC
**User Story**: As an API client, I should not be able to create a funding record with Source = CDC and FirstReportingPeriod(Id) = null

**AC**:
- A blocking validation (attribute) should be added to enforce FirstReportingPeriodId is not null when Source = CDC
- EnrollmentFunding UI form should be updated to correctly handle this error, applying error styling to the field when the newly created validation error is present in the error response

cailyncodes:
There's a partial rider to clean up for we handle warnings vs errors in the front end.

Should close #885


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -421,7 +424,13 @@ const EnrollmentFunding: Section = {
 						id="enrollment-start-date"
 						status={initialLoadErrorGuard(
 							initialLoad,
-							warningForField('entry', enrollment, 'This information is required for OEC reporting')
+							!!error
```
jlhogan: why don't these fields utilize `displayErrorOrWarning`?

src/Hedwig/ClientApp/src/utils/validations/displayErrorOrWarning.ts
```diff
@@ -0,0 +1,55 @@
+import { Dispatch, SetStateAction } from 'react';
```
jlhogan: <3 this

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: Oops that's because I missed it in the refactor haha

PR #909, opened 2020-03-26
Funding validations

### LINKED ISSUE #852: Validate that a child is not claimed on the same report twice
As a provider, if I have a child who moves from an Infant/Toddler program to a Preschool program, I can not claim them for both slots during the same reporting period.

**AC:**
* A Funding is invalid if it overlaps with another Funding (its LastReportingPeriod is the same or greater than the FirstReportingPeriod of another Funding) for the same Child and same Funding Source.
* A report with an overlapping funding cannot be submitted.
* The latter of two overlapping fundings should show a validation error on the First reporting period field with the message "Cannot claim a child twice in a reporting period."
* The First Reporting Period dropdown menu should not show any ReportingPeriod used by a previous funding, unless that reporting period is already selected (thus showing the validation error).

cailyncodes:
Should close #852


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -117,9 +113,21 @@ const EnrollmentFunding: Section = {
 						{!isPrivatePay && !sourcelessFunding && (
 							<p>
 								First reporting period:{' '}
-								{fundingFirstReportingPeriod
-									? reportingPeriodFormatter(fundingFirstReportingPeriod)
-									: InlineIcon({ icon: 'incomplete' })}
+								{fundingFirstReportingPeriod ? (
```
jlhogan: add comment

src/Hedwig/Validations/Rules/Enrollment/FundingsAreValid.cs
```diff
@@ -22,17 +23,15 @@ public ValidationError Execute(Enrollment enrollment, NonBlockingValidationConte
 			}
 
 			var fundings = enrollment.Fundings ?? _fundings.GetFundingsByEnrollmentId(enrollment.Id);
-			foreach (var funding in fundings)
+			var listFundings = fundings.ToList();
```
jlhogan: should Fundings just be a list on Enrollment ?

src/Hedwig/Validations/Attributes/Enrollment/CDC_Funding_ReportingPeriodsAreValid.cs
```diff
@@ -37,14 +46,23 @@ protected override ValidationResult IsValid(object value, ValidationContext vali
...
 						return new ValidationResult("Last reporting period for CDC funding must start before enrollment ends");
 					}
 				}
 			}
 
+			if (enrollment.Entry.HasValue && earliestFirstReportingPeriod != null
+					&& earliestFirstReportingPeriod.PeriodEnd.Date < enrollment.Entry.Value.Date
+				)
+			{
+				return new ValidationResult("First reporting period for CDC funding must not end before enrollment starts");
```
jlhogan: Should we do this validation for every funding individually? And just check that the start of first reporting period is not before enrollment start, and end of last reporting period is not after enrollment end for each funding?

src/Hedwig/Repositories/FundingRepository.cs
```diff
@@ -19,10 +20,21 @@ public List<Funding> GetFundingsByEnrollmentId(int enrollmentId)
 				.Include(f => f.LastReportingPeriod)
 				.ToList();
 		}
+
+		public List<Funding> GetFundingsByChildId(Guid childId)
+		{
+			return _context.Fundings
```
jlhogan: Make this `AsNoTracking` (to ensure we don't supplement an API response with un-requested values) ((also we should establish at least some naming conventions for the untracked query repo functions for use in validation rules but that can be for another day))

Issue #858, opened 2020-03-12
Fundings are associated with FundingSpaces (part-time/full-time splits)


abbyraskin:
**Ticket 1 of 2**

**User story:** As a provider, I want to enroll a child in a space that might be full-time, part-time, some combination thereof, or even not providing services some weeks of the year.

**Background:** Currently, our schema assumes that a child's funding corresponds to either 52 weeks of full-time care, or 52 weeks of part-time care. We indicate the FundingTime on the Funding, either full- or part-time. We use this to match a given child's funding to a FundingSpace, each of which has a FundingTime, for example, when we calculate the utilization on the Roster legend or on a ReportDetail.

But this is a simplification. In fact, a FundingSpace could have some mix of full-time weeks, part-time weeks, or weeks when services were not provided. And a Funding would not be part- or full-time, but rather associated with a FundingSpace.

There are effectively infinite combinations of part- and full-time making up FundingSpaces. You might have 38 weeks of part-time and 14-weeks of full time. You might have 40 weeks of part-time and 12 weeks of full-time. You might have 40 weeks of full-time (and 12 weeks of no services provided).

This ticket describes a refactor and schema change to create a direct association between a Funding and a FundingSpace, and to store the FundingTime exclusively on the funding space. The existing data should be migrated to the new format. It does not yet implement any features related to part-time/full-time split spaces; the only spaces will be 52 weeks of part-time or 52 weeks of full-time, and the user experience should be unchanged.

**AC:**

* Fundings have a many-to-one relationship to a FundingSpace.
* Fundings no longer have a FundingTime. However, the FundingTime must be used to identify which FundingSpace existing Fundings correspond to (along with the AgeGroup and Site from the Enrollment).
* Existing data can be migrated into this new schema.
* The Roster continues to work as expected, using the FundingTime from the associated FundingSpace instead of an attribute on the Funding model.
    * There is an opportunity to refactor the Roster legends to take advantage of this new association.
* The Enrollment and Funding section continues to work as expected, with the drop-down menu for Funding populated with available FundingSpaces for the Organization and AgeGroup.
* The ReportDetail continues to work as expected.

See #876 for part 2.

jlhogan:
https://github.com/ctoec/ecis-experimental/pull/916 part 1 of part 1, add fundingSpaceId to funding. Tested migration against copy of prod data, validated results:

```
1> select count(*) from funding where fundingspaceid = null                     
2> go                                                                           
                                                                                
[...]

PR #916, opened 2020-04-01
858: Add FundingSpace to Funding

jlhogan:
part one (next pr to remove time & make FE changes coming soon!) 

includes rider fix to avoid null reference exception on 404 responses in action filters 

relates to #858

### DIFF
--- a/src/Hedwig/Models/Funding.cs
+++ b/src/Hedwig/Models/Funding.cs
@@ -20,6 +20,9 @@ namespace Hedwig.Models
 		public int EnrollmentId { get; set; }
 		public Enrollment Enrollment { get; set; }
 
+		public FundingSpace FundingSpace { get; set; }
+		public int? FundingSpaceId { get; set; }
+
 		[JsonConverter(typeof(StringEnumConverter))]
 		public FundingSource? Source { get; set; }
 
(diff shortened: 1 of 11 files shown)


src/Hedwig/Models/Funding.cs
```diff
@@ -20,6 +20,9 @@ public class Funding : TemporalEntity, IHedwigIdEntity<int>, INonBlockingValidat
 		public int EnrollmentId { get; set; }
 		public Enrollment Enrollment { get; set; }
 
+		public FundingSpace FundingSpace { get; set; }
+		public int? FundingSpaceId { get; set; }
+
```
cmgiven, line 25: I'd suggest we also might want to drop the `Time` column and switch over to getting FundingTime from the FundingSpace. Maybe that's an unnecessary intermediate step on the way to the second schema change ticket, but it would effectively test that this refactor is working as expected.

cmgiven: Aaaaaaaaaaand, I missed the note about "part one." Ignore me.

jlhogan:
validated migration against prod db copy: https://github.com/ctoec/ecis-experimental/issues/858#issuecomment-608550010

PR #919, opened 2020-04-03
858 funding remove time

### LINKED ISSUE #858: Fundings are associated with FundingSpaces (part-time/full-time splits)

jlhogan:
part II, closes #858 

NOTE: Validations still needed, but want to make sure the actual requirements are captured before I just go writing them, so going to leave that to a separate ticket/task

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -17,7 +17,12 @@ import {
 	FundingSpace,
 } from '../../../generated';
 import UserContext from '../../../contexts/User/UserContext';
-import { validatePermissions, getIdForUser, getFundingSpaceFor, getTime } from '../../../utils/models';
+import {
+	validatePermissions,
+	getIdForUser,
+	getFundingSpaceFor,
+	getTime,
+} from '../../../utils/models';
 import { DeepNonUndefineable, DeepNonUndefineableArray } from '../../../utils/types';
 import {
 	sectionHasValidationErrors,
@@ -110,8 +115,7 @@ const EnrollmentFunding: Section = {
 								? InlineIcon({ icon: 'incomplete' })
 								: isPrivatePay
 								? 'Private pay'
-								: `CDC - ${prettyFundingTime(getTime(cdcFunding))}`
-							}
+								: `CDC - ${prettyFundingTime(getTime(cdcFunding))}`}
 						</p>
 						{!isPrivatePay && !sourcelessFunding && (
 							<p>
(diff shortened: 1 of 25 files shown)


src/Hedwig/ClientApp/src/utils/models/funding.ts
```diff
@@ -1,9 +1,22 @@
...
+/**
+ * constant for display string when an enrollment has no funding
+ */
+export const NO_FUNDING = 'private pay';
+
+/**
+ * Returns time from associated fundingSpace, or undefined
+ * @param funding
+ */
+export function getTime(funding: Funding | undefined) {
```
cmgiven, line 16: Nit: `getFundingTime` might be clearer.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -361,12 +367,14 @@ const EnrollmentFunding: Section = {
 					break;
 				case FundingType.CDC:
 					// Default to part time if none is selected
-					const time = fundingSelection.time || FundingTime.Part;
 					if (cdcFunding) {
 						updatedFundings.push(
 							updateFunding({
 								currentFunding: cdcFunding,
-								time,
+								fundingSpace: getFundingSpaceFor(fundingSpaces, {
```
mmazanec22, line 374: Since we use this three times, maybe we should shorten this section up by creating a fundingSpace variable before line 370?

PR #925, opened 2020-04-06
876 funding space time allocation

### LINKED ISSUE #876: FundingSpaces support part-time/full-time splits
**Ticket 2 of 2**

**User story:** As a provider, I want to enroll a child in a space that might be full-time, part-time, some combination thereof, or even not providing services some weeks of the year.

**Background:** See #858 for additional context. Once implemented, there is now an association between a Funding and a FundingSpace, and all existing functionality continues to work. However, a FundingSpace is still either part-time or full-time.

This follow-on ticket describes a schema change to support FundingSpaces other than 52 weeks of part-time or 52 weeks of full-time. Provided those are the only two types of FundingSpaces present in the data, however, the user experience should be unchanged.

**AC:**

* A FundingSpace no longer has a FundingTime. Instead, it has a one-to-many association to an entity (FundingSpaceCalendar?) that contains a FundingTime and a number of weeks.
    * _Implementation:_ This is probably another model. I might be tempted to check out if there's a .NET-y way to use JSON blobs, just to avoid the extra join. But my assumption is it's a model.
* Existing data can be migrated to this new schema. A full-time FundingSpace becomes a FundingSpace with 52 full-time weeks, for example.
* Existing functionality on the Roster, Enrollment and Funding section, and Reports is unaffected, provided the data contains 52 weeks of one FundingTime. If data contains FundingSpaces with multiple FundingTimes, each of these components should throw an exception.

jlhogan:
closes #876


src/Hedwig/ClientApp/src/generated/models/FundingTimeAllocation.ts
```diff
@@ -0,0 +1,66 @@
...
+	 * @type {FundingTime}
+	 * @memberof FundingTimeAllocation
+	 */
+	time?: FundingTime;
+	/**
+	 *
+	 * @type {number}
+	 * @memberof FundingTimeAllocation
+	 */
+	weeks?: number;
```
cmgiven, line 35: Is there a way to annotate the model such that these don't appear to be optional?

src/Hedwig/Models/FundingSpace.cs
```diff
@@ -20,11 +20,28 @@ public class FundingSpace : IHedwigIdEntity<int>
...
 
 		[JsonConverter(typeof(StringEnumConverter))]
-		public FundingTime Time { get; set; }
+		public Age AgeGroup { get; set; }
+		public List<Funding> Fundings { get; set; }
+		public List<FundingTimeAllocation> FundingTimeAllocations { get; set; }
+	}
 
+	public class FundingTimeAllocation
+	{
```
cmgiven, line 29: It kinda makes sense, but just flagging that this is AFAIK the first instance where we'll break the one-class-per-file rule.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -351,7 +352,8 @@ const EnrollmentFunding: Section = {
 		useEffect(() => {
 			let updatedFundings: Funding[] = [...fundings]
 				.filter(funding => funding.id !== (sourcelessFunding && sourcelessFunding.id))
-				.filter(funding => funding.id !== (cdcFunding && cdcFunding.id));
+				.filter(funding => funding.id !== (cdcFunding && cdcFunding.id))
+				.filter(funding => funding.id === 0);
```
cailyncodes, line 356: Is this supposed to only take fundings with ID = 0?

src/Hedwig/Migrations/20200406011652_UpdateFundingSpaceRemoveTimea.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Hedwig.Migrations
+{
+	public partial class UpdateFundingSpaceRemoveTimea : Migration
```
cailyncodes, line 5: Super nit: There is a stray 'a' at the end of the class and file name.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
jlhogan: ah nope, good catch. needs to filter _out_ the funding with ID = 0, to handle updates to the funding selection, however this was working for the manual testing I was doing b/c I was creating new enrollments, so they did not have any existing fundings.

kmagida110:
I think I agree with Cailyn here. If we can avoid having JSON in the database when it could be replaced by a limited set in a table (I'm assuming FT/PT for every week) that would likely be better. I honestly don't follow everything that is going on in the app, this might be a better conversation in person either Tuesday or Wednesday to honor heads-down Wednesday. How easy would it be to merge this and then make the change later? It would unblock and give me a chance to get some hands on experience.

jlhogan:
> How easy would it be to merge this and then make the change later? It would unblock and give me a chance to get some hands on experience.

Making the schema change now or later is fine; if you're interested in picking up that task as a learning opportunity then i'm on board with that plan! Otherwise I can update this PR, but the rest of the work will be unblocked either way (separation of app & schema :pray: )

kmagida110:
Let's do the schema change first but can we do it in pair programming so I have a better sense of what I'm looking at?

PR #932, opened 2020-04-08
927 slowdown explosion

### LINKED ISSUE #927: Query slow down
Since `17d65452564290b5bd962b4992fc11f2dfd8ac18` response times for enrollment queries are much slower (> 4x)

This is due to the complexity of additional joins from https://github.com/ctoec/ecis-experimental/blob/master/src/Hedwig/Repositories/EnrollmentRepository.cs#L159

According to docs, multiple includes can greatly reduce query performance, and alternative patterns for loading data are required https://docs.microsoft.com/en-us/ef/core/querying/related-data#eager-loading

cailyncodes:
Should close #927 

Includes a rider of paginating the responses from getting all the enrollments. Even with this fix, the data can add up for larger programs. This drastically increases the time to first paint for slower networks.


src/Hedwig/ClientApp/src/components/Suspend/Suspend.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type SuspendProps = {
+	waitFor: boolean;
+	fallback: React.ReactElement | null;
+};
+
+export const Suspend: React.FC<SuspendProps> = ({ waitFor, fallback, children }) => {
+	if (React.isValidElement(children)) {
+		const props = children.props;
+		console.log(props);
```
jlhogan, line 12: delete

src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -59,23 +65,59 @@ export default function useApi<TData>(
...
-			});
+		//
+		if (paginate) {
+			const paginatedQuery = (start: number, count: number) =>
+				query(api, {
+					start: start,
+					count: count,
+				})
+					.then(apiResult => {
+						// Need to use function syntax for state updates so pending updates aren't overwritten
```
jlhogan, line 84: what does this mean?

src/Hedwig/ClientApp/src/hooks/useApi.ts
```diff
@@ -59,23 +65,59 @@ export default function useApi<TData>(
...
+						if (((apiResult as unknown) as any[]).length === 0) {
+							setState(s => ({ ...s, loading: false }));
+							return;
+						}
+
+						setState(s => ({
+							...s,
+							error: null,
+							data: ([
+								...(((s.data || []) as unknown) as any[]),
```
jlhogan, line 96: is there any way to abstract this to make it more clear that what's happening here is the next page of data is being appended to the apiState data value?

src/Hedwig/ClientApp/src/hooks/useApi.ts
cailyncodes: A few lines down we update with the `setState(s => {...s})` syntax instead of `setState({...state})`. The former version takes whatever the state is now right before update whereas the latter uses the cached value of state. This is necessary b/c state updates are react async.

src/Hedwig/ClientApp/src/hooks/useApi.ts
cailyncodes: yeah it could probably be abstracted

cailyncodes: It is appending new responses to the current data set

src/Hedwig/Filters/TransformEntityFilterAttribute.cs
```diff
@@ -152,22 +155,33 @@ private void UnsetTypeSubEntities(object entity, IEnumerable<Type> entityTypes)
...
 				var propValue = prop.GetValue(entity);
 
-				// Recursively remove entities of same type as prop if prop is application model type
-				if (type.IsApplicationModel())
+				var newList = new List<Type>(typesToRemove);
+				// Add prop type to typesToremove if prop is application model type
+				// And not a read-only property
+				// Read only properties should be included no matter what
+				// e.g. reporting periods or users
+				if (type.IsApplicationModel() && !prop.IsReadOnly())
```
jlhogan, line 179: OK i know we talked thru this and moving this check for read only down here was my suggestion, but i'm rethinking it. he he Say an enrollment has a site on it, and that site is read-only (which it should be). Do we still want to unset any sites further down in the tree, on say `enrollment.site.organization.sites` ?  This doesn't matter for our current use cases, and we can ignore it for now until if or when it ever does. but i just had to say something :)

cailyncodes: Yeah I'm thinking a separate attribute makes the most sense now

PR #934, opened 2020-04-09
933: Quick fix - unblock report submission

jlhogan:
unblocks 933, but leaving that open for a more in depth discussion & solution

### DIFF
--- a/src/Hedwig/Repositories/ReportRepository.cs
+++ b/src/Hedwig/Repositories/ReportRepository.cs
@@ -68,7 +68,7 @@ namespace Hedwig.Repositories
 			var reportResult = await reportQuery.FirstOrDefaultAsync();
 
 			// Optionally manually insert time-versioned enrollment records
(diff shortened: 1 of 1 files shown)


src/Hedwig/Repositories/ReportRepository.cs
```diff
@@ -68,7 +68,7 @@ public async Task<CdcReport> GetReportForOrganizationAsync(int id, int orgId, st
 			var reportResult = await reportQuery.FirstOrDefaultAsync();
 
 			// Optionally manually insert time-versioned enrollment records
-			if (include.Contains(INCLUDE_ENROLLMENTS))
+			if (reportResult != null && include.Contains(INCLUDE_ENROLLMENTS))
```
jlhogan, line 71: rider: get the correct 404 if report with id for org does not exist, instead of 500 due to null reference in `GetEnrollmentsForReport`

PR #935, opened 2020-04-09
927: add tests to assert that running validations does not add refere…

jlhogan:
…nces to mapped objects


jlhogan:
> Looks good -- can we confirm that by removed `asNoTracking` that these tests fail?

yeah they totally did not... and revealed that more than one validation was still using a tracking query to load data. but fixed thosed, and confirmed the tests fail w/out asNoTracking. thanks!

PR #942, opened 2020-04-13
Remove required annotation on c4kcerts

### LINKED ISSUE #933: Reports cannot be submitted
report PUT request returns a 400 because the report GET request does not include C4KCertificates on the enrollment.child objects (see https://github.com/ctoec/ecis-experimental/blob/master/src/Hedwig/Repositories/ReportRepository.cs#L97 and https://github.com/ctoec/ecis-experimental/blob/master/src/Hedwig/Repositories/ReportRepository.cs#L75 -- fundings are included with enrollments by default, and children are optionally included but without checking if C4KCertificates should be included) 

PUT request error response:
```
{"errors":{"Enrollments[0].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[1].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[2].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[3].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[4].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[5].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[6].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[7].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[8].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[9].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[10].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[11].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[12].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[13].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[14].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[15].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[16].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[17].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[18].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[19].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[20].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[21].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[22].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[23].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[24].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[25].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[26].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[27].Child.C4KCertificates":["The C4KCertificates field is required."],"Enrollments[28].Child.C4KCertificates":["The C4KCertificates field is required."]},"type":"https://tools.ietf.org/html/rfc7231#section-6.5.1","title":"One or more validation errors occurred.","status":400,"traceId":"|c43a0a44-4cc08a9e31420707."}
```

from discussion in slack, decision to mark `C4KCertificates` as `required` was to improve API consumer experience:
```
We don't have to check if it's null/undefined and then do || []  everywhere
```
However, this implies that `C4KCertificates` will always be included on the child object, which is not the case -- they are optionally included via the query param `include[]=fundings`, so null or undefined is a valid value for this field.

quick fix: include C4KCertificates on the child object in report query
longer term fix: always include `C4KCertificates` everywhere child object is fetched OR remove `required `annotation (preferred solution)

cailyncodes:
Should close #933

### DIFF
--- a/src/Hedwig/Migrations/20200413181724_UpdateC4KCertificatesOnChild.cs
+++ b/src/Hedwig/Migrations/20200413181724_UpdateC4KCertificatesOnChild.cs
@@ -1,0 +1,17 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Hedwig.Migrations
+{
+	public partial class UpdateC4KCertificatesOnChild : Migration
+	{
+		protected override void Up(MigrationBuilder migrationBuilder)
+		{
+
+		}
+
+		protected override void Down(MigrationBuilder migrationBuilder)
+		{
+
+		}
+	}
+}
--- a/src/Hedwig/ClientApp/src/generated/models/Child.ts
+++ b/src/Hedwig/ClientApp/src/generated/models/Child.ts
@@ -199,7 +199,7 @@ export interface Child {
 	 * @type {Array<C4KCertificate>}
 	 * @memberof Child
 	 */
-	c4KCertificates: Array<C4KCertificate> | null;
+	c4KCertificates?: Array<C4KCertificate> | null;
 	/**
 	 *
 	 * @type {Array<ValidationError>}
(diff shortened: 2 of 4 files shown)


src/Hedwig/Migrations/20200413181724_UpdateC4KCertificatesOnChild.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.EntityFrameworkCore.Migrations;
```
jlhogan, line 1: unnecessary migration

cailyncodes: The generated code is different even though this migration looks empty.

cailyncodes: @jlhogan Just wanted to flag this again if you have more context on whether this migration is necessary vis a vis generate code?

PR #955, opened 2020-04-16
Troubleshoot selenium

mmazanec22:
At this point, each test passes on its own consistently, and the smoke screen tests pass together consistently.  The enrollment tests are flaky.  When you run all of the end to end tests, they fail.  I suspect the cleanup is generally not working as expected.

The findByValue/findByText issue is something I think we should make a note to clean up along with future form work.

Generally, even though Selenium seems to encourage otherwise, I think we should stick to the RTL library philosophy, [which the creator articulated in a blog post](https://kentcdodds.com/blog/introducing-the-react-testing-library):

"So rather than dealing with instances of rendered react components, your tests will work with actual DOM nodes. The utilities this library provides facilitate querying the DOM in the same way the user would. Finding for elements by their label text (just like a user would), finding links and buttons from their text (like a user would). It also exposes a recommended way to find elements by a data-testid as an "escape hatch" for elements where the text content and label do not make sense or is not practical."

### DIFF
--- a/src/Hedwig/ClientApp/package.json
+++ b/src/Hedwig/ClientApp/package.json
@@ -10,7 +10,7 @@
 		"storybook": "start-storybook -p 9009 -s public",
 		"_test": "CI=true react-scripts test --env=jsdom",
 		"test": "yarn _test --testPathIgnorePatterns=e2e",
-		"test:e2e": "yarn _test --testPathPattern=e2e --reporters=default --reporters=jest-junit",
+		"test:e2e": "yarn _test --testPathPattern=e2e --reporters=default --reporters=jest-junit --runInBand --detectOpenHandles",
 		"test:e2e:browserstack": "E2E_GRID_URL='http://hub-cloud.browserstack.com/wd/hub' yarn test:e2e",
 		"test:e2e:firefox": "SELENIUM_BROWSER=firefox yarn test:e2e"
 	},
--- a/src/Hedwig/ClientApp/src/e2e/tests/App.e2e.test.ts
+++ b/src/Hedwig/ClientApp/src/e2e/tests/App.e2e.test.ts
@@ -58,6 +58,6 @@ describe('Smoke screen', () => {
 	});
 });
 
-afterAll(() => {
-	driverHelper.cleanup();
+afterAll(async () => {
+	await driverHelper.cleanup();
 });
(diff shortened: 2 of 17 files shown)


src/Hedwig/ClientApp/package.json
```diff
@@ -10,7 +10,7 @@
 		"storybook": "start-storybook -p 9009 -s public",
 		"_test": "CI=true react-scripts test --env=jsdom",
 		"test": "yarn _test --testPathIgnorePatterns=e2e",
-		"test:e2e": "yarn _test --testPathPattern=e2e --reporters=default --reporters=jest-junit",
+		"test:e2e": "yarn _test --testPathPattern=e2e --reporters=default --reporters=jest-junit --runInBand --detectOpenHandles",
```
cailyncodes, line 13: Looks exciting -- don't know what this does

src/Hedwig/ClientApp/src/e2e/tests/App.e2e.test.ts
```diff
@@ -58,6 +58,6 @@ describe('Smoke screen', () => {
 	});
 });
 
-afterAll(() => {
-	driverHelper.cleanup();
+afterAll(async () => {
```
cailyncodes, line 61: Ooh nice catch

src/Hedwig/ClientApp/src/e2e/tests/Enroll.e2e.test.ts
```diff
@@ -7,6 +7,7 @@ import { beginEnroll, enterChildInfo } from '../utilities/enroll';
 
 // Set time out to 60 seconds
 jest.setTimeout(60 * 1000);
+jest.retryTimes(3);
```
cailyncodes, line 10: :raised_hands:

src/Hedwig/ClientApp/src/e2e/tests/Enroll.e2e.test.ts
```diff
@@ -114,65 +116,83 @@ describe('during an Enroll workflow', () => {
...
 
-			const enrollmentFundingInlineMissingInfo = await queryAllByLocator({
-				css: '.EnrollmentFundingSummary .oec-inline-icon--incomplete',
+			// Enrollment date should have a missing info icon with text (incomplete) for screen readers
+			const enrollmentFundingInlineMissingInfo = await findByLocator({
+				xpath:
+					"//*[text()[contains(.,'Enrollment date:')]]/descendant::span[text()[contains(.,'incomplete')]]",
 			});
-			expect(enrollmentFundingInlineMissingInfo.length).toBe(1);
+			expect(await enrollmentFundingInlineMissingInfo.getAttribute('class')).toMatch('usa-sr-only');
```
cailyncodes, line 178: Is there a reason for this change?

cailyncodes:
Only 3 tests pass in the pipeline -- I know you said they are flaky but just want to check in that the pipeline isn't too much more flaky than local

PR #959, opened 2020-04-20
939: Fixup enrollment GET, POST, PUT

jlhogan:
leaves the validation filter bug for coming report dto PR


src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -29,44 +29,7 @@ ISiteRepository sites
 			_sites = sites;
 		}
 
-		[HttpGet]
```
jlhogan, line 32: this is just reordering the controller functions for easier reading -- putting the org enrollments get at the top

jlhogan:
Legit test failures. Can't deal right now, but will address tomorrow:
```
  X HedwigTests.Validations.Rules.FundingTimelinesAreValidTests.Execute_ReturnsError_IfAnyFundingIsNotValid(f1FirstReportingPeriodStartYear: 2010, f1LastReportingPeriodEndYear: 2011, f2FirstReportingPeriodStartYear: 2012, f2LastReportingPeriodEndYear: 2013, doesError: False) [4ms]
  Error Message:
   Assert.Equal() Failure
Expected: False
Actual:   True
  Stack Trace:
     at HedwigTests.Validations.Rules.FundingTimelinesAreValidTests.Execute_ReturnsError_IfAnyFundingIsNotValid(Nullable`1 f1FirstReportingPeriodStartYear, Nullable`1 f1LastReportingPeriodEndYear, Nullable`1 f2FirstReportingPeriodStartYear, Nullable`1 f2LastReportingPeriodEndYear, Boolean doesError) in /app/test/HedwigTests/Validations/Rules/Funding/FundingTimelinesAreValidTests.cs:line 102
                                                                                                                                                                                                                         [xUnit.net 00:00:04.71]     HedwigTests.Validations.Rules.FundingTimelinesAreValidTests.Execute_ReturnsError_IfAnyFundingIsNotValid(f1FirstReportingPeriodStartYear: 2010, f1LastReportingPeriodEndYear: 2011, f2FirstReportingPeriodStartYear: 2012, f2LastReportingPeriodEndYear: null, doesError: False) [FAIL]
                                                                                                                                                                                                                           X HedwigTests.Validations.Rules.FundingTimelinesAreValidTests.Execute_ReturnsError_IfAnyFundingIsNotValid(f1FirstReportingPeriodStartYear: 2010, f1LastReportingPeriodEndYear: 2011, f2FirstReportingPeriodStartYear: 2012, f2LastReportingPeriodEndYear: null, doesError: False) [< 1ms]
  Error Message:
   Assert.Equal() Failure
Expected: False
Actual:   True
  Stack Trace:
     at HedwigTests.Validations.Rules.FundingTimelinesAreValidTests.Execute_ReturnsError_IfAnyFundingIsNotValid(Nullable`1 f1FirstReportingPeriodStartYear, Nullable`1 f1LastReportingPeriodEndYear, Nullable`1 f2FirstReportingPeriodStartYear, Nullable`1 f2LastReportingPeriodEndYear, Boolean doesError) in /app/test/HedwigTests/Validations/Rules/Funding/FundingTimelinesAreValidTests.cs:line 102
                                                                                                                                                                                                                         
Test Run Failed.
```

src/Hedwig/Validations/Rules/Funding/FundingTimelinesAreValid.cs
```diff
@@ -36,11 +39,16 @@ public ValidationError Execute(Funding funding, NonBlockingValidationContext con
...
+				? funding.FirstReportingPeriod.PeriodStart
+				: _reportingPeriods.GetById(funding.FirstReportingPeriodId.Value).PeriodStart;
+
+			DateTime? lastReportingPeriodEnd = null;
+			if (funding.LastReportingPeriodId.HasValue)
 			{
-				lastReportingPeriod = funding.LastReportingPeriod.PeriodEnd;
+				lastReportingPeriodEnd = funding.LastReportingPeriod != null
+					? funding.LastReportingPeriod.PeriodEnd
+					: _reportingPeriods.GetById(funding.LastReportingPeriodId.Value).PeriodEnd;
```
cailyncodes, line 51: I think these could be failing if `Id` field is null but the `Period` field is set. Probably wouldn't happen from the client but maybe our test only sets the actual object. This is a semantic change from the current code.

jlhogan:
This _works_ but feels like it is likely reinventing some wheels.
however, by using the dto explicitly as a guard, we avoid the bi-directional nav property issue. Without using dto, collection nav properties would be loaded as empty array, and attempting to update child on enrollment would result in enrollment being deleted (`child.enrollments`  = [], so all existing enrollments would be set to deleted state, and then not added back)

src/Hedwig/Repositories/HedwigRepository.cs
```diff
@@ -32,70 +32,215 @@ public Task SaveChangesAsync()
...
+								if (itemDTO != null)
+								{
+									// Update entity state to unchanged to 'attach' the entry
+									trackedItem.State = EntityState.Unchanged;
+									// Then set original and current values
+									trackedItem.OriginalValues.SetValues(dbValues);
+									trackedItem.CurrentValues.SetValues(itemDTO);
+
+									// And recursively update entity navigation properties
+									// UpdateNavigationProperties(trackedItem, dbValues, item, itemDTO);
```
cailyncodes, line 216: Is the recursive element intentionally commented out?

src/Hedwig/Repositories/HedwigRepository.cs
```diff
@@ -32,70 +32,215 @@ public Task SaveChangesAsync()
...
+							var dbValues = trackedItem.GetDatabaseValues();
+
+							// If there is not an original value
+							if (dbValues == null)
+							{
+								// Add principal FK id from parent
+								var parentFkProp = collectionEntry.Metadata.ForeignKey.PrincipalKey.Properties.First();
+								var fkValue = parentFkProp.PropertyInfo.GetValue(collectionEntry.EntityEntry.Entity);
+
+								// To current child navigatioSn entity
```
cailyncodes, line 197: Nit: navigations

PR #960, opened 2020-04-20
Report dtos ii

jlhogan:


src/Hedwig/Models/Reports/CdcReport/CdcReportDTO.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Collections.Generic;
+using Hedwig.Validations;
+
+namespace Hedwig.Models
+{
+	public class CdcReportDTO
+	{
+		public int Id { get; set; }
+		public FundingSource Type { get; set; }
```
cailyncodes, line 10: Is this needed?

src/Hedwig/Models/Reports/OrganizationReport/OrganizationReportSummaryDTO.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Hedwig.Models
+{
+	public class OrganizationReportSummaryDTO
+	{
+		public int Id { get; set; }
+		public FundingSource Type { get; set; }
```
cailyncodes, line 8: Likewise here

test/HedwigTests/Filters/ValidateEntityFilterTests.cs
```diff
@@ -15,64 +15,7 @@ namespace HedwigTests.Filters
...
-			_validator.Verify(v => v.Validate(It.IsAny<INonBlockingValidatableObject>(), null), times);
-		}
-
-		[Theory]
-		[InlineData(false)]
-		[InlineData(true)]
-		public void OnActionExecuted_ValidatesEntity_IfNotOnExecuting(
-			bool onExecuting
-		)
+		public void OnActionExecuted_ValidatesEntity()
```
cailyncodes, line 18: I think this is missing the `[Fact]` annotation

src/Hedwig/Models/Enrollment/EnrollmentProfile.cs
```diff
@@ -8,6 +8,9 @@ public EnrollmentProfile()
 		{
 			CreateMap<Enrollment, EnrollmentSummaryDTO>();
 			CreateMap<EnrollmentSummaryDTO, Enrollment>();
+
+			CreateMap<Enrollment, CdcReportEnrollmentDTO>()
+				.ReverseMap();
```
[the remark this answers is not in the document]
jlhogan: yes and i made that change in open 939 PR :)

src/Hedwig/Models/Reports/OrganizationReport/OrganizationReportSummaryDTO.cs
jlhogan: I imagine this is eventually needed here as we may eventually show multiple types of org reports in the report summary page? but cuold remove for now and only add when needed. (that's the reason for this being "organizationReportSummary" and not "cdcReportSummary", but i could be convinced that that is also not useful)

src/Hedwig/Controllers/ReportsController.cs
```diff
@@ -75,6 +77,7 @@ CdcReport report
 			if (report.Id != id) return BadRequest();
 			if (report.OrganizationId != orgId) return BadRequest();
 
+			_validator.Validate(report);
```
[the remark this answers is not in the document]
jlhogan: at the time i was confused abt how to update the filter setup to get this working, but i will try again :)

src/Hedwig/Models/Reports/OrganizationReport/OrganizationReportSummaryDTO.cs
cailyncodes: My thought is to remove for now (we can still call it OrganizationReportSummary) and add if/when needed. I think it's valuable to keep the complexity down (albeit true that this doesn't add much) until design comps occur for multiple reports use case

jlhogan:
>General idea for consideration: Having an interface like HedwigDTO<T> where T is the class that the DTO is based on. 

I wonder if this simplifies the stuff i've done in other 939 PR by allowing us to only pass around the DTOs? would definitely love to pair on a walkthru!

jlhogan:
> Not super necessary but it took me a little while to figure out what this DTO was actually trying to represent.

rethinking this ^, the entity the DTO represents should be clear from the folder the DTO is located in

PR #962, opened 2020-04-22
Change reporting period dropdown options

### LINKED ISSUE #879: First reporting period dropdown menu doesn't show needed options
We had a provider enroll a child 09/02/2019 and want to start them in the February 2020 reporting period, but this option was not available in the dropdown menu. How can we make it possible to select the correct option?

Could we use submitted reports? As in you wouldn't see any options for reports that are before your last submitted report?

Similarly, it's probably not necessary to go too far into the future. I doubt anyone's planning more than a month in advance.

Could the logic be take all reporting periods between the first one the child is eligible for and the one after the current reporting period, but filter out any periods before or equal to the period of the most recent submitted report?

cailyncodes:
Should close #879

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
@@ -1,8 +1,9 @@
 // Variables used in jest mockes -- must start with `mock`
-import { mockAllFakeEnrollments, mockSite } from '../../../tests/data';
+import { mockAllFakeEnrollments, mockSite, mockReport } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
+	mockApiOrganizationsOrgIdReportsGet,
 } from '../../../hooks/__mocks__/useApi';
 
 // Jest mocks must occur before later imports
@@ -12,6 +13,7 @@ jest.mock('../../../hooks/useApi', () =>
 		apiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet: mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet(
 			mockAllFakeEnrollments
 		),
+		apiOrganizationsOrgIdReportsGet: mockApiOrganizationsOrgIdReportsGet([mockReport]),
 	})
 );
 
@@ -127,7 +129,7 @@ describe('EnrollmentEdit', () => {
 			);
 			const reportingPeriodSelect = getByLabelText('First reporting period');
 			const reportingPeriodOptions = getAllByRole(reportingPeriodSelect, 'option');
-			expect(reportingPeriodOptions.length).toBe(cdcReportingPeriods.length + 1);
+			expect(reportingPeriodOptions.length).toBe(cdcReportingPeriods.length);
 		});
 	});
 
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -22,6 +22,7 @@ import {
 	ApiOrganizationsOrgIdSitesIdGetRequest,
 	C4KCertificate,
 	FundingSpace,
+	ApiOrganizationsOrgIdReportsGetRequest,
 } from '../../../generated';
 import UserContext from '../../../contexts/User/UserContext';
 import {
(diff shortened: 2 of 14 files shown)


src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
```diff
@@ -127,7 +129,7 @@ describe('EnrollmentEdit', () => {
 			);
 			const reportingPeriodSelect = getByLabelText('First reporting period');
 			const reportingPeriodOptions = getAllByRole(reportingPeriodSelect, 'option');
-			expect(reportingPeriodOptions.length).toBe(cdcReportingPeriods.length + 1);
+			expect(reportingPeriodOptions.length).toBe(cdcReportingPeriods.length);
```
mmazanec22, line 132: Previously it was plus one because there's the default "select"-- is that no longer the case?

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
```diff
@@ -78,6 +78,25 @@ export const lastEligibleReportingPeriod = (
...
 
+export const reportingPeriodsBeforeDate = (periods: ReportingPeriod[], date: Date) => {
+	return periods.filter(period => moment(period.periodStart).isBefore(date));
+};
+
+export const reportingPeriodsAfterDate = (periods: ReportingPeriod[], date: Date) => {
+	return periods.filter(period => moment(period.periodStart).isAfter(date));
+};
+
+export const reportingPeriodsBetweenDates = (
```
mmazanec22, line 89: Potentially we could make a more general "between dates" function that takes an accessor function as a parameter

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
```diff
@@ -42,7 +42,7 @@ export const periodSorter = (a: ReportingPeriod, b: ReportingPeriod, inverse: bo
 	return aMomentValue - bMomentValue;
 };
 
-const periodSorterInverse = (a: ReportingPeriod, b: ReportingPeriod) => {
+export const periodSorterInverse = (a: ReportingPeriod, b: ReportingPeriod) => {
```
mmazanec22, line 45: Could we make "inverse" a boolean flag on periodSorter?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -248,42 +265,68 @@ const EnrollmentFunding: Section = {
...
+			let allValidPeriods = periodsAfterEntryAndLastReportingPeriod;
 			// If there is currently a selected value for the reporting period
 			if (cdcReportingPeriod) {
 				// Exclude it so we can safely add it back in without creating a duplicate
-				nextPeriods = [...nextPeriods.filter(period => period.id !== cdcReportingPeriod.id)];
+				allValidPeriods = [
+					...allValidPeriods.filter(period => period.id !== cdcReportingPeriod.id),
+				];
 			}
+			console.log(allValidPeriods);
```
mmazanec22, line 315: Console log

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -211,6 +217,17 @@ const EnrollmentFunding: Section = {
...
+		const reportsParams: ApiOrganizationsOrgIdReportsGetRequest = {
+			orgId: getIdForUser(user, 'org'),
+		};
+		const { data: reports } = useApi(api => api.apiOrganizationsOrgIdReportsGet(reportsParams), {
+			skip: !user,
+		});
+		const submittedReports = (reports || [])
+			.filter(report => !!report.submittedAt)
+			.sort((a, b) => dateSorter(a.submittedAt, b.submittedAt, true));
+		console.log(submittedReports);
```
mmazanec22, line 229: Console log

src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
cailyncodes: We now only show one month in the future, so the select is still there we just are showing one less than the total number of reporting periods.

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
cailyncodes: Yeah that's a good call. This entire reportingPeriod util needs some refactor love.

src/Hedwig/ClientApp/src/utils/models/reportingPeriod.ts
cailyncodes: Will address this in the reportingPeriod refactor -- thinking a similar dateSorter with accessor

PR #967, opened 2020-04-24
Currency formatting in utilization table

### LINKED ISSUE #891: Currency numerals on report vertically align
**Blocked by:** #938 

**User story:** As a provider, I want to be able to easily read the dollar figures on my report, even in complicated cases with many rows.

**AC:**

* Rename the "Reimbursement rate (weekly)" column "Reimbursement rate"
* Set all currency columns to use tabular (monospaced) numerals.
* The reimbursement rate column should be formatted as shown, with a multiplication sign between the rate and the number of weeks, and a parenthetical indicating whether it is the part-time or full-time rate.
* The numbers should align such that the decimal is in the same position. _Implementation: maybe invisible leading zeros?_
* Note that there is a space between the dollar sign and the numerals.
* Note that negative numbers are positioned leftward so that the parenthesis overhangs the start of the cell and the dollar sign is aligned.

The preschool row shown indicates a part-time/full-time split (#896). The split reimbursement rate cell is not implemented by this issue, but maybe be a useful future need to keep in mind.

Screen Shot 2020-03-18 at 10 48 21 AM

cailyncodes:
Should close #891

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
@@ -138,7 +138,7 @@ export default function UtilizationTable(report: CdcReport) {
 				),
 			},
 			{
-				name: <>Reimbursement rate</>,
+				name: 'Reimbursement rate',
 				cell: ({ row }) => {
 					const valueBeforeDecimalPoint = getValueBeforeDecimalPoint(row.rate || 0);
 					updateMaxLengthOfReimbursement(oldMaxLengthOfReimbursementRate => {
@@ -149,21 +149,13 @@ export default function UtilizationTable(report: CdcReport) {
 					const leadingZeros =
 						numberOfLeadingZerosNeeded >= 0 ? '0'.repeat(numberOfLeadingZerosNeeded) : '';
 					return (
-						<td className="font-family-mono">
+						<td className="text-tabular">
 							{row.key !== 'total' && (
 								<>
-									<span>$</span>
-									<span> </span>
+									<span>$ </span>
 									<span style={{ visibility: 'hidden' }}>{leadingZeros}</span>
-									<span>{valueBeforeDecimalPoint}</span>
-									<span>.</span>
-									<span>{getValueAfterDecimalPoint(row.rate || 0)}</span>
-									<span> </span>
-									<span>x</span>
-									<span> </span>
-									<span>{weeksInPeriod}</span>
-									<span> </span>
-									<span>weeks</span>
+									{currencyFormatter(row.rate || 0, true)}
+									<span> &times; {weeksInPeriod} weeks</span>
 								</>
 							)}
 						</td>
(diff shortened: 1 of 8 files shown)


cailyncodes:
Will fix tests soon

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -122,33 +138,78 @@ export default function UtilizationTable(report: CdcReport) {
...
+						return Math.max(valueBeforeDecimalPoint.length, oldMaxLengthOfTotalRate);
+					});
+					const numberOfLeadingZerosNeeded = maxLengthOfTotalRate - valueBeforeDecimalPoint.length;
+					const leadingZeros =
+						numberOfLeadingZerosNeeded >= 0 ? '0'.repeat(numberOfLeadingZerosNeeded) : '';
+					return (
+						<td
+							className={cx({ 'oec-table__cell--strong': row.key === 'total' }, 'font-family-mono')}
+						>
+							<span>$</span>
```
mmazanec22, line 187: Why break these up like this rather than do string concat separately?  This seems like it will be harder to test, also adds lots of DOM elements

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -122,33 +138,78 @@ export default function UtilizationTable(report: CdcReport) {
...
+							{row.key !== 'total' && (
+								<>
+									<span>$</span>
+									<span> </span>
+									<span style={{ visibility: 'hidden' }}>{leadingZeros}</span>
+									<span>{valueBeforeDecimalPoint}</span>
+									<span>.</span>
+									<span>{getValueAfterDecimalPoint(row.rate || 0)}</span>
+									<span> </span>
+									<span>x</span>
```
cmgiven, line 162: This should be a multiplication symbol (`&times;` or U+00D7), not a lowercase x.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -122,33 +138,78 @@ export default function UtilizationTable(report: CdcReport) {
...
+					const valueBeforeDecimalPoint = getValueBeforeDecimalPoint(row.rate || 0);
+					updateMaxLengthOfReimbursement(oldMaxLengthOfReimbursementRate => {
+						return Math.max(valueBeforeDecimalPoint.length, oldMaxLengthOfReimbursementRate);
+					});
+					const numberOfLeadingZerosNeeded =
+						maxLengthOfReimbursementRate - valueBeforeDecimalPoint.length;
+					const leadingZeros =
+						numberOfLeadingZerosNeeded >= 0 ? '0'.repeat(numberOfLeadingZerosNeeded) : '';
+					return (
+						<td className="font-family-mono">
```
cmgiven, line 152: We don't actually want to use font-family-mono, because we don't want the text to be monospaced. We just need the numerals to align. So we want to use tabular-numerals.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -122,33 +138,78 @@ export default function UtilizationTable(report: CdcReport) {
...
-							(row.balance < 0 ? 'oec-table__cell--red' : '')
-						}
+						className={cx(
+							{ 'oec-table__cell--strong': row.key === 'total' },
+							{ 'oec-table__cell--red': row.balance < 0 },
+							'font-family-mono'
+						)}
 					>
-						{row.balance < 0 ? '(' : ''}
+						{row.balance < 0 ? '(' : <>&nbsp;</>}
```
cmgiven, line 212: Alternatively this could be done with relative positioning of the contents of negative cells. This approach is probably fine, but may break if we switch from a fully monospaced font to a proportional font with monospaced numerals.

cailyncodes: I tried first with relative position but it broke the line that goes across for each row. I could take a closer look at ways around that though

cmgiven: Gotcha. Yeah, you might nest a div inside the cell and relative position that, instead of the cell itself?

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
cailyncodes: It's partially an artifact of me trying to do this with relatively/flex positioned spans. I now think I can combine some of them, but will still need to separate the `$ `, the hidden zeros, and the actual value

PR #974, opened 2020-04-24
Fix flaky hosted service test

cailyncodes:


jlhogan:
I'm not against this solution but would it be less error prone to make the `SetupHostedServices` method virtual, then override it in an extension for teststartup ?

cailyncodes:
I see that point and could be easily convinced that is the better approach. My thought this explicitly writing everything out is that it requires intentionality about what we are including in our test env and helps limit a type of bug similar to this by having a code change directly to this file included.

test/HedwigTests/Fixtures/TestStartup.cs
```diff
@@ -11,9 +14,54 @@ public class TestStartup : Startup
...
+			services.ConfigureFilters();
+			services.ConfigureControllers();
+			services.ConfigureSpa();
+			services.ConfigureMapping();
+			services.ConfigureRepositories();
+			services.ConfigureAuthentication(Configuration.GetValue<string>("WingedKeysUri"));
+			services.ConfigureAuthorization();
+			services.ConfigureValidation();
+			services.AddSingleton<IDateTime, SystemDateTime>();
+			services.AddSwaggerGen(c =>
```
cailyncodes, line 31: Also perhaps we should pull this out as a `ConfigureSwagger`

jlhogan:
>My thought this explicitly writing everything out is that it requires intentionality about what we are including in our test env and helps limit a type of bug similar to this by having a code change directly to this file included.

SOLD ! :) extracting into a configure swagger also gets a big +!1 from me

PR #1004, opened 2020-04-27
Truncate site column in roster

### LINKED ISSUE #926: Truncate site column to prevent breaking to multiple lines
**User story:** As a multi-site organization, I typically have lots of kids on my roster. If the site column breaks to multiple columns, my roster becomes huge and difficult to scroll through, and printing consumes extra pages of paper. Instead, I want to ensure that each kid only takes a single row.

**AC:** Truncate the "Site" column on the roster to ensure that it always fits on a single line. If not all characters of a site are visible, use an ellipsis (…) to indicate that fact.

cailyncodes:
Should close #926

### DIFF
--- a/src/Hedwig/ClientApp/src/assets/styles/_general.scss
+++ b/src/Hedwig/ClientApp/src/assets/styles/_general.scss
@@ -12,6 +12,16 @@ h2 {
 	color: color('secondary');
 }
 
+.ellipsis-wrap-text {
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+
+	&--tight {
+		max-width: min-content;
+	}
+}
+
 // *** Alignment ***
 .flex-first-baseline {
 	align-items: first baseline;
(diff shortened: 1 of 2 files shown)


mmazanec22:
Was going to comment on this PR that we should probably take the time to consider the accessibility implications of overflow ellipsis.  I'm having trouble finding many resources about it besides this one: https://www.bennadel.com/blog/3624-reconsidering-text-overflow-ellipsis-as-a-design-smell-and-accessibility-concern.htm

cailyncodes:
I haven't read the article so maybe it addresses this.

The truncating happens with CSS, so the full site name should still be in the DOM, and thus read aloud by a screen reader.

There definitely could be other considerations at play though.

Thanks for calling attention to this.

mmazanec22:
Yes that's true that screen reader users will still hear the full text.  The aspect of WCAG that made me flag this as potentially needing more consideration was [3.1.4, which talks about abbreviations](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=313#abbreviations).

I couldn't really find any guidance about this case though, so IMO we make the note and come back if we learn new information.

PR #1009, opened 2020-04-28
1005: Fixes up withdrawal validations

jlhogan:
Also updates withdrawal form to use `displayErrorOrWarning`


src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -268,26 +269,34 @@ export default function Withdrawal({
...
+													fieldId: 'exit',
+													errorCondition: !!enrollmentEndDate,
+													saveCondition: attemptedSave, 
+													message: 'This information is required for withdrawal'
+												},
+												{
+													fieldId: 'exit',
+													errorCondition: reportingPeriodOptions.length === 0,
+													saveCondition: attemptedSave,
+													message: 'ECE Reporter only contains data for fiscal year 2020 and later. Please do not add children who withdrew prior to July 2019.'
```
jlhogan, line 294: cannot remember exactly what this was about, but i cannot create a situation where the reporting periods etc are not available. going back into 2018 I am able to withdraw enrollments. any thoughts @cmgiven

cmgiven: I don't specifically recall either.

This would be based on the contents of the ReportingPeriods table, though. In prod, I think we only have reporting periods from July 2019 forward.

PR #1010, opened 2020-04-28
Add child to c4kcert model

### LINKED ISSUE #1008: If Care 4 Kids in selected during enrollment without adding details, results in error message on save
Added all enrollment details; selected Care 4 Kids check box, but did not add certificate number or start date. Hit save, and received this error message:
Screen Shot 2020-04-28 at 9 07 49 AM

cailyncodes:
Should close #1008

### DIFF
--- a/src/Hedwig/Models/C4KCertificate/C4KCertificateProfile.cs
+++ b/src/Hedwig/Models/C4KCertificate/C4KCertificateProfile.cs
@@ -1,0 +1,14 @@
+using AutoMapper;
+using System;
+
+namespace Hedwig.Models
+{
+	public class C4KCertificateProfile : Profile
+	{
+		public C4KCertificateProfile()
+		{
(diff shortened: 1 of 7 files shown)


src/Hedwig/Models/C4KCertificate/C4KCertificateProfile.cs
```diff
@@ -0,0 +1,14 @@
+using AutoMapper;
+using System;
+
+namespace Hedwig.Models
+{
+	public class C4KCertificateProfile : Profile
+	{
+		public C4KCertificateProfile()
+		{
+			CreateMap<C4KCertificate, C4KCertificateDTO>()
```
jlhogan, line 10: can we add handling here to truncate datetimes to dates ? This avoids 'updates' to C4kCerts when they're not actually updated

PR #1014, opened 2020-04-28
Include missing useEffect dep to fix contract select

### LINKED ISSUE #995: (Needs verified) Can't select contract space after attempting to save new enrollment
(Needs verified) Can't select contract space after attempting to save the enrollment, then going back and filling in everything except that, then trying to save it

cailyncodes:
Should close #995

Include a missing useEffect dependency
Add more comments
Update reporting period options variable names and ensure we look far enough into the future
Use periodEnd as opposed to periodStart to include the current reporting period as an option

Open question on some code -- see inline comments

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
@@ -313,6 +313,7 @@ const EnrollmentFunding: Section = {
 		const [fundingSource, updateFundingSource] = useState<FundingSource | undefined>(
 			currentCdcFunding ? FundingSource.CDC : undefined
 		);
+
 		const [fundingSpace, updateFundingSpace] = useState<FundingSpace | undefined>(
 			currentCdcFunding && currentCdcFunding.fundingSpace
 				? currentCdcFunding.fundingSpace
@@ -322,6 +323,7 @@ const EnrollmentFunding: Section = {
 		const fundingSpaces = idx(site, _ => _.organization.fundingSpaces) as DeepNonUndefineable<
 			FundingSpace[]
 		>;
+
 		//#region Funding Source Options
 		useEffect(() => {
 			var privatePayOpt = {
@@ -355,6 +357,11 @@ const EnrollmentFunding: Section = {
 				}, []);
 
 			setFundingSourceOpts([privatePayOpt, ...newFundingSourceOpts]);
+			// If there are no funding source options,
+			// forcibly reset the funding source to private pay
+			if (newFundingSourceOpts.length === 0) {
+				updateFundingSource(undefined);
+			}
 		}, [site, _enrollment.ageGroup, enrollment, currentCdcFunding]);
 		//#endregion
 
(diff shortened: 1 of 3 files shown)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -213,52 +213,69 @@ const EnrollmentFunding: Section = {
...
 
-			const oneMonthFromToday = moment()
-				.add(1, 'month')
+			// This should always cover the current reporting period plus the next one
+			const twoMonthsAndAWeekFromToday = moment()
+				.add(2, 'month')
+				.add(1, 'week')
 				.toDate();
-			const reportingPeriodsAfterStartDate = propertyBetweenDates(
+			const reportingPeriodsBetweenStartDateAndNextReportingPeriod = propertyBetweenDates(
```
jlhogan, line 231: why are reporting periods only valid if they are before the next reporting period?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
```diff
@@ -362,20 +381,24 @@ const EnrollmentFunding: Section = {
...
-						!!(
-							funding.fundingSpaceId &&
-							!fundingSpaceOpts.some(opt => opt.value === `${funding.fundingSpaceId}`)
-						)
-				);
+				// filter out current CDC funding (will either be deleted, or updated)
+				.filter(funding => funding.id !== (currentCdcFunding && currentCdcFunding.id));
+			// I don't think we need this commented out code
+			// It could possibily delete old information from out underneath the user
+			// Flagging here so it's easy to see during PR review but will delete/uncomment as decided
```
jlhogan, line 393: I added this specifically to delete information out from under the user. Until we update the enrollment/funding section to handle enrollment updates, if the user switches the enrollment btwn age groups, then the previous fundings associated with funding spaces for different age groups are not valid. The alternative is to alert the user to their error when they try to switch the age group, but since the user currently has no way to access & edit past fundings, that didn't feel like a good solution. Ultimately anything we go with here will be "wrong" in some way. 

also, the validations for funding <->fundingSpace are not actually implemented. I started on them but got stuck trying to figure out how to access the full request body in the blocking validation attribute rule. Need to get back to that! the quick fix is to just implement these rules at the enrollment level, where all information is accessible (need enrollment site to check fundingspace is owned by same org, and need enrollment agegroup to make sure fundingspace agegroup matches)

jlhogan: if we keep this ( which i think we should) should add a comment explaining it what it does. my bad! 
```
// and filter out any fundings associated with funding spaces that are no longer valid for this enrollment
```

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: In the AC for #962:
> Similarly, it's probably not necessary to go too far into the future. I doubt anyone's planning more than a month in advance.

> Could the logic be take all reporting periods between the first one the child is eligible for and the one after the current reporting period, but filter out any periods before or equal to the period of the most recent submitted report?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding.tsx
cailyncodes: Okay will add this back in (and check that it doesn't reintroduce any of the issues flagged by the team in the linked issue)

PR #1020, opened 2020-04-29
895 reports record funding weeks

### LINKED ISSUE #895: Report records part-time/full-time weeks
**Prerequisites:** #858, #876 

**User story:** As a provider, I want to inform OEC when I've used one of my full-time weeks so they will reconcile my funding appropriately.

**AC:**

* When one of an organization's FundingSpaces are split between part-time and full-time, an additional question will appear on the report above the utilization table. If multiple FundingSpaces are split, a question will appear for each.
* Splits that do not add up to 52 weeks are currently not supported. If the report encounters this case, it should throw an exception.
* The title of the new question is "[Age] services were provided [Time]", where Age is the age group of the FundingSpace, and Time is either "full time" or "part time".
    * We use the less common Time for the question. So if a space has 40 weeks of part-time care and 12 weeks of full-time care, the question would use "full time". If the space has 44 weeks of full-time care and 8 weeks of part-time care, we would use "part time." If the weeks are evenly split, default to asking about "full time".
* The question uses a small text input to construct a sentence of the format "[0] of [X] weeks in [Month]. See designs. (Note that weeks remaining is added by #901, and the utilization table is updated by #896.)
* The text input should only accept whole numbers between 0 and the number of weeks in the reporting period.
* The answer should be saved with the report. Although we've only asked about either part-time or full-time, because all splits add up to 52 weeks, we know and should record both, i.e. 1 week of full-time and 4 weeks of part-time.
    * For consistency, we probably also want to record the number of weeks of part- or full-time for the spaces we didn't ask about.
* The data should be updated on defocus, so that the changes are reflected in the utilization table (#896).

## Design

Screen Shot 2020-03-19 at 1 11 28 PM

jlhogan:
should close #895


src/Hedwig/Migrations/20200417134331_UpdateCdcReportAddFundingTImeAllocationUsages.cs
```diff
@@ -0,0 +1,53 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Hedwig.Migrations
+{
+	public partial class UpdateCdcReportAddFundingTImeAllocationUsages : Migration
```
cailyncodes, line 5: This is a super small nit: `TIme` vs `Time`

src/Hedwig/ClientApp/src/components/TextInput/TextInput.tsx
```diff
@@ -76,6 +90,38 @@ export function TextInput({
 				/>
 			);
 			break;
+		case 'inline':
```
mmazanec22, line 93: The name is a little vague to me-- inline here just means that there's before and after content?  IMO layout is better handled by the parent component, and the display-inline-flex can just be added to this as a className

src/Hedwig/Migrations/20200417134331_UpdateCdcReportAddFundingTImeAllocationUsages.cs
jlhogan: changed file names and class names this time ;)

src/Hedwig/ClientApp/src/components/TextInput/TextInput.tsx
jlhogan: @mmazanec22 i could not figure out how to get the label to remain above the input, while aligning the input and text defined outside of the `TextInput` component. Pair ?

src/Hedwig/Repositories/HedwigRepository.cs
```diff
@@ -214,11 +214,23 @@ private bool IdEquals(object a, object b)
...
 
-			if (aIdProp == null || bIdProp == null || aIdProp.PropertyType != bIdProp.PropertyType)
+			// Guard for types without Ids or mismatched Id types
+			// (Should never happen)
+			if (aIdProp == null
+				|| bIdProp == null
+				|| aIdProp.PropertyType != bIdProp.PropertyType
+			)
 			{
 				return false;
```
cailyncodes, line 224: Do we want to throw an exception here instead if this should never happen?

src/Hedwig/Migrations/20200428204040_UpdateFundingTimeAllocationUseagesFK.cs
```diff
@@ -0,0 +1,37 @@
...
+					name: "FK_FundingTimeAllocationUsage_FundingTimeAllocation_FundingTimeAllocationId",
+					table: "FundingTimeAllocationUsage");
+
+			migrationBuilder.AddForeignKey(
+					name: "FK_FundingTimeAllocationUsage_FundingTimeAllocation_FundingTimeAllocationId",
+					table: "FundingTimeAllocationUsage",
+					column: "FundingTimeAllocationId",
+					principalTable: "FundingTimeAllocation",
+					principalColumn: "Id",
+					onDelete: ReferentialAction.Cascade);
```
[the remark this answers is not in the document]
jlhogan: yeah, i forgot to add the code to the context to specify that this delete behavior should be restrict. I then updated the context, and reran migrations. Going to collapse all the migrations in this PR -- they happened ad hoc as I was developing but theres no need to keep that complexity around.

src/Hedwig/Migrations/HedwigContextModelSnapshot.cs
```diff
@@ -270,6 +271,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("int");
 
                     b.Property<int?>("FundingSpaceId")
+                        .IsRequired()
```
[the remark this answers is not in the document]
jlhogan: yes, nullable type + required means we get validations for when this field is missing. If we make is not-nullable + required, then it gets a default value of 0. We could make it not nullable and add a different validation that it has a valid value (aka not default value) but this was easier. Happy to go the other route tho if we feel it's better (it is more correct).

cailyncodes:
I should add I don't think my open questions in the big comment above are blocking but directly affect the ticket I'm working on

src/Hedwig/ClientApp/src/utils/models/fundingSpace.ts
```diff
@@ -57,15 +57,16 @@ function formattedWeeks(includeWeeks: boolean, weeks: number) {
  * @param fundingSpaces
  * @param opts
  */
-export function getFundingSpacesFor(
+export function getFundingSpaces(
```
mmazanec22, line 60: Yay name change

src/Hedwig/ClientApp/src/utils/models/fundingSpace.ts
```diff
@@ -74,7 +75,9 @@ export function getFundingSpacesFor(
 		if (source) {
 			match = match && space.source === source;
 		}
-
+		if (time) {
+			match = match && space.time === time;
```
mmazanec22, line 79: This is so much cleaner than all that other stuff we were doing before

jlhogan:
TLDR:
- timesplit utilization -> time utilization (potentially duped info for full-year allocations, but thats Ok! easier for querying, and enables us to eventually track centers being closed)
- time utilizations continue to live on fundingspace
- time utilizations on the fundingspace will be filtered by reporting period to determine which FY they happened in (necessary to calculate weeks remaining in fiscal year)
    - we discussed duplicating reportingPeriod on to time utilization to make this querying easier; should we instead just leave that info via the report relationship and pull up a fiscal year property onto the utilization record? @cailyncodes

Issue #1022, opened 2020-04-29
Update C4K certificates data so that there is at most one cert with no end date


cailyncodes:
For whatever reason, current prod data with newest migrations applied results in some c4k certificates with no start nor end date. And some children have multiple certificates with no end dates.

AC: Update enrollments with no start nor end date to have an end date of the most recent start date of other certificates

PR #1023, opened 2020-04-29
1021: Updates current c4k cert logic to rely on unset end date

### LINKED ISSUE #1021: Only allow users to edit non-ended C4K certs
Currently, when a user edits an enrollments with a child with multiple c4k certificates the UI arbitrarily selects between the multiple certificates.

As a user, I want to only be able to edit the c4k certificate for a child that does not have an end date.

AC: The enrollment and funding edit form populates with the c4k certificate with no end date. If all certificates have end dates, that is functionally the same as having no c4k certificates.

Assumptions: At most one c4k certificates has no end date.

jlhogan:
should close #1021

### DIFF
--- a/src/Hedwig/ClientApp/src/utils/models/c4kCertificate.ts
+++ b/src/Hedwig/ClientApp/src/utils/models/c4kCertificate.ts
@@ -27,21 +27,20 @@ export function c4kCertificateSorter(a: C4KCertificate, b: C4KCertificate) {
 	return dateSorter(a.startDate, b.startDate);
 }
 
-export function currentC4kCertificate(
+/**
+ * Returns the c4k cert with no end date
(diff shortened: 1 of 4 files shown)


src/Hedwig/ClientApp/src/utils/models/c4kCertificate.ts
```diff
@@ -27,29 +27,28 @@ export function c4kCertificateSorter(a: C4KCertificate, b: C4KCertificate) {
 	return dateSorter(a.startDate, b.startDate);
 }
 
-export function currentC4kCertificate(
+/**
+ * Returns the c4k cert with no end date
+ * (Multiple certs with no end date is an invalid data state)
```
cailyncodes, line 32: Yes, but won't happen until after #1022

PR #1024, opened 2020-04-29
Remove doubly null cert start end date

### LINKED ISSUE #1022: Update C4K certificates data so that there is at most one cert with no end date

cailyncodes:
Should close #1022


src/Hedwig/Migrations/20200327152301_AddC4KCertificateAndRemoveC4KFunding.cs
```diff
@@ -39,30 +39,60 @@ protected override void Up(MigrationBuilder migrationBuilder)
...
-										SELECT [Enrollment].[ChildId], [Funding].[CertificateStartDate], [Funding].[CertificateEndDate]
-										FROM [Enrollment]
-										LEFT JOIN [Funding]
-										ON [Enrollment].[Id] = [Funding].[EnrollmentId]
-										WHERE [Funding].[Source] = 1 /* C4K */;
-						");
+					INSERT INTO [C4KCertificate] ([ChildId], [StartDate], [EndDate])
+							SELECT [Enrollment].[ChildId], [Funding].[CertificateStartDate], [Funding].[CertificateEndDate]
+							FROM [Enrollment]
+							LEFT JOIN [Funding]
```
cailyncodes, line 45: @jlhogan It might be due to this LEFT JOIN here where we are getting extra records with null start and null end

PR #1025, opened 2020-04-30
Only delete null/null certs that are dupes

jlhogan:


src/Hedwig/Migrations/20200327152301_AddC4KCertificateAndRemoveC4KFunding.cs
```diff
@@ -65,11 +65,23 @@ protected override void Up(MigrationBuilder migrationBuilder)
 					ON ProcessedFunding.[ChildId] = [Child].[Id]
 			");
 
-			migrationBuilder.Sql(@"
+			// Delete bad C4K Certificate data:
```
cailyncodes, line 68: We don't know how this state resulted in our data. Providers should not be able to produce this state through the UI, so it is likely a result of the data migration. @kmagida110 Something to check for when writing tests for the migration scripts.

PR #1027, opened 2020-04-30
1026: Update C4KCertificate validations

### LINKED ISSUE #1026: C4K validation errors on Child, but not individual cert
also, c4k start date required should be updated to conditionally required if end date is not null

jlhogan:
- Start date is only required if end date is null
- C4kCertificates from object are used if they exist

should close #1026


src/Hedwig/Validations/Rules/Child/C4KCertificatesAreValid.cs
```diff
@@ -17,7 +17,7 @@ IC4KCertificateRepository certificates
 
 		public ValidationError Execute(Child child, NonBlockingValidationContext context)
 		{
-			var certificates = _certificates.GetC4KCertificatesByChildId(child.Id);
+			var certificates = child.C4KCertificates ?? _certificates.GetC4KCertificatesByChildId(child.Id);
```
cailyncodes, line 20: Nice

PR #1030, opened 2020-05-04
Fix link styling

### LINKED ISSUE #992: Update styling on the multi-site selector
Designs have the text using .usa-prose-intro and the link section and icon are blue with an underline

### LINKED ISSUE #1007: "Update" link styling inconsistent
Inconsistent with link styling used on Roster
Screen Shot 2020-04-28 at 8 47 10 AM

cailyncodes:
Should close #992 and close #1007 

PR URL: https://18.222.174.132:5001


jlhogan:
LGTM, only question is what's the diff btwn `angle arrow down` and `arrow down` ? Seems it might be sufficient to just have one? but make sure that it is colored the same as the text here?

cailyncodes:
They have a different stroke width. The angle down better matches the comp for the site downdown whereas arrow down is bolder and matches for the enroll dropdown (but angle down has a fill hard coded in USWDS, so I had to copy/paste from node_modules)

RyanHansz:
A few tweaks to make this more noticeable, since it seems users were missing this control in testing:
- Font-size for the whole line [# children enrolled across # sites] can be 20px.
- Bold [# sites]
- Remove .5rem margin between the program name and this selector

mmazanec22:
Stroke width is something we can set with CSS-- so if that's the only diff maybe we should use one thing and make a "bold" param?

cailyncodes:
It's an SVG, so I don't think we can because the thinkness/width is set by the shape of the path

PR #1040, opened 2020-05-07
Misc refactor and mild test clean up

cailyncodes:
I'm about to create a ticket that documents findings re testing and touch points between application areas.

### DIFF
--- a/src/Hedwig/Configuration/EnvironmentConfiguration.cs
+++ b/src/Hedwig/Configuration/EnvironmentConfiguration.cs
@@ -5,6 +5,8 @@ namespace Hedwig.Configuration
 {
 	public static class EnvironmentConfiguration
 	{
+		private const string DEVELOPMENT = "Development";
+		private const string PRODUCTION = "Production";
 		public static IConfigurationRoot GetIConfigurationRoot()
 		{
 			return new ConfigurationBuilder()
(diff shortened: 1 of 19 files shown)


src/Hedwig/Configuration/EnvironmentConfiguration.cs
```diff
@@ -19,5 +21,15 @@ public static string GetEnvironmentVariableFromAppSettings(string name, string d
 			return GetIConfigurationRoot()
 				.GetValue<string>(name, defaultValue ?? "Development");
```
jlhogan, line 22: use the const here as well?

test/HedwigTests/Repositories/HedwigRepositoryTests.cs
```diff
@@ -1,56 +1,8 @@
...
-// 			}
-// 		}
-// 	}
-// }
+namespace HedwigTests.Repositories
+{
+	public class HedwigRepositoryTests
+	{
+		// TODO: Add tests for UpdateHedwigIdEntityWithNavigationProperties
+		// Or figure out alternative update approach
```
jlhogan, line 6: lol <3

src/Hedwig/Configuration/EnvironmentConfiguration.cs
cailyncodes: Ooh good call

PR #1043, opened 2020-05-11
Fix address formatter

### LINKED ISSUE #1038: Address on the child summary doesn’t show Address Line 2
When I look at a Child Summary, if there is an Address Line 2 filled out, I should see it in the Summary. 

Currently, when I fill out Address Line 2, I do not see it appear in the Summary. I have to click "Update" to see it.

cailyncodes:
Should close #1038

### DIFF
--- a/src/Hedwig/ClientApp/src/utils/models/family.ts
+++ b/src/Hedwig/ClientApp/src/utils/models/family.ts
@@ -19,13 +19,17 @@ export function addressFormatter(family?: Family) {
 	if (!family) {
 		return [''];
 	}
-	var street = family.addressLine1 ? family.addressLine1 : '';
+	var addressLine1 = family.addressLine1 ? family.addressLine1 : '';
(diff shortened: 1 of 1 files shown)


src/Hedwig/ClientApp/src/utils/models/family.ts
```diff
@@ -19,13 +19,17 @@ export function addressFormatter(family?: Family) {
 	if (!family) {
 		return [''];
 	}
-	var street = family.addressLine1 ? family.addressLine1 : '';
+	var addressLine1 = family.addressLine1 ? family.addressLine1 : '';
```
mmazanec22, line 22: Why var?

jlhogan: my money is on whoever wrote this came from writing a bunch of c# code, which uses `var` keyword :)

jlhogan: wait actually, i wrote this i think and i bet i just was being a JS n00b way back when.

PR #1050, opened 2020-05-12
New form

cailyncodes:


src/Hedwig/ClientApp/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default function Checkbox({
+	id,
+	text,
+	name,
+	onChange,
+	defaultValue,
+	className,
+	disabled,
+}: CheckboxProps) {
+	return (
```
mmazanec22, line 22: Maybe this component can be "solo checkbox" or something, and under the hood it can use the checkbox from the choicelist?  Or the choicelist can use the checkbox from here, and the SoloCheckbox component handles layout consistency issues?

cailyncodes:
With what may have been the most difficult type gymnastics I have ever had to do, I solved the two outstanding typing issues we had.

PR #1096, opened 2020-05-14
Prod guard to not dev

cailyncodes:


cailyncodes:
@jlhogan I added some new code based on @speedcm's comment in slack that proper X509 cers are created in all envs

src/Hedwig/Startup.cs
```diff
@@ -62,10 +62,17 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 				app.UpdateDatabase();
 			}
 
+			// We want to be able to see errors in every non-prod stage
```
jlhogan, line 65: nit :s/stage/env?

cailyncodes: Oops thanks

PR #1098, opened 2020-05-15
Guard UsaSpa and add default handler

cailyncodes:


src/Hedwig/Startup.cs
```diff
@@ -79,52 +84,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
 
-			app.UseSpa(spa =>
-			{
-				spa.Options.SourcePath = "ClientApp";
-
-				// If we are development, that means we aren't serving SPA files
-				// as static, compiled resources. So we need to forward requests a
-				// development server.
-				if (env.IsDevelopment())
+			// UseSpa internally calls UseStaticFiles which only supports GET or OPTIONS
```
jlhogan, line 98: tsk tsk useSpa maintainers!

src/Hedwig/Startup.cs
```diff
@@ -79,52 +84,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
-				}
-			});
 
+					// If we aren't in development, UseSpa will internally call UseStaticFiles
+					// on the directory specified by AddSpaStaticFiles
+				})
+			);
+
+			// Catch any remaining requests and return 404 not found
+			app.Run(
```
jlhogan, line 139: does this `Run()` only run after the above `UseWhen()` ? im confused about the request flow here. what happens to the PUT/POST/DELETES?

cailyncodes: GET/HEAD will fall through the UseSpa above.

PUT/POST/DELETE/OPTIONS will not match the UseWhen so will continue to the app.Run

jlhogan: got it! forgot abt the endpoint delegates set up by the controllers.

src/Hedwig/Startup.cs
```diff
@@ -79,52 +84,64 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+							// script and provided the referenced executable exists on the path
+							// the script will run.
+							spa.UseReactDevelopmentServer(npmScript: "start");
+						}
 					}
-				}
-			});
 
+					// If we aren't in development, UseSpa will internally call UseStaticFiles
+					// on the directory specified by AddSpaStaticFiles
```
jlhogan, line 134: does setting `spa.Options.SourcePath` implicitly do the same thing as `AddSpaStaticFiles`?

cailyncodes: Nope -- that tells the UseReactServer where the code lives to serve

cailyncodes: UseSpa does the same logic as AddSpaStaticFiles under the scenes

cailyncodes: AddSpaStaticFiles is necessary because it updates DI container for use in UseSpa

jlhogan: got confused b/c `UseSpaStaticFiles` went away, but `AddSpaStaticFiles` didn't show up. see now it's called in service extension method.

PR #1100, opened 2020-05-15
Add back UseSpaStaticFiles

cailyncodes:


src/Hedwig/Startup.cs
```diff
@@ -92,12 +92,23 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 				endpoints.MapControllers();
 			});
 
+			// If we are not in development, try serving static files according to
+			// the path specified in AddSpaStaticFiles
```
jlhogan, line 96: maybe call out that `AddSpaStaticFiles` is called in configureServices?

cailyncodes: Sounds good

PR #1101, opened 2020-05-18
Add validations for funding's fundingspace

jlhogan:


src/Hedwig/Validations/Attributes/Enrollment/Fundings_FundingSpacesAreValid.cs
```diff
@@ -0,0 +1,61 @@
+using System.ComponentModel.DataAnnotations;
```
cailyncodes, line 1: Should the various checks be separated into different validation attributes?

src/Hedwig/Validations/Attributes/Enrollment/Fundings_FundingSpacesAreValid.cs
```diff
@@ -0,0 +1,61 @@
...
+			{
+				if (funding.FundingSpace == null)
+				{
+					continue;
+				}
+
+				var ageGroupResult = AgeGroupMatches(enrollment, funding.FundingSpace);
+				if (ageGroupResult != null) return ageGroupResult;
+
+				var siteRepo = validationContext.GetService(typeof(ISiteRepository)) as ISiteRepository;
```
cailyncodes, line 26: Probably should be pulled outside of the for-loop.

test/HedwigTests/Validations/Attributes/Enrollment/Fundings_FundingSpacesAreValidTests.cs
```diff
@@ -0,0 +1,84 @@
...
+	public class Fundings_FundingSpacesAreValidTests
+	{
+
+		[Theory]
+		[InlineData(Age.Preschool, Age.Preschool, false)]
+		[InlineData(Age.Preschool, Age.InfantToddler, true)]
+		public void ReturnsValidationResult_IfFundingSpaceAgeGroupDoesNotMatch(
+			Age fundingSpaceAge,
+			Age enrollmentAge,
+			bool returnsResult
```
cailyncodes, line 18: Nit: `returnsResult` should probably be renamed to indicate a check for null is happening with the result

test/HedwigTests/Validations/Attributes/Funding/FundingSpaceSourceMatchesTests.cs
```diff
@@ -0,0 +1,15 @@
...
+using Hedwig.Validations.Attributes;
+using Xunit;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace HedwigTests.Validations.Attributes
+{
+	public class FundingSpaceSourceMatchesTests
+	{
+		// TODO: Update this test when multiple FundingSource types exist
```
cailyncodes, line 11: Can we create a corresponding github ticket for todos in source code?

src/Hedwig/Validations/Attributes/Enrollment/Fundings_FundingSpacesAreValid.cs
cailyncodes: Or at least organization matches and age group matches? I feel like this class is doing several related but distinct things that will be harder to maintain down the road.

PR #1110, opened 2020-05-19
Fix validation bug

### LINKED ISSUE #1085: Can't see/update missing info on child record
In staging, Alan Rickman is labeled as having missing info. When I go to update his info, nothing appears to be missing.
Screen Shot 2020-05-13 at 10 44 12 AM

cailyncodes:
Rider: Re-imagine validation utils

Should close #1085


src/Hedwig/ClientApp/src/utils/validations/displayValidationStatus.ts
```diff
@@ -0,0 +1,127 @@
+import { FormStatusProps } from '../../components';
+import { ApiError } from '../../hooks/useApi';
+import { ValidationError } from '../../generated';
+import { elementIdFormatter } from '../stringFormatters';
+import { isBlockingValidationError } from './isBlockingValidationError';
+
+type ValidationResponse = ApiError | ValidationError[];
```
jlhogan, line 7: comments here to be explicit about what we're smushing together here; ValidationProblemDetails (from api error, unfortunate naming due to .NET standards + swagger) and ValidationError (from api success response cast to data object)... just because the naming is so close and thus confusing, i think it's worth calling out.

jlhogan:
also could you call out what the bug + the fix was? so much diff and only one commit hard to find in the refactor haystack :)

cailyncodes:
@jlhogan Go ahead and push up the comments change.

The bug fix was resolved by listing out all fields in ChildInfo to check for validation, instead of inspecting the entire child object. That included the c4kcertificate stuff which was causing a missing info icon to show up in summary.

jlhogan:
ok will do. one other question that somehow didn't get posted yesterday: What happens to the `hasAlertedOnError/setHasAlertedOnError` for displaying catchall error state? we used to pass down the set state func to enable the form status determination delegate (aka the utils) to report back up if the catchall error needed to be shown or not. does this still happen?

cailyncodes:
My understanding as I was walking through this code is that it was only ever triggered on based on the result of `error` in the section container. So, I got rid of it, and just had a check on initial load in the error useEffect. As in, the error message would show iff the GET request returned a non-200 status code.

PR #1111, opened 2020-05-20
Fix typings

cailyncodes:
Adds the missing types that got skipped in a PR merge


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.tsx
```diff
@@ -38,6 +45,8 @@ import {
 	REQUIRED_FOR_ENROLLMENT,
 	REQUIRED_FOR_OEC_REPORTING,
 } from '../../../utils/validations/messageStrings';
+import FormNew from '../../../components/Form_New/Form';
```
jlhogan, line 48: nit; these shouldn't be here

PR #1114, opened 2020-05-20
Add date native event handler

cailyncodes:
Previous onChange instances were changed to onChange_Old and the new system is just onChange now.

### DIFF
--- a/src/Hedwig/ClientApp/src/components/DateInput/DateInput.tsx
+++ b/src/Hedwig/ClientApp/src/components/DateInput/DateInput.tsx
@@ -8,7 +8,7 @@ import { parseStringDateInput } from '../../utils/stringFormatters';
 import { InputField } from '../../utils/forms/form';
 
 export type DateInputProps = {
-	onChange_Old: (newDate: InputField<Moment | null>) => any | void;
+	onChange_Old?: (newDate: InputField<Moment | null>) => any | void;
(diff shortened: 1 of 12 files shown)


jlhogan:
you've got fix typings commit in here, but shouldn't be

cailyncodes:
Note: Make `onChange_Old` optional in the type and add `inputOnChange_Old && ` to line 74

src/Hedwig/ClientApp/src/components/DateInput/DateInput.tsx
```diff
@@ -60,7 +71,16 @@ export const DateInput: React.FC<DateInputProps> = ({
 			// Only call the callback if it's a valid date
 			// Spread operator will not copy prototype
 			// https://dmitripavlutin.com/object-rest-spread-properties-javascript/
-			inputOnChange(Object.assign(moment(), input, { name: name || '' }));
+			inputOnChange_Old(Object.assign(moment(), input, { name: name || '' }));
```
jlhogan, line 74: outstanding task? `Note: Make onChange_Old optional in the type and add inputOnChange_Old && to line 74`

cailyncodes: Yeah that was a note to myself because I would have forgotten it -- I needed to do that to make using it with the new form work

PR #1115, opened 2020-05-20
896 utilization table

mmazanec22:


src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -1,57 +1,52 @@
...
+import { prettyAge, prettyFundingTime, fundingSpaceSorter } from '../../../utils/models';
 import currencyFormatter from '../../../utils/currencyFormatter';
 import cx from 'classnames';
 import { DeepNonUndefineableArray } from '../../../utils/types';
+import {
+	countFundedEnrollments,
+	calculateRate,
+	makePrefixerFunc,
+	ReimbursementRateLine,
+} from '../../../utils/utilizationTable';
```
jlhogan, line 22: :+1:

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -77,39 +72,74 @@ export default function UtilizationTable(report: CdcReport) {
...
+			}
+		} else if (fundingTime === FundingTime.Full) {
+			fullWeeks = weeksInPeriod;
+		} else if (fundingTime === FundingTime.Part) {
+			partWeeks = weeksInPeriod;
+		}
+
+		const cappedCount = Math.min(count, capacity);
+		const total = cappedCount * (fullWeeks * ftRate) + cappedCount * (partWeeks * ptRate);
+		const paid = capacity * (fullWeeks * ftRate) + capacity * partWeeks * ptRate;
```
jlhogan, line 117: add final () for parallelism

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -77,39 +72,74 @@ export default function UtilizationTable(report: CdcReport) {
...
+			ptRate,
+			total,
+			balance,
+			showFundingTime: cdcFundingSpaces.filter(fs => fs.ageGroup === ageGroup).length > 1,
+			weeksSplit: {
+				partWeeks,
+				fullWeeks,
+			},
+			maxes: {
+				// TODO: make this the greater of the two vals if there are two time vals
```
jlhogan, line 136: not sure what this means.

jlhogan: edit: ok got it. can we add a comment somewhere (maybe on the interface?) to explain why we need maxes? if i understand, it is to determine the appropriate amount of spacing to prefix all the stuffs with ?

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -77,39 +72,74 @@ export default function UtilizationTable(report: CdcReport) {
...
-				fundingTime,
-				count,
-				capacity,
-				rate,
+		// TODO: probably a util
+		let fullWeeks = 0,
+			partWeeks = 0;
+		if (fundingTime === FundingTime.Split) {
+			const { timeSplitUtilizations } = space;
+			const thisUtilization = (timeSplitUtilizations || []).find(u => u.reportId === report.id);
```
jlhogan, line 104: these values will  not reflect the current state as the defaults and user input are added to the report, not the funding space. so either:
- pass in reportingPeriodId and grab the utilization off the report, filtering by reportingPeriodId
- pass in the state variable from reportSubmitForm that contains the utilization values for this report (either default values or user-defined)

src/Hedwig/ClientApp/src/utils/utilizationTable.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
```
jlhogan, line 1: :clap:

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -184,14 +184,14 @@ export default function ReportSubmitForm({
 		report.timeSplitUtilizations
```
jlhogan, line 184: small remaining bug: when the page first settles, the value of `timeSplitUtilizations` is `[]` and so the row for the splittime fundingspace in the ut table is empty. any blur/change for the utilization question triggers a value to be generated and the row to appear. was also able to achieve this by adding to line 196: 
```
	useEffect(() => {
		// recreate defaults when report (and thus available fundingspaces) changes
		updateTimeSplitUtilizations(
			(report.timeSplitUtilizations && report.timeSplitUtilizations.length) ? report.timeSplitUtilizations : 
			splitTimeFundingSpaces.map(fundingSpace =>
				getSplitUtilization(
					fundingSpace.timeSplit,
					0,
					reportingPeriodWeeks,
					report.reportingPeriod,
					report
				)
			)
		)
	}, [report])
```
but not sure this is the right solution

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
```diff
@@ -37,13 +38,14 @@ interface UtilizationTableRow {
...
 	maxes: {
+		// These are needed to format the numbers for alignment
 		total: number;
 		rate: number;
 		balance: number;
 	};
 }
 
-export default function UtilizationTable(report: CdcReport) {
+export default function UtilizationTable({ report, timeSplitUtilizations }: { report: CdcReport, timeSplitUtilizations: FundingTimeSplitUtilization[] }) {
```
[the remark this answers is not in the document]
mmazanec22: This snippet isn't valid javascript-- I'm seeing this pattern where we are using named exports though:

```
export const UtilizationTable: React.FC<{ report: CdcReport, timeSplitUtilizations: FundingTimeSplitUtilization[] }> = ({
	report,
	timeSplitUtilizations,
}) => {
```

Are we moving toward named exports?  Or was the spirit of this comment more that this is looking weird because I was lazy and used the type inline rather than extracting it out?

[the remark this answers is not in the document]
jlhogan: ha my bad! I was just passing on a comment @cailyncodes made on my last PR that we should explicitly type components as `React.FC<{...}>` but i'll allow her to specify

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
mmazanec22: How about we just grab them off of the report until there are user supplied time split utilizations?

jlhogan: they don't exist on the report until after the post request. The default values (not really user-defined, but a valid state of the data that the user could submit) need to make their way into the table

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/UtilizationTable.tsx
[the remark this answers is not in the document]
cailyncodes: Yeah I lean towards using an explicit type when supplying it to `React.FC<...>`.

Re: the default/named export -- you can get around it by:
```
const X: React.FC<XProps> = <something>;
export default X;
```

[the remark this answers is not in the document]
mmazanec22: Yes I am familiar with default exports.  My comment was about where I am seeing that pattern in use in our code base, versus where we are not using it, since we seem to be a little inconsistent.  If we have agreed to standardize on that particular pattern I will go ahead and implement it.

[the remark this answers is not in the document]
jlhogan: i dont know that we've agreed on something -- i personally have components out there all over the place. dunno that right now is the moment to do a full press on standardizing but we _could_ just make a call and at least conform from here on out? 

sorry for any confusion! was just trying to be consistent. I think i misunderstood the original intent of cailyn's point which is to create types if you are passing them in to `React.FC<TypeProps>`  (as opposed to using anonymous type like `React.FC<{id: number, prop: string}>`). I had interpreted it as: define all functional components with `React.FC` type explicitly

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
jlhogan: (note: Im confused that this is the case -- i would expect `fundingSpaceUtilizations` to get the correct value whether or not the user interacts with the weeks used input, otherwise the defaults don't actually work which i am pretty sure i tested for when first implementing this feature. Soooo, happy to pair on this if that's helpful, because it seems mysterious to me and maybe is a bug)

PR #1117, opened 2020-05-21
FormContext gets ObjectDriller instance

jlhogan:
Also, change ObjectDriller constructor to use data object,
instead of copy.

### DIFF
--- a/src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
+++ b/src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
@@ -1,5 +1,6 @@
 import React, { FormHTMLAttributes, PropsWithChildren, useState, useEffect } from 'react';
 import { FormProvider } from './FormContext';
+import { ObjectDriller } from './ObjectDriller';
 
 type FormProps<T> = {
 	onSubmit: (_: T) => void;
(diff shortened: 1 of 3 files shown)


src/Hedwig/ClientApp/src/components/Form_New/ObjectDriller.ts
```diff
@@ -22,7 +20,7 @@ export class ObjectDriller<T> {
 	path: string;
 
 	constructor(obj: T, path?: string) {
-		this.value = cloneDeep(obj);
+		this.value = obj;
```
cailyncodes, line 23: Do we want to be using the same instance everywhere?

src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
```diff
@@ -43,10 +44,17 @@ const Form = <T extends any>({
 		onSubmit(_data);
 	};
 
+	/**
+	 * dataDriller instance to pass around for convenience,
```
mmazanec22, line 48: Yay clear comment!

src/Hedwig/ClientApp/src/components/Form_New/ObjectDriller.ts
cailyncodes: I guess this is fine just by itself here, but with the above change of adding it to the context

jlhogan: I believe we do, otherwise we'd need to be re-creating all driller instances whenever `obj` changes to have it reflecting up-to-date form data

PR #1126, opened 2020-05-23
Clean up some react warnings and errors

cailyncodes:
Addresses some easy wins for clean up.

Most of the warnings in the enrollment sections are left so as to not conflict with the refactor.

Some other warnings are left because they may require larger overhead to fix, namely most of the remaining useEffects dependencies issues.


src/Hedwig/ClientApp/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -242,11 +242,11 @@ export const ChoiceList: React.FC<ChoiceListProps> = ({
 				{options.map(option => {
 					const expansion = option.expansion;
 					return (
-						<>
+						<React.Fragment key={`${option.value}-expansion`}>
```
jlhogan, line 245: :raised_hands:

PR #1128, opened 2020-05-27
Extract updates to field group types from cailyn's ongoing 1102 work

jlhogan:


src/Hedwig/ClientApp/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -107,3 +111,35 @@ export const CheckboxGroup: React.FC<CheckboxGroupProps & FieldSetProps> = ({
...
+	hint,
+	horizontal,
+	disabled,
+	...props
+}: PropsWithChildren<CheckboxGroupForFormProps<T>>) => {
+	return (
+		<FormFieldSet<T>
+			id={`${id}-fieldset`}
+			className={className}
+			legend={legend || ''}
```
mmazanec22, line 133: We should not default to an empty string for legend

cailyncodes: It's somewhat misleading because that `|| ''` isn't actually doing anything. Should definitely be removed, but the legend prop is coming from FieldSetProps which already enforces it not being empty.

PR #1129, opened 2020-05-28
Update typescript, prettier and refactor useApi for better testing

cailyncodes:
Previously, the `ChildInfo` tests were manually setting the error prop on the `<ChildInfo.Form />` component. The pending `ChildInfo` rewrite does not use the error prop, instead processing errors directly from the `useApi` hook in the component. Following the previous pattern, I tried mocking `useApi` with a Jest mock factory (see any of the container test files for example) by returning an error. This didn't work. After investigating, I saw that our mock implementation of `useApi` would only return the supplied value on the first render and all subsequent renders would return a default `{ error: null, loading: false, data: null }`. This works for our current tests that either don't cause multiple render or don't utilize supplied mock values different from the default. In the new `ChildInfo` component, `attemptSave` is set to `true` when the request should be fired and then immediately set to false. The act of setting it to false (which is a state update) causes a render. In the real `useApi`, we are able to maintain state and only update the `loading` return value, leaving `error` and `data` unchanged. However, in the mock, we have to return the default object which overrides `error` and `data`. This clears `error` so then the `ChildInfo` component renders without an error, causing the tests to fail.

The solution (which also better conforms to testing best practices) is to mock the API call, not the entire hook. The API itself is the generated `HedwigApi` class. We instantiate an instance of it in the `useApi` hook. In order to mock this separately, we need to effectively simulate dependency injection and have the construction (or at least factory function) of the object occur outside of the hook. That is done here by refactoring `useApi` into a set of related files. The `constructApi` function is placed into its own module which allows jest to mock it. Then `useApi` in our tests will use the mocked API but the real `useApi` hook logic, which allow us to keep `error` and `data` state across renders but bypass making network requests. We also need to mock `parseError` because we only supply the actual error, instead of a HTTP response which resolves to an error.

In order to make the separate modules and keep the internal functions unexposed to callers of `useApi`, we need to export just the types from those internal files. This exporting of types only was introduced in TypeScript 3.8. Previously, we were using 3.7. This updates to the current version which is 3.9. The export type only syntax is also not supported by our current version of prettier. So that is also updated to the current 2.0 version. (This also caused a lot of files to be updated with just syntax changes).

This refactor does not require any changes to the current `ChildInfo` implementation. I propose that as we continue to re-write components that have tests using `useApi` mocks, we update the tests to use this new approach. Comments for context are included in `ChildInfo.test.tsx`.

This also adds a `@deprecated` annotation to the default export of `__mock__/useApi`.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -32,6 +32,7 @@ services:
       - winged-keys
   client:
     image: node:12
+    stdin_open: true
     volumes:
       - ./src/Hedwig/ClientApp:/home/node/app:cached
       - ./entrypoint-client.sh:/entrypoint.sh
--- a/src/Hedwig/ClientApp/src/components/ButtonWithDropdown/ButtonWithDrowdown.tsx
+++ b/src/Hedwig/ClientApp/src/components/ButtonWithDropdown/ButtonWithDrowdown.tsx
@@ -55,7 +55,7 @@ const ButtonWithDrowdown: React.FC<ButtonWithDrowdownProps> = ({
 						{dropdownElement}
 					</>
 				}
-				onClick={() => setIsComponentVisible(hide => !hide)}
+				onClick={() => setIsComponentVisible((hide) => !hide)}
 			/>
 			<div
 				className={cx(
@@ -64,11 +64,11 @@ const ButtonWithDrowdown: React.FC<ButtonWithDrowdownProps> = ({
 					dropdown
 				)}
 			>
-				{options.map(_option => (
+				{options.map((_option) => (
 					<Link
 						key={_option.value}
 						className={option}
-						onClick={() => setIsComponentVisible(hide => !hide)}
+						onClick={() => setIsComponentVisible((hide) => !hide)}
 						to={_option.value}
 					>
 						{_option.text}
--- a/src/Hedwig/ClientApp/src/components/Card/Card.tsx
+++ b/src/Hedwig/ClientApp/src/components/Card/Card.tsx
@@ -79,7 +79,7 @@ export function Card({
 			>
 				{showTag && <Tag className="oec-card-tag" text="NEW" color="theme-color-primary" />}
 				<div className="oec-card-cell">
-					{Children.map(children, child => {
+					{Children.map(children, (child) => {
 						if (!isValidElement(child)) {
 							throw new Error('Invalid card child element');
 						}
@@ -95,7 +95,7 @@ export function Card({
 					})}
 				></div>
 				<div className={cx('oec-card-cell', 'oec-card-expansion')} hidden={!isExpanded}>
-					{Children.map(children, child => {
+					{Children.map(children, (child) => {
 						if (!isValidElement(child)) {
 							throw new Error('Invalid card child element');
 						}
--- a/src/Hedwig/ClientApp/src/components/SideNav/SideNav.stories.tsx
+++ b/src/Hedwig/ClientApp/src/components/SideNav/SideNav.stories.tsx
@@ -3,9 +3,10 @@ import { storiesOf } from '@storybook/react';
 
 import { SideNav } from '..';
 import { action } from '@storybook/addon-actions';
+import { SideNavItemProps } from './SideNavItem';
 
 const onClick = action('onChange');
-const exampleItems = [
+const exampleItems: SideNavItemProps[] = [
 	{
 		titleLink: {
 			text: 'Item title is a link',
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.test.tsx
@@ -3,7 +3,7 @@ import { mockAllFakeEnrollments, mockSite } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
 
 // Jest mocks must occur before later imports
 jest.mock('../../../hooks/useApi', () =>
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
@@ -4,7 +4,7 @@ import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
 	mockApiOrganizationsOrgIdReportsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
 
 // Jest mocks must occur before later imports
 jest.mock('../../../hooks/useApi', () =>
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.test.tsx
@@ -6,7 +6,7 @@ import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesIdGet,
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPut,
 	mockApiOrganizationsOrgIdReportsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
 
 // Jest mocks must occur before later imports
 jest.mock('../../../hooks/useApi', () =>
@@ -83,7 +83,7 @@ describe('EnrollmentNew', () => {
 			<TestProvider history={history}>
 				<Route
 					path={'/roster/sites/:siteId/enrollments/:enrollmentId/new/:sectionId'}
-					render={props => (
+					render={(props) => (
 						<EnrollmentNew
 							history={props.history}
 							match={{
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/Update/EnrollmentUpdate.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/Update/EnrollmentUpdate.test.tsx
@@ -3,7 +3,7 @@ import { mockAllFakeEnrollments, mockSite } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
 
 // Jest mocks must occur before later imports
 jest.mock('../../../hooks/useApi', () =>
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
@@ -3,13 +3,19 @@
 // logic in this file, please update that comment accordingly.
 
 // Tell jest we are planning on mocking these modules
-jest.mock('../../../hooks/useApi/error');
+// Mock parseError to become the identity function.
+// We are going to provide the actual error in the API mock.
+jest.mock('../../../hooks/useApi/error', () => ({
+	parseError: (_: any) => _,
+}));
+// Don't provide a factory because we are going to supply a mock for each test as
+// needed.
 jest.mock('../../../hooks/useApi/api');
+
 // We must import with the wildcard selector so we can access the specific
 // property values for jest's mocking. That is, jest requires we do
-// useApiError.parseError.mock instead of importing parseError directly
-// and doing parseError.mock.
-import * as useApiError from '../../../hooks/useApi/error';
+// useApiApi.constructApi.mock instead of importing constructApi directly
+// and doing constructApi.mock.
 import * as useApiApi from '../../../hooks/useApi/api';
 
 import React from 'react';
@@ -29,10 +35,6 @@ describe('enrollment sections', () => {
 		// Tests for whether blocking errors appear on the page
 		describe('shows blocking errors', () => {
 			beforeAll(() => {
-				// Mock parseError to become the identity function.
-				// We are going to provide the actual error in the API mock.
-				const mockedParseError = useApiError as jest.Mocked<typeof useApiError>;
-				mockedParseError.parseError.mockImplementation((_) => _);
 				// Mock constructApi with an extended version of HedwigApi that
 				// mocks the specific methods we needed mocked for this component.
 				// Provide the explicit resolve/reject promise values.
--- a/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
@@ -1,8 +1,22 @@
+// Note: Readers are instructued to go to this file in the mock useApi.ts file
+// for example on how to mock constructApi and parseError. If you change that
+// logic in this file, please update that comment accordingly.
+
+// Tell jest we are planning on mocking these modules
+jest.mock('../../../hooks/useApi/error');
+jest.mock('../../../hooks/useApi/api');
+// We must import with the wildcard selector so we can access the specific
+// property values for jest's mocking. That is, jest requires we do
+// useApiError.parseError.mock instead of importing parseError directly
+// and doing parseError.mock.
+import * as useApiError from '../../../hooks/useApi/error';
+import * as useApiApi from '../../../hooks/useApi/api';
+
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, fireEvent } from '@testing-library/react';
 import ChildInfo from './ChildInfo';
 import { DeepNonUndefineable } from '../../../utils/types';
-import { Enrollment } from '../../../generated';
+import { Enrollment, HedwigApi } from '../../../generated';
 import { mockCompleteEnrollment } from '../../../tests/data';
 import { getValidationError } from '../../../tests/helpers';
 import {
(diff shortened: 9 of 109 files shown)


docker-compose.yaml
```diff
@@ -32,6 +32,7 @@ services:
       - winged-keys
   client:
     image: node:12
+    stdin_open: true
```
cailyncodes, line 35: This is needed because the new version of create-react-app (which got updated when I removed `yarn.lock` to update typescript because the `package.json` did not specify a fixed version and the semantic versioning found a newer "non-breaking" change version [however it was actually breaking for docker]) spawns a shell process, which causes the PID 1 main process to terminate, which closes docker. This holds that process open so docker can continue to run.

src/Hedwig/ClientApp/src/components/ButtonWithDropdown/ButtonWithDrowdown.tsx
```diff
@@ -55,7 +55,7 @@ const ButtonWithDrowdown: React.FC<ButtonWithDrowdownProps> = ({
 						{dropdownElement}
```
cailyncodes, line 55: Only syntax

src/Hedwig/ClientApp/src/components/Card/Card.tsx
```diff
@@ -79,7 +79,7 @@ export function Card({
 			>
```
cailyncodes, line 79: Only syntax

cailyncodes: Actually, there are more of these. I will comment which ones have actually changed moving forward.

src/Hedwig/ClientApp/src/containers/Enrollment/Detail/EnrollmentDetail.test.tsx
```diff
@@ -3,7 +3,7 @@ import { mockAllFakeEnrollments, mockSite } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 6: Changed with the refactor

src/Hedwig/ClientApp/src/containers/Enrollment/Edit/EnrollmentEdit.test.tsx
```diff
@@ -4,7 +4,7 @@ import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
 	mockApiOrganizationsOrgIdReportsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 7: Changed with the refactor

src/Hedwig/ClientApp/src/containers/Enrollment/New/EnrollmentNew.test.tsx
```diff
@@ -6,7 +6,7 @@ import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesIdGet,
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPut,
 	mockApiOrganizationsOrgIdReportsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 9: Changed with the refactor

src/Hedwig/ClientApp/src/containers/Enrollment/Update/EnrollmentUpdate.test.tsx
```diff
@@ -3,7 +3,7 @@ import { mockAllFakeEnrollments, mockSite } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGet,
 	mockApiOrganizationsOrgIdSitesIdGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 6: Changed with the refactor

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -1,120 +1,158 @@
+// Note: Readers are instructued to go to this file in the mock useApi.ts file
```
cailyncodes, line 1: Changed with the refactor. Substantial change

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -1,120 +1,158 @@
...
-					siteId={1}
-					enrollment={{} as DeepNonUndefineable<Enrollment>}
-					updateEnrollment={async () => null}
-					error={{
-						errors: { 'Child.FirstName': ['error'] },
-						status: 400,
-					}}
-				/>
-			);
+		// Tests for whether blocking errors appear on the page
```
cailyncodes, line 29: The tests changed but semantically test the same thing.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyIncome/FamilyIncome.test.tsx
```diff
@@ -1,3 +1,12 @@
+import mockUseApi from '../../../../hooks/useApi/__mocks__/useApi';
+
+// Tell jest to create a factory mock for this module
+// Because we never use the API during these tests, supply
+// an empty object to mockUseApi. This allows tests to run without issue
+// (if we don't give it the factory, jest returned undefined to the useApi
+// call that we thus cannot destructure) and without making a network request.
+jest.mock('../../../../hooks/useApi', () => mockUseApi({}));
```
cailyncodes, line 8: I am not entirely sure why this was working before, but this is needed per the comment above.

src/Hedwig/ClientApp/src/containers/NotSignedInRedirect/NotSignedInRedirect.tsx
```diff
@@ -1,10 +1,10 @@
-import React, { useContext, PropsWithChildren } from 'react';
+import React, { useContext } from 'react';
 import UserContext from '../../contexts/User/UserContext';
 import { useHistory } from 'react-router';
 
 type NotSignedInRedirectProps = {};
 
-export const NotSignedInRedirect = ({ children }: PropsWithChildren<NotSignedInRedirectProps>) => {
+export const NotSignedInRedirect: React.FC<NotSignedInRedirectProps> = ({ children }) => {
```
cailyncodes, line 7: Following the pattern for typing react components with `React.FC` when we can (aka not a generic type component)

src/Hedwig/ClientApp/src/containers/NotSignedInRedirect/NotSignedInRedirect.tsx
```diff
@@ -14,7 +14,7 @@ export const NotSignedInRedirect = ({ children }: PropsWithChildren<NotSignedInR
 			history.push('/');
 			return <></>;
 		} else {
-			return children;
+			return <>{children}</>;
```
cailyncodes, line 17: The new version of typescript does not allow simply returning `children` which is (i think) `React.ReactNode` when it is expecting a `React.ReactElement`.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportDetail.test.tsx
```diff
@@ -4,13 +4,10 @@ import {
...
 	mockCompleteEnrollment,
 	mockEnrollmentWithFoster,
-	mockFullTimeInfantEnrollment,
-	mockPartTimeInfantEnrollment,
-	mockPartTimeEnrollment,
 } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdEnrollmentsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 10: Changed with the refactor.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportDetail.test.tsx
```diff
@@ -4,13 +4,10 @@ import {
 	mockReport as _mockReport,
 	mockCompleteEnrollment,
 	mockEnrollmentWithFoster,
-	mockFullTimeInfantEnrollment,
-	mockPartTimeInfantEnrollment,
-	mockPartTimeEnrollment,
```
cailyncodes, line 9: Unused imports

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.test.tsx
```diff
@@ -8,7 +8,7 @@ import {
 } from '../../../tests/data';
 import mockUseApi, {
 	mockApiOrganizationsOrgIdEnrollmentsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 11: Changed with the refactor.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -9,7 +9,8 @@ import {
 	FundingTimeSplit,
 	ReportingPeriod,
 } from '../../../generated';
-import useApi, { ApiError } from '../../../hooks/useApi';
+import useApi from '../../../hooks/useApi';
+import { ApiError } from '../../../hooks/useApi/error';
```
cailyncodes, line 13: This actually shouldn't have changed. Will make a quick push for that.

src/Hedwig/ClientApp/src/containers/Reports/ReportsSummary/ReportsSummary.test.tsx
```diff
@@ -7,13 +7,13 @@ import {
 import mockUseApi, {
 	mockApiOrganizationsIdGet,
 	mockApiOrganizationsOrgIdReportsGet,
-} from '../../../hooks/__mocks__/useApi';
+} from '../../../hooks/useApi/__mocks__/useApi';
```
cailyncodes, line 10: Changed with refactor. Rest of file is syntax.

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -19,7 +19,8 @@ import {
 	getIdForUser,
 	prettyFundingSpaceTime,
 } from '../../utils/models';
-import useApi, { ApiError } from '../../hooks/useApi';
+import useApi from '../../hooks/useApi';
+import { ApiError } from '../../hooks/useApi/error';
```
cailyncodes, line 23: Shouldn't have changed. Will make a quick push.

src/Hedwig/ClientApp/src/contexts/__mocks__/CommonContextProviderMock.tsx
```diff
@@ -74,7 +74,7 @@ const CommonContextProviderMock: React.FC<CommonContextProviderMockProps> = ({
 	history = createMemoryHistory(),
 }) => {
 	return (
-		<UserContext.Provider value={{ user }}>
+		<UserContext.Provider value={{ user, loading: false }}>
```
cailyncodes, line 77: New typescript did not allow a fully specified provider value object. (`loading` is defined in `UserContext`)

src/Hedwig/ClientApp/src/hooks/useApi/__mocks__/useApi.ts
```diff
@@ -1,3 +1,7 @@
+// This file is @depracated
```
cailyncodes, line 1: Changed wit the refactor.

src/Hedwig/ClientApp/src/hooks/useApi/__mocks__/useApi.ts
```diff
@@ -91,14 +105,23 @@ export const mockApiOrganizationsOrgIdReportsIdGet = (report: CdcReport) => (par
 	data: report,
 });
 
+/**
```
cailyncodes, line 108: More changes.

src/Hedwig/ClientApp/src/hooks/useApi/__mocks__/useApi.ts
```diff
@@ -7,16 +11,18 @@ import {
...
+import { ApiParamOpts } from '..';
 import { useEffect } from 'react';
 
 export const mockApiOrganizationsOrgIdEnrollmentsGet = (enrollments: Enrollment[]) => (
 	params: ApiOrganizationsOrgIdEnrollmentsGetRequest
 ) => {
 	const _enrollments = enrollments
-		.filter(e => !params.siteIds || params.siteIds.includes(e.siteId))
-		.filter(e => {
+		.filter((e) => !params.siteIds || params.siteIds.includes(e.siteId))
```
cailyncodes, line 24: Mostly syntax here except for new `mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsPost` (which can actually be removed, so will make a quick push)

src/Hedwig/ClientApp/src/hooks/useApi/__mocks__/useApi.ts
```diff
@@ -28,15 +34,23 @@ export const mockApiOrganizationsOrgIdEnrollmentsGet = (enrollments: Enrollment[
...
+	params: ApiOrganizationsOrgIdSitesSiteIdEnrollmentsPostRequest
+) => {
+	const thisEnrollment = params.enrollment;
+	if (!thisEnrollment) return;
+	return { loading: false, error: null, data: thisEnrollment };
+};
+
 export const mockApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPut = (enrollments: Enrollment[]) => (
-	params: ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGetRequest
+	params: ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPutRequest
```
cailyncodes, line 51: Fixed this incorrect type

src/Hedwig/ClientApp/src/hooks/useApi/__mocks__/useApi.ts
```diff
@@ -91,14 +105,23 @@ export const mockApiOrganizationsOrgIdReportsIdGet = (report: CdcReport) => (par
...
+ */
 export default (mockApi: any) => (
 	query: (api: any, opt?: any) => any,
 	opts: ApiParamOpts<any> = { skip: false }
 ) => {
 	const { skip, callback, deps } = opts || {};
 
 	useEffect(() => {
-		if (callback && !skip) callback();
+		if (callback && !skip) callback(null);
```
cailyncodes, line 124: `useApi` was changed to have `callback` accept a parameter of `TData | null`. New typescript didn't like that we were using undefined.

src/Hedwig/ClientApp/src/hooks/useApi/api.ts
```diff
@@ -0,0 +1,17 @@
+import { HedwigApi, Configuration } from '../../generated';
```
cailyncodes, line 1: Refactor -- pulling out constructApi

src/Hedwig/ClientApp/src/hooks/useApi/error.ts
```diff
@@ -0,0 +1,26 @@
+import {
```
cailyncodes, line 1: Refactor -- pulling out `parseError`

src/Hedwig/ClientApp/src/hooks/useApi/index.ts
```diff
@@ -0,0 +1,5 @@
+export type { ApiError } from './error';
```
cailyncodes, line 1: Refactor -- export type in practice

src/Hedwig/ClientApp/src/hooks/useApi/paginate.ts
```diff
@@ -0,0 +1,23 @@
+export type ApiExtraParamOpts = {
```
cailyncodes, line 1: Part of refactor -- pulling out paginate for modularity and future potential testing/mocking.

src/Hedwig/ClientApp/src/hooks/useApi/query.ts
```diff
@@ -1,17 +1,9 @@
-import {
-	ValidationProblemDetails,
-	ProblemDetails,
-	HedwigApi,
-	Configuration,
-	ProblemDetailsFromJSON,
-	ValidationProblemDetailsFromJSON,
-} from '../generated';
-import { DeepNonUndefineable } from '../utils/types';
 import { useContext, useEffect, useState } from 'react';
```
cailyncodes, line 1: Previous all inclusive `useApi` file -- paired down to just use the imports and remove inline functions.

src/Hedwig/ClientApp/src/utils/fundingType.tsx
```diff
@@ -23,7 +24,7 @@ export function getFundingTag(options?: {
 }) {
 	const { index, className, fundingSource, fundingTime } = options || {};
 	if (!fundingSource) {
-		return;
+		return <></>;
```
cailyncodes, line 27: Typescript did not like undefined as a return value.

docker-compose.yaml
cailyncodes: https://github.com/facebook/create-react-app/issues/8688

src/Hedwig/ClientApp/src/components/SideNav/SideNav.stories.tsx
```diff
@@ -3,9 +3,10 @@ import { storiesOf } from '@storybook/react';
 
 import { SideNav } from '..';
 import { action } from '@storybook/addon-actions';
+import { SideNavItemProps } from './SideNavItem';
 
 const onClick = action('onChange');
-const exampleItems = [
+const exampleItems: SideNavItemProps[] = [
```
cailyncodes, line 9: Needed for typescript to be nice

cailyncodes:
Found this article helpful in better understanding jest mocks: https://medium.com/better-programming/typescript-jest-testing-challenges-c010eaa8f3f2

src/Hedwig/ClientApp/src/contexts/__mocks__/CommonContextProviderMock.tsx
jlhogan: New typescript did not allow a _partially_ specified provider value object?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -1,120 +1,158 @@
...
+		describe('shows blocking errors', () => {
+			beforeAll(() => {
+				// Mock parseError to become the identity function.
+				// We are going to provide the actual error in the API mock.
+				const mockedParseError = useApiError as jest.Mocked<typeof useApiError>;
```
jlhogan, line 34: can / should this happen by default for the mock?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -1,120 +1,158 @@
...
+				mockedParseError.parseError.mockImplementation((_) => _);
+				// Mock constructApi with an extended version of HedwigApi that
+				// mocks the specific methods we needed mocked for this component.
+				// Provide the explicit resolve/reject promise values.
+				const mockedUseApiApi = useApiApi as jest.Mocked<typeof useApiApi>;
+				mockedUseApiApi.constructApi.mockReturnValue(
+					new (class extends HedwigApi {
```
jlhogan, line 41: this is great

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
```diff
@@ -1,120 +1,158 @@
...
+
+// Tell jest we are planning on mocking these modules
+jest.mock('../../../hooks/useApi/error');
+jest.mock('../../../hooks/useApi/api');
+// We must import with the wildcard selector so we can access the specific
+// property values for jest's mocking. That is, jest requires we do
+// useApiError.parseError.mock instead of importing parseError directly
+// and doing parseError.mock.
+import * as useApiError from '../../../hooks/useApi/error';
+import * as useApiApi from '../../../hooks/useApi/api';
```
jlhogan, line 13: useApiApi gets real weird to look at real fast but it's the best name for the thing i think. but like, the name of some tropical bird? or a volcano ? `useapiapi` ???

jlhogan:
also APPRECIATIONS for the amazing amazing PR description and in line comments !!! made reviewing this a pleasure

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
cailyncodes: Lolol love it -- if you ever change careers and explore remote island and discover a new bird species, you know what to do

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo.test.tsx
cailyncodes: Hmm yeah that makes sense. We can mock it out as a factory to always just use the identity function. (It shouldn't break anything because we should never actually fall into the error processing path unless we have specified an error, in which case we wouldn't want the parsing anyways)

src/Hedwig/ClientApp/src/contexts/__mocks__/CommonContextProviderMock.tsx
cailyncodes: Nope :confused: :woman_shrugging:

src/Hedwig/ClientApp/src/hooks/useApi/__mocks__/useApi.ts
cailyncodes: It's "deprecated" Cailyn, lol

PR #1132, opened 2020-05-28
Change column widths to be consistent across empty tables

### LINKED ISSUE #1082: Empty roster sections have misaligned columns
Looks like when there are no rows in one of the age group sections on the roster, the columns are not lining up with the other sections.

image

cailyncodes:
Should close #1082 

A stop gap measure to make the column consistent. I think the issue is that the table width percentages are not fully deterministic and influenced by the full string of the site name (before truncation). So the site column steals space from other columns, making them smaller and pushing the columns out of alignment. Not really sure how to fix that.

Rider: Adds title attribute so hovering over the text shows the full name


mmazanec22:
I looked into this a bit when I was trying to mess with tabular num alignment-- it seems like a way to have more control over table layout would be to use a library like react-table that can give us divs with roles rather than the native table element.  That seemed like a ton of work and also not really ideal for accessibility so I decided not to delve any further.  I support stop gap measures!

PR #1134, opened 2020-05-29
Refactor Withdrawal page to use new form

### LINKED ISSUE #1133: Refactor Withdrawal to use new Form Component

cailyncodes:
Should close #1133 

Updates Withdrawal to use our new form component. It takes inspiration of a `mutatedEnrollment` state variable a la `ChildInfo/NewForm`.

Also adds logic to prevent the last reporting period dropdown from showing a reporting period before the first reporting period.

Includes a few updates to form to allow for function style state update and updates to utils to support removing DeepNonUndefinable.

### DIFF
--- a/src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
+++ b/src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
@@ -1,7 +1,6 @@
 import React, { FormHTMLAttributes, PropsWithChildren, useState, useEffect } from 'react';
 import { FormProvider } from './FormContext';
 import { ObjectDriller } from './ObjectDriller';
-import cx from 'classnames';
 
 type FormProps<T> = {
 	onSubmit: (_: T) => void;
--- a/src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
+++ b/src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
@@ -1,6 +1,7 @@
 import React, { FormHTMLAttributes, PropsWithChildren, useState, useEffect } from 'react';
 import { FormProvider } from './FormContext';
 import { ObjectDriller } from './ObjectDriller';
+import cx from 'classnames';
 
 type FormProps<T> = {
 	onSubmit: (_: T) => void;
--- a/src/Hedwig/ClientApp/src/components/Form_New/FormContext.ts
+++ b/src/Hedwig/ClientApp/src/components/Form_New/FormContext.ts
@@ -25,6 +25,10 @@ type FormContextType = {
 export type GenericFormContextType<T> = {
 	data: T;
 	dataDriller: ObjectDriller<T>;
+	// Allows for function style update (setState(s => s) as opposed to setState(s))
+	// Needed for components that make multiple edits to the same object
+	// Otherwise just passing the object would result in one of the updates being
+	// overwritten.
 	updateData: React.Dispatch<React.SetStateAction<T>>;
 };
 /**
--- a/src/Hedwig/ClientApp/src/components/Form_New/FormContext.ts
+++ b/src/Hedwig/ClientApp/src/components/Form_New/FormContext.ts
@@ -25,7 +25,7 @@ type FormContextType = {
 export type GenericFormContextType<T> = {
 	data: T;
 	dataDriller: ObjectDriller<T>;
-	updateData: (_: T) => void;
+	updateData: React.Dispatch<React.SetStateAction<T>>;
 };
 /**
  * Utility for casting the un-typed context to the generic with type parameter,
--- a/src/Hedwig/ClientApp/src/components/Form_New/index.ts
+++ b/src/Hedwig/ClientApp/src/components/Form_New/index.ts
@@ -1,0 +1,4 @@
+export { default as Form } from './Form';
+export { default as FormField } from './FormField';
+export * from './FormFieldSet';
+export { default as FormSubmitButton } from './FormSubmitButton';
--- a/src/Hedwig/ClientApp/src/components/Select/Select.tsx
+++ b/src/Hedwig/ClientApp/src/components/Select/Select.tsx
@@ -22,7 +22,7 @@ export type SelectProps = {
 	unselectedText?: string;
 	disabled?: boolean;
 	onChange: React.ChangeEventHandler<HTMLSelectElement>;
-} & Omit<HTMLAttributes<HTMLSelectElement>, 'defaultValue'> &
+} & Omit<HTMLAttributes<HTMLSelectElement>, 'defaultValue' | 'onChange'> &
 	FormFieldStatusProps;
 
 /**
--- a/src/Hedwig/ClientApp/src/containers/Withdrawal/Fields/LastReportingPeriod.tsx
+++ b/src/Hedwig/ClientApp/src/containers/Withdrawal/Fields/LastReportingPeriod.tsx
@@ -10,15 +10,15 @@ import { displayValidationStatus } from '../../../utils/validations/displayValid
 import moment from 'moment';
 import { lastNReportingPeriods, reportingPeriodFormatter } from '../../../utils/models';
 import { Select } from '../../../components/Select/Select';
-import { ErrorAlertState } from '../../../hooks/useCatchallErrorAlert';
+import { ErrorAlertState } from '../../../hooks/useCatchAllErrorAlert';
 
 type LastReportingFieldProps = {
 	errorAlertState?: ErrorAlertState;
 	error: ApiError | null;
 };
 /**
- * This component is used in Withdrawal to update the exit field on enrollment,
- * and certificateEndDate in C4K Certificates (if they exist).
+ * This component is used in Withdrawal to update the last reporting period
+ * on the current cdc funding in an enrollment.
  */
 export const LastReportingPeriodField: React.FC<LastReportingFieldProps> = ({
 	errorAlertState,
@@ -52,7 +52,7 @@ export const LastReportingPeriodField: React.FC<LastReportingFieldProps> = ({
 			}))}
 			onChange={(event) => {
 				const newReportingPeriodId = parseInt(event.target.value);
-				// Update the current c4kCertificate end date with the end date
+				// Update the last reporting period id on the current cdc funding
 				updateData((enrollment) =>
 					produce<Enrollment>(enrollment, (draft) =>
 						set(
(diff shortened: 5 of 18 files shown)


src/Hedwig/ClientApp/src/components/Form_New/Form.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { FormHTMLAttributes, PropsWithChildren, useState, useEffect } from 'react';
 import { FormProvider } from './FormContext';
 import { ObjectDriller } from './ObjectDriller';
+import cx from 'classnames';
```
cailyncodes, line 4: This can be removed

src/Hedwig/ClientApp/src/components/Form_New/FormContext.ts
```diff
@@ -25,7 +25,7 @@ type FormContextType = {
 export type GenericFormContextType<T> = {
 	data: T;
 	dataDriller: ObjectDriller<T>;
-	updateData: (_: T) => void;
+	updateData: React.Dispatch<React.SetStateAction<T>>;
```
cailyncodes, line 28: I needed to be able to use the function style update in Withdrawal because multiple edits to the same object needed to be made in response to one change from the user. Just passing the object would result in one of the updates being overwritten.

src/Hedwig/ClientApp/src/components/Form_New/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as Form } from './Form';
```
cailyncodes, line 1: Created an export file for ease of use in importing these components

src/Hedwig/ClientApp/src/components/Select/Select.tsx
```diff
@@ -22,7 +22,7 @@ export type SelectProps = {
 	unselectedText?: string;
 	disabled?: boolean;
 	onChange: React.ChangeEventHandler<HTMLSelectElement>;
-} & Omit<HTMLAttributes<HTMLSelectElement>, 'defaultValue'> &
+} & Omit<HTMLAttributes<HTMLSelectElement>, 'defaultValue' | 'onChange'> &
```
cailyncodes, line 25: There were two onChange function signatures TS was picking up, making it's typing unsure in calling code

src/Hedwig/ClientApp/src/containers/Withdrawal/Fields/LastReportingPeriod.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Select } from '../../../components/Select/Select';
+import { ErrorAlertState } from '../../../hooks/useCatchallErrorAlert';
+
+type LastReportingFieldProps = {
+	errorAlertState?: ErrorAlertState;
+	error: ApiError | null;
+};
+/**
+ * This component is used in Withdrawal to update the exit field on enrollment,
+ * and certificateEndDate in C4K Certificates (if they exist).
```
cailyncodes, line 21: Bad comment copy pasta -- need to fix

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -1,42 +1,31 @@
-import React, { useContext, useReducer, useState, useEffect } from 'react';
+import React, { useContext, useState, useEffect } from 'react';
```
cailyncodes, line 1: A big file because there was a lot of clean up

src/Hedwig/ClientApp/src/utils/models/funding.ts
```diff
@@ -81,11 +81,11 @@ function isCurrentToRangeCDC(funding: Funding, range: DateRange): boolean {
  * @param fundings
  */
 export function getCurrentCdcFunding(
-	fundings: DeepNonUndefineable<Funding[]> | null
+	fundings: Funding[] | null
```
jlhogan, line 84: :+1: and i've got changes in the open 1104 PR to remove `DeepNonUndefineable` from enrollmentDetail and elsewhere

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -56,164 +45,104 @@ export default function Withdrawal({
...
+	const commonParams:
+		| ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdGetRequest
+		| ApiOrganizationsOrgIdSitesSiteIdEnrollmentsIdPutRequest = {
+		id: enrollmentId,
+		orgId: getIdForUser(user, 'org'),
+		siteId: siteId,
+	};
 
-	const { error: getRequestError, data: getRequestData } = useApi<Enrollment>(
+	const { error: errorOnGet, data: enrollment } = useApi<Enrollment>(
```
jlhogan, line 65: I just wish we could pass around the enrollment instead of having to re-get (from enrollment new into enrollment detail, enrollment detail into enrollmentUpdate and withdraw)

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -56,164 +45,104 @@ export default function Withdrawal({
...
-	const [error, setError] = useState<ApiError | null>(getRequestError);
-	const [attemptedSave, setAttemptedSave] = useState(false); // Only matters once for sake of not showing client errors on first load, before user has a chance to do anything
-	const [attemptingSave, setAttemptingSave] = useState(false); // Matters when deciding to run "mutate" effect
-	useFocusFirstError([error]);
+	const { fundings, site } = mutatedEnrollment || {};
+	const cdcFunding = getCurrentCdcFunding(fundings || []);
+
+	// Tracks whether to show client errors. On first load, we don't show errors,
+	// because the user hasn't had chance to do anything
+	const [attemptedSave, setAttemptedSave] = useState(false);
```
jlhogan, line 87: wish there was a way to rename this for parity with `sectionTouched` & `onSectionTouch`  (think theres room for improvement there with clarity of the names as well) but not a blocker at all -- maybe just making `sectionTouch`/`onSectionTouch` be `attemptedSectionSave` and `onAttemptedSectionSave` ?

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -56,164 +45,104 @@ export default function Withdrawal({
...
-			throw new Error('User is not authorized');
-		}
-		setRequestParams({
-			// These are used in get and put
-			id: enrollmentId,
-			orgId: getIdForUser(user, 'org'),
-			siteId: siteId,
-		});
-	}, [enrollmentId, user, siteId, setRequestParams]);
+	const [error, setError] = useState<ApiError | null>(null);
```
jlhogan, line 48: not sure a separate `error` state var is required; should be able to use `errorOnSave` to display errors..

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -56,164 +45,104 @@ export default function Withdrawal({
...
 		{
-			skip: !attemptingSave || !requestParams || !enrollmentEndDate,
+			skip: !attemptingSave,
 			callback: () => setAttemptingSave(false),
 		}
 	);
+	const errorAlertState = useCatchallErrorAlert(error);
+	useFocusFirstError([error]);
 
 	useEffect(() => {
```
jlhogan, line 115: if opt not to use additional `error` state as mentioned above (and maybe do a check that site has a value somewhere else? altho could also be fixed by marking `site` required in the backend (happy to make changes against the `include` pattern if they'll fix the front end)), this can be refactored to more closely match the api handling we see in other forms a la;
```
useEffect(() => {
  if(saving || errorOnSave) {
    return;
  }
  if(returnedEnrollment) {
    setAlerts([childWithdrawnAlert(nameFormatter(returnedEnrollment.child), site.name || '')]);
    history.push(`/roster`);
 }
}, [...]);
```

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -56,164 +45,104 @@ export default function Withdrawal({
...
 		if (!attemptedSave) {
-			// Only need to mess with this the first time
+			// Only need to update this the first time the user tries to save
 			setAttemptedSave(true);
 		}
 		if (enrollmentEndDate) {
-			// The enrollment end date is verified on the client side because it's how we know we're attempting a withdrawal
-			// If we set the attempting save variable without an end date, then it won't get unset, because the request won't hit the server, and the callback won't be called
+			// The enrollment end date is verified on the client side because it's how we
+			// know we're attempting a withdrawal. If we set the attempting save variable
```
jlhogan, line 138: this comment that i wrote is not the best. improvement? (and maybe hoist up to above the if on 136)
```
Only attempt to save if the user is attempting a valid withdrawal update, which requires `enrollmentEndDate` has a value. Server does not enforce this validation rule because it is only a rule in the context of a withdrawal. 
```

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
```diff
@@ -56,164 +45,104 @@ export default function Withdrawal({
...
+			// without an end date, then it won't get unset, because the request won't hit
+			// the server, and the callback won't be called
 			setAttemptingSave(true);
+			setMutatedEnrollment(userModifiedEnrollment);
```
jlhogan, line 142: logically this _could_ happen in or outside of the if, right? but it only _needs_ to happen if we're actually making the request

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
cailyncodes: We could use an enrollment context? With a hashmap of id to enrollment? Effectively keeping a cache?

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
cailyncodes: The old form also displayed if the GET returned an error. If we don't want to support that, we can just get rid of this extra state var

src/Hedwig/ClientApp/src/containers/Withdrawal/Withdrawal.tsx
cailyncodes: I think that's right

src/Hedwig/ClientApp/src/components/Form_New/FormContext.ts
mmazanec22: These comments are super helpful for understanding what's happening in this PR!  Inline comments would be welcome too-- I'm sure I'll look at this code later and be like "hmm, Cailyn said something about this, but I don't remember what"

PR #1136, opened 2020-05-29
Handle invalid data state error and deep links

### LINKED ISSUE #1109: When navigating between pages after timed logged out, user gets an error message
After leaving staging idle for a time, I tried to navigate to another page and got this error:
Screen Shot 2020-05-19 at 10 52 03 AM

Believe the desired behavior would be to return me to the logged-out homepage.

cailyncodes:
Should close #1109

Provides a better solution to the issue identified in #1109. I don't think it is actually a timeout issue. As long as the user has the browser tab open, the App should be making requests for new access tokens, thus never having a session timeout. Instead, what I think it happening is that a deployment is made on staging, which updates the database and ids. When a user then tries to navigate back to the roster (for example), `useApi` gets an error and sets `data` to `null`. The typing on `useApi` is such that `data` will never be null. So calling code does not handle that case. This PR fixes that and creates a general purpose something went wrong issue with suggestions to refresh or sign out and back in. In case there is actually a 500 error (and not just missing aligned IDs), we don't automatically refresh the page, which may result in infinite loops.

This PR also addresses the issue where link to specific pages would not work, instead being redirected to the roster. The issue was happening because `useApi` was returning `loading: false, data: null` when it could not create the API (because it was waiting for the accessToken). Thus, the `UserContext` was returning `loading: false, user: null` to `NotSignedInRedirect`. This file seeing this state (which looks like user not signed in), redirected the user to the homepage. On subsequent renders on the home page, `user` is now set, so it redirects to the roster.

The last rider is some more general clean up of `DeepNonUndefineable`.


cailyncodes:
Don't actually review this until #1133 lands (or at least is rebased on current master)

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo/ChildInfo.test.tsx
```diff
@@ -58,7 +59,14 @@ describe('enrollment sections', () => {
 
 			it('if rendered without a child first name', async () => {
 				const { findByText } = render(
-					<ChildInfo.Form siteId={1} enrollment={null} updateEnrollment={jest.fn()} error={null} />
+					<TestProvider>
```
cailyncodes, line 62: Wrapping `ChildInfo.Form` is necessary with the changes to `useApi`. An API is only created if the `accessToken` from `AuthenticationContext` is not null. Without providing an `AuthenticationProvider` for the `useContext` to read from, it would return null, which results in the POST query never being fired, which means our mock of the API never is invoked to return the errors.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/ChildInfo/NewForm.tsx
```diff
@@ -33,14 +33,6 @@ export const NewForm: React.FC<SectionProps> = ({
...
 	onSectionTouch,
 }) => {
-	// If this is the first visit, then no enrollment will have been created.
-	// So, check that there is a siteId so we can associate the enrollment with a site
-	// on creation. The siteId should come from the URI.
-	// If this is not the first visit, an enrollment will already have been created.
-	if (!enrollment && !siteId) {
-		throw new Error('ChildInfo rendered without an enrollment or a siteId');
-	}
-
```
cailyncodes, line 43: I don't think this is necessary. `siteId` will always be defined and `enrollment` may or may not be, but that is handled else where later in this file.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportDetail.tsx
```diff
@@ -28,9 +27,15 @@ export default function ReportDetail() {
 		return <div className="Report"></div>;
 	}
 
-	const numEnrollmentsMissingInfo = (report.enrollments || []).filter<
-		DeepNonUndefineable<Enrollment>
-	>((e) => !!e.validationErrors && e.validationErrors.length > 0).length;
+	if (!report.reportingPeriod || !report.organization) {
```
cailyncodes, line 30: This can be changed to only check for `!report.organization`. The backend model was updated to always have `reportingPeriod` defined on `report`.

cailyncodes: This is needed to ensure that we have the organization name to display.

src/Hedwig/ClientApp/src/containers/Reports/ReportDetail/ReportSubmitForm.tsx
```diff
@@ -43,16 +43,17 @@ import { sumWeeksUsed } from '../../../utils/models/fundingTimeSplitUtilization'
...
 };
 
 export default function ReportSubmitForm({
-	report,
+	report: __report,
 	error: inputError,
 	canSubmit,
 }: ReportSubmitFormProps) {
+	const report = __report as DeepNonUndefineable<CdcReport>;
 	const history = useHistory();
```
cailyncodes, line 57: Stop gap enroute to getting rid of `DeepNonUndefineable`

src/Hedwig/ClientApp/src/containers/Reports/ReportsSummary/ReportsSummary.tsx
```diff
@@ -53,13 +51,13 @@ export default function ReportsSummary() {
 						</Link>
 					</th>
 				),
-				sort: (row) => row.reportingPeriod.period.getTime() || 0,
+				sort: (row) => (row.reportingPeriod && row.reportingPeriod.period.getTime()) || 0,
```
cailyncodes, line 54: This might not be needed anymore

src/Hedwig/ClientApp/src/containers/Reports/ReportsSummary/ReportsSummary.tsx
```diff
@@ -53,13 +51,13 @@ export default function ReportsSummary() {
...
 				width: '21%',
 			},
 			{
 				name: 'Type',
 				cell: ({ row }) => <td>{row.type}</td>,
-				sort: (row) => row.type,
+				sort: (row) => row.type as FundingSource,
```
cailyncodes, line 60: Or this

src/Hedwig/ClientApp/src/containers/Reports/ReportsSummary/ReportsSummary.tsx
```diff
@@ -83,25 +81,21 @@ export default function ReportsSummary() {
 				name: 'Due date',
 				cell: ({ row }) => (
 					<td className="oec-table__cell--tabular-nums oec-table__cell--gray">
-						{dateFormatter(row.reportingPeriod.dueAt)}
+						{row.reportingPeriod ? dateFormatter(row.reportingPeriod.dueAt) : ''}
 					</td>
 				),
-				sort: (row) => row.reportingPeriod.dueAt.getTime() || 0,
+				sort: (row) => (row.reportingPeriod && row.reportingPeriod.dueAt.getTime()) || 0,
```
cailyncodes, line 87: Or these

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -42,7 +42,7 @@ export default function Roster() {
 	);
 
 	const sites = organization && organization.sites;
-	const siteIds = (sites || []).map((s) => s.id);
+	const siteIds = (sites || []).map((s) => s.id as number);
```
cailyncodes, line 45: We can probably make a backend change to make site id required

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
```diff
@@ -64,7 +64,22 @@ export default function Roster() {
...
+
+	// If we stopped loading, and still don't have these values
+	// Then an error other than a validation error ocurred.
+	// (Or if in staging, it is possible a new deployment
+	// happened, and then a user navigates back to roster after a delay, which causes
+	// 401/403 errors to occur unless a hard refresh occurs.)
+	// For now, show a general purpose alert message.
+	if (!organization || !_enrollments) {
+		return <Alert {...somethingWentWrongAlert}></Alert>;
+	}
```
cailyncodes, line 80: If we try making a hard refresh here, it could cause an infinite loop if the error is actually a result of a 500 server error

src/Hedwig/ClientApp/src/contexts/__mocks__/CommonContextProviderMock.tsx
```diff
@@ -1,85 +0,0 @@
-import React from 'react';
```
cailyncodes, line 1: We didn't use this mock anywhere

src/Hedwig/ClientApp/src/contexts/__mocks__/TestProvider.tsx
```diff
@@ -20,11 +20,15 @@ const TestProvider: React.FC<TestProviderProps> = ({
 	history = createMemoryHistory(),
 }) => {
 	return (
-		<UserContext.Provider value={{ user, loading: false }}>
-			<ReportingPeriodContext.Provider value={{ cdcReportingPeriods }}>
-				<Router history={history}>{children}</Router>
-			</ReportingPeriodContext.Provider>
-		</UserContext.Provider>
+		<AuthenticationContext.Provider
```
cailyncodes, line 23: See note about the need to add `AuthenticationContext.Provider` in `ChildInfo.test.tsx`

src/Hedwig/ClientApp/src/hooks/useApi/error.ts
```diff
@@ -17,7 +17,7 @@ export const parseError: (error: any) => Promise<ApiError | null> = async (error
 		}
 	} catch (e) {
 		console.error('Error cannot be converted to JSON');
-		console.error(e);
+		console.error(error);
```
cailyncodes, line 20: `e` is just always the error message about not being able to convert to JSON. Show the actually error message from the server.

src/Hedwig/ClientApp/src/hooks/useApi/query.ts
```diff
@@ -40,31 +39,43 @@ export default function useApi<TData>(
...
 
+	// Keep a state variable of the API
+	const [api, setApi] = useState<Api>();
+	// Update it every time the access token changes
+	useEffect(() => {
+		if (accessToken !== null) {
+			const newApi = constructApi(accessToken);
+			setApi(newApi);
+		}
+	}, [accessToken]);
```
cailyncodes, line 50: This was added so we weren't constructing a new API every request

src/Hedwig/ClientApp/src/hooks/useApi/query.ts
```diff
@@ -40,31 +39,43 @@ export default function useApi<TData>(
...
+
 	// Set initial api state
 	const { skip, callback, deps, defaultValue, paginate } = opts;
 	const [state, setState] = useState<ApiState<TData>>({
 		error: null,
 		data: defaultValue || null,
-		loading: true,
+		loading: !skip,
```
cailyncodes, line 57: Set a smart `loading` value -- this is needed for tests that use the useApi hook and have certain expectations on the loading value. Also helpful in `NotSignedInRedirect` so it knows whether a request has actually been fired

src/Hedwig/ClientApp/src/hooks/useApi/query.ts
```diff
@@ -40,31 +39,43 @@ export default function useApi<TData>(
...
 	useEffect(() => {
+		if (!api) {
+			// The we have not yet intialized the API
+			// Keep defaults and exit
+			return;
+		}
+
 		if (skip) {
+			// We are to skip this request, so set loading to false
 			setState({ ...state, loading: false });
```
cailyncodes, line 72: This will already been false if the initial skip value was true. This is needed if a setState changes the skip value in calling code

src/Hedwig/ClientApp/src/utils/forms/form.ts
```diff
@@ -57,7 +57,11 @@ function isChangeEvent<T extends HTMLChoiceElement | HTMLTextAreaElement>(
  * @param state Current state
  * @param updateArg Update mechanism
  */
-export function formReducer<S extends object>(state: S, updateArg: FormReducerUpdate<S>) {
+export function formReducer<S extends object | null>(_state: S, updateArg: FormReducerUpdate<S>) {
+	if (!_state) {
+		return {} as S;
+	}
+	const state = _state as NonNullable<S>;
```
cailyncodes, line 64: These changes allow for state to actually be null -- this is required when we made the `data` property nullable in `useApi`

src/Hedwig/ClientApp/src/containers/Roster/Roster.tsx
jlhogan: were we making a hard refresh here before? confused by this comment

PR #1137, opened 2020-05-29
1103 family info fields

mmazanec22:


src/Hedwig/ClientApp/src/components/Checkbox/Checkbox.tsx
```diff
@@ -44,3 +44,5 @@ export default function Checkbox({
 		</div>
 	);
 }
+
+export default Checkbox;
```
cailyncodes, line 48: Do we want to be exporting this twice?

src/Hedwig/ClientApp/src/components/Select/Select.tsx
```diff
@@ -87,23 +87,23 @@ export const Select: React.FC<SelectProps> = ({
...
 				))}
-				<option key={`${id}-unselected`} value="">
-					{unselectedText || '- Select -'}
-				</option>
 			</select>
 			{options.map((option) => {
 				const expansion = option.expansion;
 				return (
-					<>
+					<div key={`${id}-${option.value}-expansion`}>
```
cailyncodes, line 102: Nit: Should this be a `div` or a `React.Fragment`?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Fields/FosterCheckbox.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const FosterCheckbox: React.FC<FamilyInfoFormFieldProps> = ({ initialLoad }) => (
+	<FormField<Enrollment, CheckboxProps, boolean | null>
+		getValue={(data) => data.at('child').at('foster')}
+		value={'foster'}
+		parseOnChangeEvent={(e) => e.target.checked}
+		inputComponent={Checkbox}
+		id="foster"
+		text={fosterText()}
+		className="margin-top-3"
```
cailyncodes, line 16: Just curious if we need to include this on this component or if it can be lifted to it's parent element when called?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Fields/State.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Enrollment } from '../../../../../generated';
+
+export const State: React.FC<FamilyInfoFormFieldProps> = ({ initialLoad }) => (
+	<FormField<Enrollment, SelectProps, string>
+		id="state"
+		label="State"
+		inputComponent={Select}
+		getValue={(data) => data.at('child').at('family').at('state')}
+		parseOnChangeEvent={(e) => e.target.value}
+		options={['CT', 'MA', 'NY', 'RI'].map((state) => ({ text: state, value: state }))}
```
cailyncodes, line 16: Super super nit: Perhaps we pull this array out as a const in this file?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+	error: inputError,
+	successCallback,
+	onSectionTouch,
+	touchedSections,
+}) => {
+	if (!enrollment || !enrollment.child) {
+		throw new Error('FamilyInfo rendered without a child');
+	}
+
+	// Can we move this duplicated logic up to a wrapper component or something?
```
cailyncodes, line 29: Can you expand on this?

cailyncodes: Oh like not just the `useContext` but like the fair amount of the following lines of code?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+			skip: !attemptSave,
+			callback: () => {
+				setAttemptSave(false);
+				onSectionTouch && onSectionTouch(FamilyInfo);
+			},
+		}
+	);
+
+	// do we need this??
+	const initialLoad = touchedSections ? !touchedSections[FamilyInfo.key] : false;
```
cailyncodes, line 66: This was originally added to prevent showing errors before the user hit save. Not sure if after the refactor it is necessary tho

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+	const [error, setError] = useState<ApiError | null>(inputError);
+	useFocusFirstError([error]);
+	useCatchAllErrorAlert(error);
+
+	const child = mutatedEnrollment.child;
+	const { family } = child || {};
+	const { addressLine1, addressLine2, town, state, zip } = family || {};
+
+	useEffect(() => {
+		// If the request went through, then do the next steps
```
cailyncodes, line 78: Nit: I think it makes more sense if the comment matches the `not` case that this `if` is representing. So like `if the request is still loading, do nothing` or something like that

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+	const [attemptSave, setAttemptSave] = useState(false);
+
+	const initialEnrollment = {
+		...{
+			child: {
+				family: createEmptyFamily(getIdForUser(user, 'org'), enrollment.child.familyId || 0),
+			},
+		},
+		...enrollment,
+	};
```
cailyncodes, line 41: Nit: I see that with this ordering, if `child.family` already exists, it won't be overwritten. I think it would help either to add a comment to that effect and/or wrap this adding of family in an `if (!enrollment.child.family)`

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+	return (
+		<Form<Enrollment>
+			className="FamilyInfoForm usa-form"
+			data={mutatedEnrollment}
+			noValidate
+			autoComplete="off"
+			onSubmit={onFormSubmit}
+		>
+			<h2>Address</h2>
+			<Address {...{ initialLoad, child, addressLine1, addressLine2, zip, state, town }} />
```
cailyncodes, line 105: I'm confused to what is happening here with this destructuring?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Status.ts
```diff
@@ -0,0 +1,15 @@
+import { hasValidationErrors } from '../../../../utils/validations';
+import idx from 'idx';
+import { SectionProps } from '../../enrollmentTypes';
+import { StepStatus } from '../../../../components';
+
+// TODO: this does not need to be an object
```
cailyncodes, line 6: What does this comment mean?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Fields/FosterCheckbox.tsx
jlhogan: i think either is fine :)

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Fields/common.ts
```diff
@@ -0,0 +1,9 @@
+import { ErrorAlertState } from '../../../../../hooks/useCatchAllErrorAlert';
+import { ValidationResponse } from '../../../../../utils/validations/displayValidationStatus';
+
+// Copied from child info-- probably can be consolidated?
```
jlhogan, line 4: agreed. imagine we'll end up doing one last pass to remove `DeepNonUndefineable` and make lil things like this consistent.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
jlhogan: these useEffects to handle the api response can likely be extracted into a reusable hook that takes in some function to call when not error, not loading, and yes response

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
jlhogan: in the `once all the pieces are done lets do a little clean up for consistency`, we can pull the enrollment PUT/POST from the `NewForm`s into  out into `EnrollmentNew` and the enrollment PUT from `UpdateForm` out into `EnrollmentUpdate`

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
mmazanec22: Lol I think I initially wrote is as not, months ago with the useApi refactor, and changed it after a comment to write it this way (and just copied and pasted for this PR).  Happy to do either way, but it seems like the team has divergent personal preferences on language.

@jlhogan do you think that's within scope for this PR?

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Status.ts
mmazanec22: I think it's cruft leftover from mid-refactor-- thanks for catching it!

src/Hedwig/ClientApp/src/components/Checkbox/Checkbox.tsx
mmazanec22: Yes-- we're using named exports for all of the components so we can nicely import them in the containers, but in the components that use Checkbox we're relying on the default export.  I decided editing all of those files was out of scope for this PR and just added what I needed to use the Checkbox on its own.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
jlhogan: no i think we do that in a second pass once the enrollment/funding stuff is done.&  i wouldn't worry abt the comment language. we can figure that out in that final clean up PR.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/Fields/FosterCheckbox.tsx
mmazanec22: I think it needs to be here-- a solo checkbox is the only form component that doesn't have this top margin already.  So any solo checkbox needs to have this specified, and it needs to be specified wherever we use it.  We could add a "solo" flag on the checkbox component, or a "SoloCheckbox" export from the components that is just a checkbox with the margin specified or something.

mmazanec22:
@cailyncodes it looks like we're not naming things that way in ChildInfo or FamilyIncome so I'm going to leave it for now-- in this case I prefer the less verbose naming, though if down the road we standardize on the style you're suggesting that's also fine with me.  For now leaving it because this seems to be consistent with what's currently in the rest of the code base.

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/FamilyInfo/NewForm.tsx
mmazanec22: This is cruft left over from trying to refactor bit by bit-- thanks for seeing it!

PR #1147, opened 2020-06-03
Add tab nav with example

### LINKED ISSUE #1146: Tabs component

cailyncodes:
Should close #1146 

Will remove `Test` after review

### DIFF
--- a/src/Hedwig/ClientApp/src/components/TabNav/TabNav.tsx
+++ b/src/Hedwig/ClientApp/src/components/TabNav/TabNav.tsx
@@ -21,13 +21,13 @@ export const TabNav: React.FC<TabNavProps> = ({ items }) => {
 	};
 
 	const tabs = items.map(({ text }, index) => (
-		<div
+		<button
 			key={index}
 			className={cx('oec-tab-nav--tab', { 'oec-tab-nav--tab__active': index === activeTabIndex })}
 			onClick={() => onClick(index)}
 		>
 			{text}
-		</div>
+		</button>
 	));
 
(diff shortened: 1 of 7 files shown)


src/Hedwig/ClientApp/src/components/TabNav/TabNav.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+	const onClick = (index: number) => {
+		setActiveTabIndex(index);
+	};
+
+	const tabs = items.map(({ text }, index) => (
+		<div
+			key={index}
+			className={cx('oec-tab-nav--tab', { 'oec-tab-nav--tab__active': index === activeTabIndex })}
+			onClick={() => onClick(index)}
```
mmazanec22, line 27: It is always preferred to use native HTML elements instead of rolling our own.  In this case, there are a few issues with using a div instead of a button: even though this element is interactive, it is not automatically focusable with the keyboard, and therefore cannot be interacted with by anyone who can't use a mouse.  You could get around that by using `tabindex="0"` but even then, you would need to roll your own keyboard activity and add an aria role.

Some resources on why native elements are preferred:

- [#1 rule of using aria is don't](https://www.deque.com/blog/top-5-rules-of-aria/)
- [MDN accessibility basics for HTML](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML)
- [WCAG 2.1 on keyboard accessibility](https://www.w3.org/WAI/WCAG21/Understanding/keyboard-accessible)

Issue #1153, opened 2020-06-05
PR system - docker - docker fails to execute - OS watchers limit exceeded


speedcm:
When the PR system spins up, an EC2 startup script is run and the docker command aborts resulting in the following errors:
  - User Interface
    - An unhandled exception occurred while processing the request.
  - Docker Log 
    - Error: ENOSPC: System limit for number of file watchers reached, watch '/home/node/app/public

The EC2 startup script is located at:
  - https://github.com/ctoec/ecis-experimental/blob/master/installer/tf/prsystem/startup.sh

This issue is that the OS watchers limit is being exceeded.  For more information see:
  - https://github.com/gatsbyjs/gatsby/issues/11406

To resolve the issue we need to add the following line to the EC2 startup script (startup.sh):
  - echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p

PR #1154, opened 2020-06-05
fix docker os watchers limit (#1153)

speedcm:
Fixes issue #1153


speedcm:
This issue is fixed.  A PR system was spun up with the PR and is accessible via:
  - https://18.222.125.14:5001

PR #1172, opened 2020-06-10
1105 finish enrollment funding

jlhogan:


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding_NEW/Fields/Funding/index.tsx
```diff
@@ -33,10 +40,11 @@ export const FundingField: React.FC<FundingFormFieldProps> = ({
 			return;
 		}
 
+		// Filter valid funding spaces on enrollment ageGroup
 		const _validFundingSpaces = allFundingSpaces.filter((space) => space.ageGroup === ageGroup);
 
 		setValidFundingSpaces(_validFundingSpaces);
-	}, [dataDriller, allFundingSpaces]);
+	}, [JSON.stringify(dataDriller), allFundingSpaces]);
```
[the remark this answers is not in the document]
cailyncodes: It would be necessary if we are mutating the dataDriller object, but if we only update it with immer.js, then the previous version is okay

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding_NEW/UpdateForm/index.tsx
```diff
@@ -50,19 +49,22 @@ export const UpdateForm: React.FC<SectionProps> = ({
 		}
 
 		// If the request successed, process the response
-		if(returnedEnrollment) {
+		if (returnedEnrollment) {
 			setMutatedEnrollment(returnedEnrollment);
+			// force any open edit forms closed
 			setForceCloseEditForms(true);
+			// then reset forceClosed to enable the user to continue to edit
+			setTimeout(() => setForceCloseEditForms(false), 0);
```
cailyncodes, line 57: Can you explain this?

src/Hedwig/Data/DbInitializer.cs
```diff
@@ -277,6 +277,7 @@ private void DeleteAllData()
 			_context.Sites.RemoveRange(_context.Sites.ToList());
 			_context.Organizations.RemoveRange(_context.Organizations.ToList());
 			_context.Users.RemoveRange(_context.Users.ToList());
+			_context.SaveChanges();
```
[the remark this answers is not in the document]
cailyncodes: :cry:

src/Hedwig/Filters/DTOProjectionFilter.cs
```diff
@@ -35,7 +35,7 @@ public void OnActionExecuted(DTOProjectionFilterAttribute attribute, ActionExecu
 			{
 				// Filter through dto to remove undesired properties
 				var responseEntityAsDto = _mapper.Map(responseEntity, responseEntityType, attribute.Type);
-				responseEntity = _mapper.Map(responseEntityAsDto, attribute.Type, responseEntityType);
+				objectResult.Value = _mapper.Map(responseEntityAsDto, attribute.Type, responseEntityType);
```
cailyncodes, line 38: What does this change do?

jlhogan: without this change, the response coming back out from the API was not the result of the DTO mapping (i.e. the reporting periods and funding spaces were not populated until I made this change)

src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding_NEW/UpdateForm/index.tsx
jlhogan: yeah, please let me know what needs to end up in the comments to make this clear but the gist is:
- set forceClose for all card expansions on a successful edit, to force the open forms closed
- if user wants to be able to edit again after that, but forceClose = true, they cannot (the card expansions will not open)
- so, we just want to set forceClose to true momentarily to close all of the forms, but we do not want to persist this true state moving forward b/c we want to allow the user to make more edits (open more card expansions containing forms)

PR #1180, opened 2020-06-15
947: Update sidenav to function more like tabnav

jlhogan:
this was motivated by how the sidenav is used in the specs for batchedit, specifically:
- need to control cycle logic beyond just via clicks (i.e. completing the last enrollment in sidenav list will cycle back to first enrollment that still has missing info, or if no enrollment with missing info then no item selected)
- need to display content when no item is selected 

it does this by exposing the active item index. Not sure how we feel about exposing the index to the user to control selection (vs some other more logical means) BUT since it is a *list* i figured some sort of ordering was reasonable to assume & leverage.

### DIFF
--- a/src/Hedwig/ClientApp/src/components/SideNav/SideNav.tsx
+++ b/src/Hedwig/ClientApp/src/components/SideNav/SideNav.tsx
@@ -7,21 +7,17 @@ export type SideNavProps = {
 	noActiveItemContent?: JSX.Element;
 };
 
-export const SideNav = ({ 
-	items,
-	externalActiveItemIndex,
-	noActiveItemContent
-}: SideNavProps) => {
+export const SideNav = ({ items, externalActiveItemIndex, noActiveItemContent }: SideNavProps) => {
 	const [activeItemIndex, setActiveItemIndex] = useState(externalActiveItemIndex);
 
 	useEffect(() => {
 		setActiveItemIndex(
 			externalActiveItemIndex != undefined && !isNaN(externalActiveItemIndex)
-			? externalActiveItemIndex
-			: undefined
+				? externalActiveItemIndex
+				: undefined
 		);
-	}, [externalActiveItemIndex])
-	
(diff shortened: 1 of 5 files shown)


src/Hedwig/ClientApp/src/components/SideNav/SideNav.tsx
```diff
@@ -1,20 +1,54 @@
...
 	return (
-		<nav className="oec-sidenav">
-			<div className="tablet:grid-col-4">
-				<ul>
-					{items.map((item) => (
-						<SideNavItem {...item} />
-					))}
-				</ul>
+		<div className="oec-sidenav grid-row grid-gap">
+			<div className="mobile-lg:grid-col-4">
```
jlhogan, line 27: should this get included in a `oec-sidenav-items-menu` class?

src/Hedwig/ClientApp/src/components/SideNav/SideNav.tsx
```diff
@@ -1,20 +1,54 @@
...
+								{...item} 
+								key={idx} 
+								onClick={_onClick} 
+								active={idx === activeItemIndex}
+							/>
+						})}
+					</ul>
+				</nav>
+			</div>
+			<div className="margin-top-2 mobile-lg:grid-col-8">
```
jlhogan, line 46: should this be defined in an `oec-sidenav-content` class?

src/Hedwig/ClientApp/src/components/SideNav/SideNavItem.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import cx from 'classnames';
 import { InlineIcon, Icon } from '..';
+import { TextWithIcon } from '../TextWithIcon/TextWithIcon';
```
mmazanec22, line 4: Seems like this import is not being used

Issue #1139, opened 2020-06-01
DTOs for all controller functions


jlhogan:
**context**: All entities with navigation properties that are returned by controllers should be processed with the DTO projection filter to ensure no cyclical navigation properties are serialized & returned (keeps responses from ballooning with duplicative data)

**AC**:
- DTO exists for result of Organizations `/id` GET request 
(organization including site, site.enrollments, fundingspaces)
- organizations `/id` GET has DTO projection filter with new DTO type
- DTO exists for result of Sites `/id` GET request (site including enrollments, organization)
- Sites `/id` GET has DTO projection filter with new DTO type

PR #1195, opened 2020-06-18
Feature/1139 DTOs for all controller functions

MikeGodin:
This PR partially fulfills #1139 by replacing all the API "GET" calls used to load the main landing page (and a few other calls) to deliver DTOs.  Rather than using projection filters, this PR creates new `Repository` methods that query and return DTOs directly, streamlining the SQL queries to request only the data required to complete the API call, and allowing non-lazy loading of data that is destined for serialization.  This has considerably improved landing page load efficiency with the default database, reducing the number of SQL queries from 315 to 32.


src/Hedwig/Controllers/EnrollmentsController.cs
```diff
@@ -65,17 +65,17 @@ IMapper mapper
...
 			}
 			else
 			// If no sites are specified, get all enrollments for the organization
 			{
-				enrollments = await _enrollments.GetEnrollmentsForOrganizationAsync(orgId, from, to, asOf, skip, take);
+				enrollments = await _enrollments.GetEnrollmentDTOsForOrganizationAsync(orgId, from, to, asOf, skip, take);
 			}
 
-			return enrollments;
+			return Ok(_mapper.Map<List<EnrollmentDTO>,List<Enrollment>>(enrollments));
```
jlhogan, line 78: since we're mapping from the DTO here, we don't need the projection filter anymore

src/Hedwig/Repositories/C4KCertificateRepository.cs
```diff
@@ -17,10 +17,42 @@ public List<C4KCertificate> GetC4KCertificatesByChildId(Guid childId)
...
 		}
+
+		public List<C4KCertificateDTO> GetC4KCertificateDTOsByChildId(Guid childId)
+		{
+			return GetC4KCertificateDTOsByChildIds(new List<Guid>{ childId });
+		}
+
+		public List<C4KCertificateDTO> GetC4KCertificateDTOsByChildIds(IEnumerable<Guid> childIds)
+		{
+			_context.ChangeTracker.LazyLoadingEnabled = false;
```
jlhogan, line 28: what does this do?

src/Hedwig/Repositories/ChildRepository.cs
```diff
@@ -43,6 +49,62 @@ public Child GetChildById(Guid id)
...
+		}
+	}
+
+	public static class ChildQueryExtensions
+	{
+		public static IQueryable<EnrollmentChildDTO> SelectEnrollmentChildDTO(this IQueryable<Child> query)
+		{
+			return query.Select(c => new EnrollmentChildDTO()
+			{
+				Id = c.Id,
```
jlhogan, line 84: pretty much the whole projection filter endeavor was to avoid writing these hard-coded object translations. But if there are benefits to doing it this way (which, sounds like there are!) then i can get on board with a switch 

cc @cailyncodes

src/Hedwig/Repositories/FamilyRepository.cs
```diff
@@ -19,9 +24,44 @@ public Family GetFamilyById(int id)
 
 			return family.FirstOrDefault();
 		}
+
+		public List<EnrollmentFamilyDTO> GetEnrollmentFamilyDTOsByIds(IEnumerable<int> ids)
+		{
+			_context.ChangeTracker.LazyLoadingEnabled = false;
```
jlhogan, line 30: should lazy loading just be turned off across the whole app?

MikeGodin: I'm not 100% familiar with all the back-end processes, which is why I disabled it only on repository methods called by controllers.  If there is a cron-like job that generates reports on the back-end, it may make sense to keep lazy loading enabled (or enable it when those methods are called, if ever).

PR #1197, opened 2020-06-19
1184 embedded step list

### LINKED ISSUE #1184: Styling for step list when embedded in other containers
We reuse the step-list container in batch-edit, but the styling for this step-list when embedded inside another container should be different

Work to this end was started in the `StepList/index.css` file but it is probably bad.

**AC**:
- step list can be styled differently based on _some_ component property & classname
- when this new prop indicates step list is embedded (or whatever you land on to denote this new other style of step list), it looks like the step list n the comps 

https://skylight.invisionapp.com/console/HANDOFF_BatchEdit-Specs-ck8yt41bw00lt010ag6mt32gw/ck8yt41vo00lu010am5yf2bba/play#project_console

mmazanec22:
Closes #1184


src/Hedwig/ClientApp/src/components/StepList/StepList.tsx
```diff
@@ -12,13 +18,16 @@ export type StepProps<T> = {
...
 	Form: React.FC<T>;
+	headerLevel?: PossibleHeaderLevels;
 };
 
 export type StepListProps<T> = {
 	steps: StepProps<T>[];
 	props: T;
 	activeStep: string;
 	type?: 'normal' | 'embedded';
+	// https://dev.to/s_aitchison/psa-stop-hard-coding-heading-levels-in-your-react-components-2ekp
```
jlhogan, line 29: i love these tidbits you leave around, i always learn something new. 
:clap: 

could also do something like this for all the labels etc on inputs and fieldsets .. one day... :)

PR #1198, opened 2020-06-19
Realized this wasn't working how it actually needed to work

jlhogan:
encapsulates the funding creation logic inside the funding option expansion, where it actually should belong

(realized while doing change enrollment work that this had holes; this is not just some correctness pursuit but was actually broken)


src/Hedwig/ClientApp/src/containers/Enrollment/_sections/EnrollmentFunding/Fields/Funding/ContractSpace.tsx
```diff
@@ -65,20 +65,23 @@ const SingleContractSpaceField: React.FC<SingleContractSpaceFieldProps> = ({
...
-		updateData((_data) =>
-			produce<Enrollment>(_data, (draft) =>
-				set(
-					draft,
-					dataDriller
-						.at('fundings')
-						.find((f) => f.id === fundingId)
-						.at('fundingSpaceId').path,
-					fundingSpace.id
+		if(thisFunding.value && thisFunding.at('fundingSpaceId').value != fundingSpace.id) {
```
cailyncodes, line 73: Nit: `!==`

PR #1200, opened 2020-06-22
Sidenav description

### LINKED ISSUE #1185: Batch edit sidenav item descriptions
In batch edit container, each side nav item has a description about which fields are missing

**AC**
- a function somewhere exists that takes in an enrollment and returns a missing info description string
- If enrollment is missing birth certificate or income determination (or re-determination), I see those missing fields named, and then I see a count of how many other fields are missing.
- If enrollment is not missing birth certificate or income determination, but are missing other fields, I only see the count of how many fields are missing: "5 fields missing".

mmazanec22:
Closes #1185

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
+++ b/src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
@@ -34,8 +34,9 @@ const BatchEdit: React.FC<BatchEditProps> = ({ match: { params: sectionId } }) =
 		return <>Loading...</>;
 	}
 
-	const needInfoEnrollments = (enrollments || []).filter((e) =>
-		e.validationErrors && e.validationErrors.length && isFunded(e, { source: FundingSource.CDC })
+	const needInfoEnrollments = (enrollments || []).filter(
+		(e) =>
+			e.validationErrors && e.validationErrors.length && isFunded(e, { source: FundingSource.CDC })
 	);
 	return (
 		<CommonContainer>
--- a/src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
+++ b/src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
@@ -36,11 +36,9 @@ const BatchEdit: React.FC<BatchEditProps> = ({ match: { params: sectionId } }) =
 		return <>Loading...</>;
 	}
 
-	const needInfoEnrollments = (enrollments || []).filter((e) => {
-		return (
-			e.validationErrors && e.validationErrors.length && isFunded(e, { source: FundingSource.CDC })
-		);
-	});
+	const needInfoEnrollments = (enrollments || []).filter((e) =>
+		e.validationErrors && e.validationErrors.length && isFunded(e, { source: FundingSource.CDC })
+	);
 	return (
 		<CommonContainer>
 			<div className="grid-container">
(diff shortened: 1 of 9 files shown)


src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
```diff
@@ -24,6 +24,8 @@ const BatchEdit: React.FC<BatchEditProps> = ({ match: { params: sectionId } }) =
 		orgId: getIdForUser(user, 'org'),
 		startDate: startDate.toDate(),
 		endDate: endDate.toDate(),
+		// Need to include these fields so we can indicate how many are missing data
+		include: ['child', 'family', 'determinations', 'fundings'],
```
jlhogan, line 28: `include` QS param has been deprecated in the API  (moving away from pretending this is GraphQL), so this should not be necessary -- if an of this data is missing in the response, that will need to be a backend change

src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
```diff
@@ -34,10 +36,11 @@ const BatchEdit: React.FC<BatchEditProps> = ({ match: { params: sectionId } }) =
 		return <>Loading...</>;
 	}
 
-	const needInfoEnrollments = (enrollments || []).filter(
-		(e) =>
+	const needInfoEnrollments = (enrollments || []).filter((e) => {
+		return (
```
jlhogan, line 40: why this change?

src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
```diff
@@ -145,6 +145,7 @@ export const SingleEnrollmentEdit: React.FC<SingleEnrollmentEditProps> = ({
 					</div>
 				</div>
 				<div className="margin-top-1">
+					{/* TODO: why isn't this just a link? */}
```
jlhogan, line 148: I dont know. Nothing i can think of / remember

src/Hedwig/ClientApp/src/utils/validations/getMissingInfoPrettyString.ts
```diff
@@ -0,0 +1,91 @@
...
+		missingInfoString = splitCamelCase(birthCertValidationErrors[0].field);
+	}
+
+	if (
+		incomeDeterminationValidationErrors.length > 1 ||
+		!enrollment.child ||
+		!enrollment.child.family
+	) {
+		// Income household size, income and date -> Income determination)
+		missingInfoString += ', income determination';
```
jlhogan, line 71: what if only income determination, not birth certificate is missing? will the string be like ", income determination fields" ?

src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.tsx
mmazanec22: Was probably console logging something-- will change

src/Hedwig/ClientApp/src/utils/validations/getMissingInfoPrettyString.ts
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if (remainingValidationErrors.length) {
+		if (
+			birthCertValidationErrors.length === 0 &&
+			incomeDeterminationValidationErrors.length === 0
+		) {
+			// If enrollment is not missing birth certificate or income determination, but are missing other fields,
+			// I only see the count of how many fields are missing: "5 fields missing".
+			return `${remainingValidationErrors.length} fields`;
```
cailyncodes, line 84: Does "missing" need to be added to this return string?

mmazanec22: No and I should probably take out these comments-- the AC didn't exactly match the prototype, so I followed the prototype

PR #1229, opened 2020-06-25
Fix missing info string

### LINKED ISSUE #1207: Enrollment detail needs to be loaded earlier for batch edit description
~Even though we're creating an empty family in the family info section, it doesn't save when you press save.  Potentially user is not defined yet when we're trying to get the org id?~

We're setting the description before making the enrollment detail get request, so it doesn't include family or sub objects when it first loads.  Fix is to include enrollment details on initial get for enrollments or make a separate endpoint.

mmazanec22:
Closes #1207


src/Hedwig/Models/Enrollment/EnrollmentSummaryDTO.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace Hedwig.Models
 {
+	// TODO: maybe remove this?
```
jlhogan, line 8: i would vote for clean it up!!! we have the code in version control should we want ot bring it back. but cleaning up kruft later is way less likely to happen imo.

PR #1256, opened 2020-07-08
1253 batch edit has report enrollments

### LINKED ISSUE #1253: Batch edit shows only enrollments that are missing info for the specific report I am trying to submit
When I try to submit a report with missing info and click "Update roster," and am taken to batch edit, I see only the enrollments in this report that are missing info.

jlhogan:
closes #1253

### DIFF
--- a/src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
+++ b/src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
@@ -3,7 +3,6 @@ import qs from 'query-string';
 import { useHistory } from 'react-router';
 import { Link } from 'react-router-dom';
 import { Enrollment } from '../../generated';
-import { hasValidationErrors } from '../../utils/validations';
 import { StepProps, StepList } from '../../components';
 import { lastFirstNameFormatter } from '../../utils/stringFormatters';
 import dateFormatter from '../../utils/dateFormatter';
@@ -16,6 +15,7 @@ import FamilyInfo from './_sections/FamilyInfo';
 import FamilyIncome from './_sections/FamilyIncome';
 import EnrollmentFunding from './_sections/EnrollmentFunding';
 import useCatchAllErrorAlert from '../../hooks/useCatchAllErrorAlert';
+import { hasChildInfoSectionErrors, hasFamilyInfoSectionErrors, hasEnrollmentFundingSectionErrors, hasFamilyIncomeSectionErrors } from './utils';
 
 type SingleEnrollmentEditProps = {
 	enrollment: Enrollment;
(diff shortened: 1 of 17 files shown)


src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
```diff
@@ -40,64 +42,54 @@ export const SingleEnrollmentEdit: React.FC<SingleEnrollmentEditProps> = ({
...
+		}
+		if (hasEnrollmentFundingSectionErrors(enrollment)) {
+			_steps.push(EnrollmentFunding);
 		}
 
-		path += `#${firstStep}`;
-		history.push(path);
-	}, [enrollment.id]);
+		setSteps(_steps);
+	}, []);
```
cailyncodes, line 63: Is there a reason this is in a useEffect instead of just directly in the component body?

src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
```diff
@@ -40,64 +42,54 @@ export const SingleEnrollmentEdit: React.FC<SingleEnrollmentEditProps> = ({
...
+
+	useEffect(() => {
+		const firstStep = steps.length ? steps[0].key : 'complete';
+		const queryString = qs.parse(history.location.search);
+		queryString['enrollmentId'] = `${mutatedEnrollment.id}`;
+		history.push({
+			pathname: history.location.pathname,
+			search: qs.stringify(queryString),
+			hash: `${firstStep}`,
+		});
```
cailyncodes, line 73: I think there might be a `createPath` or something like that in the `history` or `location` package that does this concatenation of URL bits. But this might also be the way to do it here.

src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
jlhogan: steps are only set up on initital render to persist ALL the steps as the user moved thru them, to match AC of seeing completed steps 

I can't understand exactly why it is rerendering a thus was recreating the steps based off a more complete enrollment without this restriction, but alas

src/Hedwig/ClientApp/src/containers/BatchEdit/SingleEnrollmentEdit.tsx
jlhogan: not sure there's really a difference between using create path and pushing the path components directly. They'll get concat to the same overall location. 

Tried only pushing the parts that needed to change (here: search + hash, but elsewhere only hash), and that overwrote the other parts as well

PR #1257, opened 2020-07-09
1211 header levels

mmazanec22:
Things are still kind of screwy because of how we're using fieldset legend vs headers to delineate sections in different ways, but this is at least more consistent and clear structurally


src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.test.tsx
```diff
@@ -0,0 +1,62 @@
+// Variables used in jest mockes -- must start with `mock`
```
jlhogan, line 1: omg i love you

src/Hedwig/ClientApp/src/components/FieldSet/FieldSet.tsx
```diff
@@ -2,11 +2,13 @@ import React from 'react';
 import { FormStatus, FormStatusProps } from '..';
 import cx from 'classnames';
 
+export type LegendStyleOptions = 'normal' | 'title';
```
jlhogan, line 5: should this really exist? I made it but now it seems like the right thing to do is define header-style-legends outside of the fieldset?

src/Hedwig/ClientApp/src/tests/data/enrollment.ts
```diff
@@ -10,13 +10,13 @@ import {
...
 } from './fundingSpace';
 
-const enrollmentValidationError = [
-	{
-		message: 'Child has validation errors',
-		isSubObjectValidation: true,
-		field: 'Child',
-	},
-];
+const getValidationError = (field: string, isSubObjectValidation = false) => {
```
jlhogan, line 13: there is a test helper for this  https://github.com/ctoec/ecis-experimental/blob/master/src/Hedwig/ClientApp/src/tests/helpers.ts#L38

src/Hedwig/ClientApp/src/components/FieldSet/FieldSet.tsx
mmazanec22: Probably what we really need to do is define header styles as classes, then have a Heading component that is either just styles or an actual header, and you pass it 'h1' etc, and then just pass fieldset a component, and then not use headers where we should actually be using legends anymore... this seemed like a lighter lift for now though

src/Hedwig/ClientApp/src/containers/BatchEdit/BatchEdit.test.tsx
mmazanec22: This was copy pasted from another file so I think you love @cailyncodes, haha

